PR #2, opened 2022-03-25
Fix incorrectly specified trans header locations for installation.

marsdeno:
Trans headers were not getting installed as not searched for in correct location.

### DIFF
--- a/src/trans/cpu/CMakeLists.txt
+++ b/src/trans/cpu/CMakeLists.txt
@@ -58,7 +58,7 @@ endforeach()
 
 ## Install trans interface
 
-file( GLOB trans_interface interface/* )
+file( GLOB trans_interface include/ectrans/* )
 install(
   FILES        ${trans_interface}
   DESTINATION  include/ectrans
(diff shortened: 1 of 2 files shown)


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf-ifs/ectrans?pullRequest=2) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf-ifs/ectrans?pullRequest=2) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf-ifs/ectrans?pullRequest=2) it.</sub>

wdeconinck:
I believe this was already taken care of in the main branch.

marsdeno:
> I believe this was already taken care of in the main branch.

Although given how I added the gpu codepath, I'm not sure merging in the relevant commit from main branch would be sufficient.

wdeconinck:
In fact the includes (or public interfaces) of trans/gpu and trans/cpu are or should be identical (as opposed to the modules). Indeed we would want the cpu and gpu implementation to be interchangeable in user code. 
We could instead have the interfaces then in "src/ectrans/include", and installed only once therefore.

marsdeno:
Yeah good point, will adjust accordingly.

PR #3, opened 2022-06-09
Add IO-free driver code

samhatfield:


samhatfield:
Example benchmark from

```
OMP_NUM_THREADS=<number of threads> mpirun -n <number of ranks> ./bin/driver-spectrans-sp 79 10 1
```
:

```
Number of ranks | OMP_NUM_THREADS | Total wallclock time | Relative
              1 |               1 |               10.537 | 1.00
              1 |               2 |                6.226 | 0.59
              1 |               4 |                4.161 | 0.39
              1 |               8 |                4.607 | 0.44
              2 |               4 |                3.598 | 0.34
```

samhatfield:
New help message:

```
$ ./bin/driver-spectrans-sp -h

NAME    driver-spectrans-dp
DESCRIPTION
        This program tests ecTrans by transforming fields back and forthbetween spectral 
        space and grid-point space (single-precision version)

USAGE
        driver-spectrans-sp [options]
OPTIONS
        -h      Print this message
        -n iterations
                Run for this many inverse/direct transform iterations
        -t truncation
                Run with this triangular spectral truncation


```

wdeconinck:
Very nice Sam. I have added a little more options:
```
NAME    ectrans-benchmark1-dp

DESCRIPTION
        This program tests ecTrans by transforming fields back and forth between spectral 
        space and grid-point space (double-precision version)

USAGE
        ectrans-benchmark1-dp [options]

OPTIONS
    -h                  Print this message
    -v                  Run with verbose output
    -t, --truncation T  Run with this triangular spectral truncation (default = 79)
    -g, --grid GRID     Run with this grid. Possible values: O<N>, F<N>
                        If not specified, O<N> is used with N=truncation+1 (cubic relation)
    -n, --niter NITER   Run for this many inverse/direct transform iterations (default = 10)
    -f, --nfld NFLD     Number of scalar fields (default = 1)
```
A few things to note.
1) It seems that nprtrv and nprtrw are undefined, see some failing tests.
2) The --nfld option is for now not used
We could already merge I believe if these items are sorted

Further down the line we may need to extend to also use the vorticity-divergence  to wind option simultaneously or separately. I think @ioanhadade  has a full featured driver somewhere that we can look at?

samhatfield:
Good spot on the uninitialised variables - fixed.

There is currently a very dumb way of visualising fields that simply dumps all grid points values serially to unformatted binary files. I have a Python script which can reassemble these files but it only works for a single rank. This was super useful when testing things like half precision as you can instantly spot errors that would be hard to see just from looking at norms. Would you be okay with leaving this feature in and making it off by default, and activated with, say, `-d --dump-values`?

wdeconinck:
> Good spot on the uninitialised variables - fixed.

That's what unit tests are for ;)
> There is currently a very dumb way of visualising fields that simply dumps all grid points values serially to unformatted binary files. I have a Python script which can reassemble these files but it only works for a single rank. This was super useful when testing things like half precision as you can instantly spot errors that would be hard to see just from looking at norms. Would you be okay with leaving this feature in and making it off by default, and activated with, say, `-d --dump-values`?

Sure no problem with that, but perhaps not with the '-d', just with '--dump-values'. We may want to keep the '-d' in our pocket.

marsdeno:
Very nice work guys, will  be very useful!

wdeconinck:
I further adapted the benchmark / test so that it works without MPI as well, e.g. with mpi_serial backend of fiat.
And I implemented more options, making e.g. vordiv2uv, scalar derivatives and uv derivatives an "opt in".
That will be useful for unit tests to narrow down what works and doesn't, especially when porting to gpu.
Many more tests are added that test the benchmark in various configurations in low resolution (T47-O48).

PR #7, opened 2022-07-19
Improve benchmark code

samhatfield:
This is a fairly low priority PR for tidying up the benchmark code. I've already made some quick changes that, in my opinion, improve readability.

This includes:
- Ensuring spaces either side of an =
- Enforcing a 100-column line length
- Spacing out if statements
- Ensuring arguments to functions are separated by spaces

Other things that could be improved:
- Commenting
- Cleaning up the timing and error calculation to avoid repetition.


samhatfield:
I've changed this to a PR for general improvements of the benchmark code, since the latest commit on verbosity levels is not exactly about tidying up.

wdeconinck:
I definitely think that the gstats output should be with "-v" as well.
"-vv" is more for debugging.
What would be the intent of running the benchmark without "-v" (i.e. no output)?
The only reason I could think of is to run it with a profiler, or see if it crashes?

samhatfield:
Yeah I haven't thought that through 😅

It's a benchmark, so people will be running it to see how fast it goes primarily. Then, some basic profiling should be provided at verbosity level 0 (i.e. without `-v` or `-vv`).

As you say,  we could have only verbosity levels 0 (equal to current `-v` level 1), and 1 (equal to current `-vv` level 2).

wdeconinck:
I was just running
```
ectrans-benchmark-dp --grid O320 -t 319 -n 10 --nfld 20
```
And I think we should aim for output like this:
```
======= Start of spectral transforms  =======
 
time step      1 took  4.5541
time step      2 took  5.2916
time step      3 took  5.0156
time step      4 took  4.8555
time step      5 took  4.8176
time step      6 took  4.3537 
time step      7 took  4.5978
time step      8 took  4.3869
time step      9 took  4.0378
time step     10 took  4.5145
 
======= End of spectral transforms  =======

======= Start of time step stats =======
 
Inverse transforms
------------------
avg  (s):   2.4532
min  (s):   2.0991
max  (s):   2.8292
med  (s):   2.3802
 
Direct transforms
-----------------
avg  (s):   2.1893
min  (s):   1.9387
max  (s):   2.5616
med  (s):   2.1302
 
Inverse-direct transforms
-------------------------
avg  (s):   4.6425
min  (s):   4.0378
max  (s):   5.2916
med  (s):   4.5541
loop (s):  46.4677
 
======= End of time step stats =======
 
===-=== START OF TIMING STATISTICS ===-===
 
STATS FOR ALL TASKS
 NUM ROUTINE                                     CALLS      MEAN(ms)       MAX(ms)   FRAC(%)  UNBAL(%)
   0 transform_test - Main transform loop            1     46469.821     46469.821    100.00      0.00
 102 LTINV_CTL      - Inv. Legendre transform       10       269.444       269.444      5.80      0.00
 103 LTDIR_CTL      - Dir. Legendre transform       10       302.693       302.693      6.51      0.00
 106 FTDIR_CTL      - Dir. Fourier transform        10       521.794       521.794     11.23      0.00
 107 FTINV_CTL      - Inv. Fourier transform        10       578.920       578.920     12.46      0.00
 140 SULEG          - Comp. of Leg. poly.            1       132.831       132.831      0.29      0.00
 152 LTINV_CTL      - M to L transposition          10        45.117        45.117      0.97      0.00
 153 LTDIR_CTL      - L to M transposition          10        82.178        82.178      1.77      0.00
 157 FTINV_CTL      - L to G transposition          10      1546.665      1546.665     33.28      0.00
 158 FTDIR_CTL      - G to L transposition          10      1261.537      1261.537     27.15      0.00
 400 GSTATS         - GSTATS itself               3515         0.002         0.002      0.01      0.00
TOTAL MEASURED IMBALANCE =       0.0 SECONDS,  0.0 PERCENT
TOTAL WALLCLOCK TIME      46.470 CPU TIME    126.775 VECTOR TIME     126.775


===-=== END   OF TIMING STATISTICS ===-===
```

Now, would it make sense to add some more gstats labels so that the timings of the time step listed in the first sections are also in gstats?
- time steps
- Inverse transforms
- Direct transforms

Further I was thinking that perhaps norm computations are something we want to output and test in the tests, but not necessarily with the full verbose output.
Could we add a "--norms" option to enable norms computations? not by default enabled with (current) "-vv".
It should be documented in the "--help" that this will skew the timings. That could also be made clear by doing the norms computations within a gstats region.

samhatfield:
Interesting, yes. I mean, if we leave profiling entirely to GSTATS then we wouldn't need to do any of the current manual profiling anyway. I will have a play around later.

wdeconinck:
And additionally, even without verbosity, I would be interested in the time spent in the trans_setup.

We could also introspect memory.
- record memory before setup.
- record memory after setup.
- record memory before/after each time step. The memory should not grow much between time steps. Between first and second there will be a difference because of additional memory allocations on first use.

samhatfield:
How is that done in practice? I can do this manually with, e.g.

```
ectrans-benchmark-sp <options> 2>&1 > /dev/null & watch "grep VmRSS /proc/$!/status"
```

Do we need to use a C standard library call to get such a number from within the benchmark? The only solutions I've seen from within Fortran involve literally reading `/proc/<pid>/status`, [e.g.](https://stackoverflow.com/questions/22028571/track-memory-usage-in-fortran-90).

marsdeno:
> I was just running
> 
> ```
> ectrans-benchmark-dp --grid O320 -t 319 -n 10 --nfld 20
> ```
> 
> And I think we should aim for output like this:
> 
> ```
> ======= Start of spectral transforms  =======
>  
> time step      1 took  4.5541
> time step      2 took  5.2916
> time step      3 took  5.0156
> time step      4 took  4.8555
> time step      5 took  4.8176
> time step      6 took  4.3537 
> time step      7 took  4.5978
> time step      8 took  4.3869
> time step      9 took  4.0378
> time step     10 took  4.5145
>  
> ======= End of spectral transforms  =======
> 
> ======= Start of time step stats =======
>  
> Inverse transforms
> ------------------
> avg  (s):   2.4532
> min  (s):   2.0991
> max  (s):   2.8292
> med  (s):   2.3802
>  
> Direct transforms
> -----------------
> avg  (s):   2.1893
> min  (s):   1.9387
> max  (s):   2.5616
> med  (s):   2.1302
>  
> Inverse-direct transforms
> -------------------------
> avg  (s):   4.6425
> min  (s):   4.0378
> max  (s):   5.2916
> med  (s):   4.5541
> loop (s):  46.4677
>  
> ======= End of time step stats =======
>  
> ===-=== START OF TIMING STATISTICS ===-===
>  
> STATS FOR ALL TASKS
>  NUM ROUTINE                                     CALLS      MEAN(ms)       MAX(ms)   FRAC(%)  UNBAL(%)
>    0 transform_test - Main transform loop            1     46469.821     46469.821    100.00      0.00
>  102 LTINV_CTL      - Inv. Legendre transform       10       269.444       269.444      5.80      0.00
>  103 LTDIR_CTL      - Dir. Legendre transform       10       302.693       302.693      6.51      0.00
>  106 FTDIR_CTL      - Dir. Fourier transform        10       521.794       521.794     11.23      0.00
>  107 FTINV_CTL      - Inv. Fourier transform        10       578.920       578.920     12.46      0.00
>  140 SULEG          - Comp. of Leg. poly.            1       132.831       132.831      0.29      0.00
>  152 LTINV_CTL      - M to L transposition          10        45.117        45.117      0.97      0.00
>  153 LTDIR_CTL      - L to M transposition          10        82.178        82.178      1.77      0.00
>  157 FTINV_CTL      - L to G transposition          10      1546.665      1546.665     33.28      0.00
>  158 FTDIR_CTL      - G to L transposition          10      1261.537      1261.537     27.15      0.00
>  400 GSTATS         - GSTATS itself               3515         0.002         0.002      0.01      0.00
> TOTAL MEASURED IMBALANCE =       0.0 SECONDS,  0.0 PERCENT
> TOTAL WALLCLOCK TIME      46.470 CPU TIME    126.775 VECTOR TIME     126.775
> 
> 
> ===-=== END   OF TIMING STATISTICS ===-===
> ```
> 
> Now, would it make sense to add some more gstats labels so that the timings of the time step listed in the first sections are also in gstats?
> 
> * time steps
> * Inverse transforms
> * Direct transforms
> 
> Further I was thinking that perhaps norm computations are something we want to output and test in the tests, but not necessarily with the full verbose output. Could we add a "--norms" option to enable norms computations? not by default enabled with (current) "-vv". It should be documented in the "--help" that this will skew the timings. That could also be made clear by doing the norms computations within a gstats region.

I definitely agree with the idea of --norms :+1:

marsdeno:
Question about dimensions setup from sumpini (l 285 in your revised version) do we want to expose something like nprtrv as a run-time argument ?

wdeconinck:
> Question about dimensions setup from sumpini (l 285 in your revised version) do we want to expose something like nprtrv as a run-time argument ?

Good point. This is quite important to control. Are there more?

samhatfield:
There are I'm sure several variables previously read from a namelist that could be parametrised as options. `nprtrv` but also `nprgpns` are good candidates. There'd have to be some validation to ensure both can divide into the total number of requested MPI ranks.

We could also reintroduce the namelist approach, but I'm personally reluctant since it adds another couple of steps to running the benchmark.

samhatfield:
@wdeconinck - @marsdeno pointed out that `nflevg` is set to `nfld` currently: https://github.com/ecmwf-ifs/ectrans/blame/main/src/programs/ectrans-benchmark.F90#L233. Should it actually be set to `nlev`?

wdeconinck:
Indeed that should be `nlev`. It must be a remnant of when nlev was nfld and nfld was 1, or something of that sort.

wdeconinck (approved):

samhatfield:
Another thing to consider is `meminfo.txt`. Currently output is printed to `meminfo.txt` when calling `MPL_END` (we don't provide any `LDMEMINFO` argument to `MPL_END` in the benchmark, but it's `.TRUE.` by default). Subsequent invocations of the benchmark in the same directory append to this same file.

If we wanted to support a `--meminfo` option for the benchmark which outputs to stdout, what would be the easiest way to do that?

If we didn't have to worry about breaking all of Fiat's other dependencies (namely the IFS) then it would be nifty to remove the `LDMEMINFO` argument and replace it with an integer, with `-1` sending the output to `meminfo.txt`, and every other value interpreted as a unit number, so `6` would send it to stdout, for example.

Alternatively maybe there's a way in the benchmark to intercept the `501` unit number [used by `EC_MEMINFO`](https://github.com/ecmwf-ifs/fiat/blob/main/src/fiat/util/ec_meminfo.F90#L219) so it prints to stdout instead of `meminfo.txt`. But that's a bit hacky.

samhatfield:
The latter option isn't possible.

@wdeconinck - what do you think about adding an optional argument to `MPL_END` (and `EC_MPI_FINALIZE`) which contains the unit number for `EC_MEMINFO`? We could leave the `LDMEMINFO` option untouched so we don't break existing code. `6`  would be given in the benchmark for this option.

Or is this all too much effort for something trivial?

wdeconinck:
I would call the `ec_meminfo` function manually with all the arguments you need.
```
#include "ec_meminfo.intfb.h"
call ec_meminfo( ... )
```
and then not ask `mpl_end` to print:
```
mpl_end(ldmeminfo=.false.)
```

Issue #13, opened 2022-11-23
Intel compile failure from illegal code


DJDavies2:
I am getting compile failures with Intel, for example

/home/d03/frwd/cylc-run/AddEctransFiat/share/mo-bundle/ectrans/src/trans/internal/ftdirad_mod.F90(92): error #8284: If the actual argument is scalar, the dummy argument shall be scalar unless the actual argument is of type character or is an element of an array that is not assumed shape, pointer, or polymorphic.   [A]
    CALL FFT992(PREEL(1,IOFF),T%TRIGS(1,KGL),&

wdeconinck:
Hi @DJDavies2  thank you for this report.
In order for us to reproduce, which environment, architecture and version of Intel compiler is being used?
We already have a continuous integration in place using Intel OneAPI which is not failing like that
(e.g. https://github.com/ecmwf-ifs/ectrans/actions/runs/3512552361/jobs/5884373860).

PR #14, opened 2022-11-24
Fix illegal subroutine calls.

DJDavies2:
Addresses https://github.com/ecmwf-ifs/ectrans/issues/13.

I think the code concerned is not standard conforming, it appears in four places.  It concerns passing array elements to F77-style arrays; this is okay in many situations but in some circumstances it isn't.  For example if the array being passed is an assumed shape dummy argument it is not okay.  It is to do with array contiguity; passing arrays via single elements in this way relies on the arrays being known to be contiguous which for an assumed shape dummy argument is not the case (for example for the Fortran 2003 FCD see 12.4.1.2 several paragraphs down).

Code builds on all the platforms I have.


FussyDuck:
[CLA assistant check](http://claassistant.ecmwf.int:80/ecmwf-ifs/ectrans?pullRequest=14) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](http://claassistant.ecmwf.int:80/ecmwf-ifs/ectrans?pullRequest=14) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](http://claassistant.ecmwf.int:80/check/ecmwf-ifs/ectrans?pullRequest=14) it.</sub>

src/trans/internal/ftdir_mod.F90
```diff
@@ -86,7 +86,7 @@ SUBROUTINE FTDIR(PREEL,KFIELDS,KGL)
 
     IF( T%LUSEFFT992(KGL) )THEN
 
-      CALL FFT992(PREEL(1,IOFF),T%TRIGS(1,KGL),&
+      CALL FFT992(PREEL(1:,IOFF:),T%TRIGS(1,KGL),&
```
wdeconinck, line 89: The first argument to FFT992 is expected to be a 1-dimensional array.
Would it then not be sufficient to pass `PREEL(:,IOFF)` ?

DJDavies2: Yes you are correct.  I have pushed an update to the branch.

DJDavies2:
Is this okay to merge?

wdeconinck:
> Is this okay to merge?

Yes it is, as soon as the CLA is signed by all contributors.

wdeconinck:
Thank you @DJDavies2 for this fix!

PR #18, opened 2022-12-19
Introduce V-W set specification as command line arguments in benchmark

samhatfield:
I had a case at TCO7999 where the default behaviour of setting NPRTRV and NPRTRW didn't work. In this case it's useful to override this behaviour and force the parameters yourself, so I've added them as command line arguments. You can provide just one of the two, in which case the existing logic will determine the other by dividing NPROC by the provided value. It will also make sure the product of the two equals NPROC.

I've also included a commit which moves all default option values from `get_command_line_arguments` to the top of the program, for consistency,


wdeconinck (approved):
Looks great!

PR #19, opened 2022-12-20
Qualify the module imports of SULEG

samhatfield:
I found it quite hard to track down variables like `D`, `R`, `F`, etc. in `SULEG`. This PR qualifies all module imports in `SULEG` so it's clear where external variables came from. I focused on `SULEG` because it has an unusually large number of module imports, but I think we should gradually start doing this elsewhere as well.


samhatfield:
Debug, linux intel : segfault. Well that was unexpected. A debug build with intel/2021.4.0 on the Atos HPC passes all tests.

wdeconinck:
It seems that "develop" is also failing now. Perhaps something intermittent or new compiler version used by intel oneapi.

samhatfield:
Yeah it's very weird.

samhatfield:
@wdeconinck  The tests are magically passing now.

wdeconinck:
Yes I tried rerunning yesterday and now all tests passed. It's what I expected as these changes should not have had any negative impact.

wdeconinck (approved):
Nice cleanup!

PR #21, opened 2023-01-03
Implement correctness checker

samhatfield:
This PR is intended to be a discussion area for a "correctness checker", which checks that fields are being transformed back and forth correctly. This should be implemented by comparing the fields in spectral space after one or more transform cycles with the initial condition and verifying that they match to some tolerance. They won't be identical due to unavoidable rounding errors, but should roughly match to the machine epsilon.

The correctness checker is activated by passing `--check` to the benchmark code. It can be run manually or through the standard CTest command.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -106,7 +106,6 @@ real(kind=jprb), pointer :: zspvor(:,:) => null()
 real(kind=jprb), pointer :: zspdiv(:,:) => null()
 real(kind=jprb), pointer :: zspsc3a(:,:,:) => null()
 real(kind=jprb), allocatable :: zspsc2(:,:)
-real(kind=jprb), allocatable :: zsp_check(:)
 
 logical :: lstack = .false. ! Output stack info
 logical :: luserpnm = .false.
@@ -201,8 +200,6 @@ logical :: luse_mpi = .true.
 
 character(len=16) :: cgrid = ''
 
-real(kind=jprb) :: zmax_diff
-
 !===================================================================================================
 
 #include "setup_trans0.h"
@@ -245,7 +242,6 @@ call dr_hook_init()
 
 if( lstats ) call gstats(0,0)
 ztinit = timef()
-
 
 ! only output to stdout on pe 1
 if (nproc > 1) then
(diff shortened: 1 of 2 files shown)


samhatfield:
Currently I check that the initial and final fields differ by no more than the machine epsilon, and this fails for most setups. I'm not sure if that indicates something is going wrong. We may need to make it less strict.

wdeconinck:
That's great Sam. As first observation, tolerances seem a bit too strict.

reuterbal:
Not sure if it's directly transferable: CLOUDSC dwarf flags deviations greater than 10*eps as problematic: https://github.com/ecmwf-ifs/dwarf-p-cloudsc/blob/8cc1c8b1959a353984c8e39a9f21268906fedf19/src/common/module/validate_mod.F90#L289

samhatfield:
10 times the epsilon would be a fairer test. In fact I think it's best to adapt the existing spectral norm error calculations and check those instead. Then all fields will be checked, not just the surface fields. With this (and starting with a tolerance of `100.0_jprb * epsilon (1.0_jprb)`) I'm finding that most tests with FLT enabled are failing, as well as some residual ones with various combinations of ranks, threads, and options. However these remaining failures aren't always reproducible which points to some race condition somewhere, within ecTrans itself. I'll dig deeper.

samhatfield:
Yup, I was testing the norm on all ranks, but only the first rank has a number that makes sense. With that fix, only (and all of) the 9 double-precision tests with the FLT enabled fail (with this albeit quite large tolerance).

samhatfield:
The double-precision FLT cases fail because they give errors of around 1E-13 which is above the current arbitrary threshold of ~1E-14 (100 times the machine epsilon). The FLT error can be reduced by lowering [this number](https://github.com/ecmwf-ifs/ectrans/blob/main/src/trans/internal/suleg_mod.F90#L185) which controls the level of compression. In other words, the FLT error is expected and normal.

I can think of two options then for letting the double-precision FLT cases pass:
1. Increase the tolerance value of the test to 1000 times the machine epsilon. Downside: this would make the test for the other non-FLT cases less stringent.
2. Make the `--check` argument accept an argument which determines the tolerance, and make this different and slightly bigger for the FLT test cases, keeping it to the lower value, 100 times the machine epsilon, for the others. Downside: I have to think of a way to pass a floating-point literal on the command line...

samhatfield:
I went for the second option, so that the FLT tests use 1000 times the epsilon as the threshold, and the rest use 100 times the epsilon.

wdeconinck (approved):
All very nice! Happy to merge unless anyone has something else to add.

PR #30, opened 2023-08-09
Destroy FFT plans at job end

raguridan:


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ectrans?pullRequest=30) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ectrans?pullRequest=30) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/ectrans?pullRequest=30) it.</sub>

wdeconinck:
Is this not also required for the CPU version?

marsdeno:
> Is this not also required for the CPU version?

That is a good point, it would make sense for this to be done for FFTW as well.

wdeconinck:
In fact for CPU, we already had the call to destroy_fftw_plans when deallociting a resolution:
https://github.com/ecmwf-ifs/ectrans/blob/main/src/trans/internal/dealloc_resol_mod.F90#L165-L170
So, why was this not sufficient, and why can a similar approach for the gpu not be achieved?

raguridan:
I cannot find dealloc_resol_mod.F90 in the GPU version.
Is DEALLOC_RESOL called in the GPU version as well?

marsdeno:
Yes, Willem is right, putting it in dealloc_resol is better than putting it straight into trans_end; dealloc_resol (you should see it in src/trans/gpu/internal) gets called from trans_end.

marsdeno:
My only hesitation now would be whether we're sure that the RESTRICT_PLANS code has well and truly done its time?

raguridan:
I asked if I should remove that mechanism or not and the reply was that we don't need it on the amdgpu branch.

After debugging through that RESTRICT_PLANS mechanism, I would also not recommend keeping it around (destroying a plan while it's being created/executed does not seem like a good idea on any system).

wdeconinck (approved):
All good. @marsdeno  if you agree, please Approve.

reuterbal:
Confirmed working on LUMI-G

marsdeno:
> I asked if I should remove that mechanism or not and the reply was that we don't need it on the amdgpu branch.
> 
> After debugging through that RESTRICT_PLANS mechanism, I would also not recommend keeping it around (destroying a plan while it's being created/executed does not seem like a good idea on any system).

Fair enough, good for me.

marsdeno (approved):

PR #32, opened 2023-08-25
Have all ranks error stop on failed correctness check

samhatfield:
Previously only the root rank stopped, which I don't think was raceproof.


wdeconinck (approved):
Thanks @samhatfield this is a good fix!

PR #33, opened 2023-08-25
Discover MPI and propagate OMP for CCE

reuterbal:
The MPI feature did not depend on actually finding MPI - which only the GPU-aware MPI did. I changed the logic to make the feature MPI depend on finding MPI, and GPU-aware MPI depend on MPI being enabled.

The GPU version currently requires MPI, thus changing the LIBRARY dependency to plain MPI being available.

Lastly, fix an issue with Cray compiler where some OpenMP symbols would not be resolved if the OpenMP dependency isn't propagated via the build interface.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ ecbuild_find_package( NAME fiat REQUIRED )
 
 ecbuild_add_option( FEATURE MPI
                     DESCRIPTION "Support for MPI distributed memory parallelism"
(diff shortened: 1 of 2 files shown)


wdeconinck (changes requested):
As brought up above, some changes are requested.

wdeconinck:
@reuterbal please see if the alternative suggestion in #40 also works, and if could work for the redgreengpu branch.

reuterbal:
Superseded by #40

PR #36, opened 2023-09-05
Direct transform optimizations

PaulMullowney:
This PR optimizes portions of the direct transform. I see a pretty nice boost.

1) When not using GPU-aware MPI, I allocate ZGTF in ftdir_ctl_mod. This variable is passed in as PGLAT in the TRGTOL subroutine.
2) Then, the methods for computing PGLAT are ported to device for ACC and OMP. This includes the "Copy local contribution" and "unpack" regions.
3) There is code at the end TRGTOL() that does ACC updates based on an IF(.NOT. LGPNORM) condition. I don't think either of these steps are necessary however I think a reviewer should look carefully at my changes.


src/trans/gpu/internal/trgtol_mod.F90
```diff
@@ -1482,13 +1548,12 @@ SUBROUTINE TRGTOL(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,KPTRGP,&
 ! when run as gpnorm on cpu needs to avoid this update call
 IF(.NOT.LGPNORM) THEN
 #ifdef ACCGPU
-  !$ACC UPDATE DEVICE(PGLAT)
-#endif
-#ifdef OMPGPU
-  !$OMP TARGET UPDATE TO(PGLAT)
+!! All data is updated on device already. PJM 9/05/2023
+!  !$ACC UPDATE DEVICE(PGLAT)
```
samhatfield, line 1552: Hi @PaulMullowney - I see what you mean. I don't know the issue with LGPNORM. However one level up in ftdir_ctl_mod.F90, we _always_ update `ZGTF`:
```
#ifdef ACCGPU
!$ACC UPDATE DEVICE(ZGTF)
#endif
#ifdef OMPGPU
!$OMP TARGET UPDATE TO(ZGTF)
#endif
```

Is that what you're referring to?

PaulMullowney: Any update on getting this PR reviewed, tested and merged?

samhatfield:
Thanks for chasing this up Paul. I'm not an expert on TRGTOL, but this is what I think you are proposing:

- Some values in PGLAT can be populated without MPI communication because the data is already resident on the current compute rank. PGLAT will end up on the device eventually, so we might as well do this local transposition on the device.
- The rest of the values of PGLAT come from MPI communication. This is the non-GPU-aware routine, so this communication should take place using host buffers, then the result can be transferred to device to finish the populating of PGLAT. PGLAT is now complete and on device, so no `UPDATE DEVICE(PGLAT)` is required.

Is that roughly right?

I have a few comments but I'll put those in the code.

src/trans/gpu/internal/trgtol_mod.F90
```diff
@@ -1243,67 +1260,102 @@ SUBROUTINE TRGTOL(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,KPTRGP,&
...
+    !$OMP TARGET DATA MAP(TO:IFLDOFF,IGPTROFF,LLGP2,LLGP3A,LLGP3B,LLUV,IGPTRSEND)
+#endif
     CALL GSTATS(1601,0)
-#ifdef NECSX
-  !$OMP PARALLEL DO SCHEDULE(DYNAMIC) PRIVATE(JFLD,JBLK,JK,IFLD,IPOS,IFIRST,ILAST)
-#else
-  !$OMP PARALLEL DO SCHEDULE(STATIC) PRIVATE(JFLD,JBLK,JK,IFLD,IPOS,IFIRST,ILAST)
+    IF(LLPGPONLY) THEN
+#ifdef ACCGPU
+      !$ACC PARALLEL LOOP COLLAPSE(3) PRIVATE(IPOS,IFIRST,ILAST,IFLD,JK)
```
samhatfield, line 1282: @reuterbal - we are aiming for all parallel regions to have `DEFAULT(NONE)` right? In which case this would have to be added to all new parallel regions in this file. `PGLAT` would have to be explicitly declared then as well (and maybe other variables I missed).

src/trans/gpu/internal/trgtol_mod.F90
```diff
@@ -1482,13 +1548,12 @@ SUBROUTINE TRGTOL(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,KPTRGP,&
...
-#endif
-#ifdef OMPGPU
-  !$OMP TARGET UPDATE TO(PGLAT)
+!! All data is updated on device already. PJM 9/05/2023
+!  !$ACC UPDATE DEVICE(PGLAT)
 #endif
 #ifdef ACCGPU
-  !$ACC WAIT
+!! I don't understand the need for this. PJM 9/05/2023
+!  !$ACC WAIT
```
samhatfield, line 1556: Hmm there are no `ASYNC` statements in this file. Maybe this is just leftover code?

src/trans/gpu/internal/trgtol_mod.F90
PaulMullowney: I can comply with style.

samhatfield:
I'm not sure why but the linking fails when I try to build this branch on Frontier:

```
ld.lld: error: unable to find library -lpgas-shmem
ld.lld: error: unable to find library -lmodules
ld.lld: error: unable to find library -lfi
ld.lld: error: unable to find library -lcraymath
ld.lld: error: unable to find library -lf
ld.lld: error: unable to find library -lu
ld.lld: error: unable to find library -lcsup
```

PaulMullowney:
> Thanks for chasing this up Paul. I'm not an expert on TRGTOL, but this is what I think you are proposing:
> 
> * Some values in PGLAT can be populated without MPI communication because the data is already resident on the current compute rank. PGLAT will end up on the device eventually, so we might as well do this local transposition on the device.
> * The rest of the values of PGLAT come from MPI communication. This is the non-GPU-aware routine, so this communication should take place using host buffers, then the result can be transferred to device to finish the populating of PGLAT. PGLAT is now complete and on device, so no `UPDATE DEVICE(PGLAT)` is required.
> 
> Is that roughly right?

This is correct. In the current, non-GPU aware version of the code, PGLAT gets transferred from device at the end TRGTOL and then immediately back to the GPU in FTDIR_CTL_MOD. My primary objective is to eliminate these unnecessary memcopies.

src/trans/gpu/internal/trgtol_mod.F90
PaulMullowney: Not sure what to do with this code. Anyone else know why we need this?

src/trans/gpu/internal/trgtol_mod.F90
PaulMullowney: I've just updated the PR which complies, mostly, with the style conventions.

src/trans/gpu/internal/ftdir_ctl_mod.F90
```diff
@@ -204,14 +202,14 @@ END SUBROUTINE cudaProfilerStop
 !$ACC END DATA
 #endif
 #else
-CALL TRGTOL(ZGTF,KF_FS,KF_GP,KF_SCALARS_G,IVSET,KPTRGP,&
- &PGP,PGPUV,PGP3A,PGP3B,PGP2)
 #ifdef ACCGPU
-!$ACC UPDATE DEVICE(ZGTF)
+!$ACC DATA CREATE(ZGTF)
```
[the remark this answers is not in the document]
PaulMullowney: This is gone down. TRGTOL and TRGTOL_CUDAAWARE have much more similar pathways for GPU usage. It's worth looking at ftdir_ctl_mod to see what I mean.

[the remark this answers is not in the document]
samhatfield: I agree. I would love to see those two merged at some point. For now, let's keep them separate, as a couple of us are also tweaking these subroutines and hope to submit PRs eventually.

src/trans/gpu/internal/trgtol_mod.F90
samhatfield: I think it's because `IGP3BLEVS` is only conditionally present. If `PRESENT(PGP3B) == .FALSE.` then it will not be copied to device on [line 1244](https://github.com/PaulMullowney/ectrans/blob/ftdir_opt/src/trans/gpu/internal/trgtol_mod.F90#L1244C18-L1244C32) and hence won't be in the present table for the enclosed compute constructs. Talking to @reuterbal about this, we think that you can delete `IGP3BLEVS` and other variables from the present lists of all compute constructs inside the main data region (starting [line 1237](https://github.com/PaulMullowney/ectrans/blob/ftdir_opt/src/trans/gpu/internal/trgtol_mod.F90#L1237)). Even though the compute constructs have `DEFAULT(NONE)`, we think it shouldn't be a problem because they inherit the present list of the enclosing data region. The OpenACC standard is not entirely clear on this though. For now I suggest to delete `IGP3BLEVS` and other problematic variables from the present list.

samhatfield:
Thanks Paul, I will take a look again soon.

@wdeconinck - should Paul rebase these changes against redgreengpu, instead of amdgpu?

samhatfield:
Hi @PaulMullowney - I spoke to Willem and he said that the amdgpu branch is intended to be fixed going forwards. That's supposed to be a stable branch for Destination Earth. For development we instead have redgreengpu. This branch is supposed to have the latest state of the art implementation that runs on both Nvidia and AMD accelerators. Could you rebase your PR against redgreengpu? Then I can give it a spin on Frontier.

PaulMullowney:
> Hi @PaulMullowney - I spoke to Willem and he said that the amdgpu branch is intended to be fixed going forwards. That's supposed to be a stable branch for Destination Earth. For development we instead have redgreengpu. This branch is supposed to have the latest state of the art implementation that runs on both Nvidia and AMD accelerators. Could you rebase your PR against redgreengpu? Then I can give it a spin on Frontier.

No problem. I'll rebase.

samhatfield:
amdgpu is no longer maintained. This PR could probably be applied to redgreengpu, but that branch will also be defunct once redgreen-optimized is merged. I'll leave this open for now.

samhatfield:
Since we're finally making progress with AMD support in develop, I will close this.

PR #39, opened 2023-09-07
Fix GPU-aware MPI

samhatfield:
A potentially contentious list of changes. Needs discussion with @reuterbal, @anmrde and @marsdeno. These are the things I needed to change to get the MPI-aware ecTrans benchmark working on Frontier. With this, the norms are bit-identical between GPU-aware and non-GPU-aware benchmarks.

Main things are:
- Why were those ASYNC statements there?
- Is that OMP barrier needed?
- Are the changes in 2aa77298ccd134a64db221355211a793b44f6cdd simply because `MPICH_GPU_SUPPORT_ENABLED=1` wasn't exported when running?


samhatfield:
Awesome! I think everything is in place for a merge now.

reuterbal (approved):

samhatfield:
Hmmm. I tested this branch again and the non-GPU-aware version segfaulted once. I repeated the test and it ran fine. @reuterbal - could you just submit two tests again, on e.g. LUMI, with and without GPU-aware MPI? If yours go through fine I think we can ignore it.

anmrde:
@samhatfield I haven't worked on GPU aware MPI since Summit and neither the barrier nor the async existed in the version that I had on Summit.

samhatfield:
samhatfield/fix_gpu_aware_mpi rebased onto new version of redgreengpu.

samhatfield:
This PR can be closed BUT note that this only fixed GPU-aware MPI for the simplified benchmark, ectrans-benchmark, not the IFS-like one ectrans-benchmark-ifs. We discovered this after doing some more testing. Expect another PR in future with further tweaks to GPU-aware MPI.

PR #40, opened 2023-09-08
Propagate OpenMP flags and remove MPI option [replaces #33]

wdeconinck:
This is an alternative PR to #33 which removes the need for the MPI option.
The "GPU_AWARE_MPI" option itself will find MPI if needed.
Some fixes have been applied to allow the benchmark with gpu to run without MPI in a 'srun' context.


wdeconinck:
It would be even better if MPI is not needed at all, and MPI calls can be delegated (and implemented if needed) to fiat's MPL. That should not block this PR for now and can be done later.

marsdeno:
> 

I hope to have this done for the redgreen-opt branch I'm working on, so don't let that hold this PR up.

wdeconinck:
> > It would be even better if MPI is not needed at all, and MPI calls can be delegated (and implemented if needed) to fiat's MPL. That should not block this PR for now and can be done later.
> 
> I hope to have this done for the redgreen-opt branch I'm working on, so don't let that hold this PR up.

If you mean that the redgreen-opt branch will use MPL instead of MPI that's perfect! Indeed I think if you (@marsdeno and @reuterbal) agree, this can go into the amdgpu and redgreen (non-opt) branch.

reuterbal (approved):
This is great and much better than my previous attempt. GTG for both amdgpu and redgreen!

wdeconinck:
FYI @reuterbal, @marsdeno , I have rebased redgreengpu on top of updated amdgpu and force-pushed. Please update history carefully.

PR #42, opened 2023-10-09
Add missing variables to OpenACC directive

samhatfield:
DEFAULT(NONE) is specified so these loop variables must be declared as private.


samhatfield:
@wdeconinck - I hope this is uncontroversial? Could you merge it?

PR #51, opened 2023-11-29
Delete orphaned variables IBUFLENS and IBUFLENR

samhatfield:
These variables were used for determining whether to allocate ZCOMBUFS and ZCOMBUFR back before TRGTOL was split into TRGTOL_PROLOG and TRGTOL_COMM. Now they have no purpose.


marsdeno:
Looks good to me.

wdeconinck (approved):

PR #52, opened 2023-12-01
CPU/GPU ectrans benchmark with precision selected at runtime.

piotrows:
Extension of PR #45, but based on redgreengpu branch. CPU and GPU with legacy benchmarks, each benchmark compiles with SP and DP version of the library. Only works with static linking.


samhatfield:
Closed as no longer actively under development.

PR #55, opened 2024-01-18
Preprocessor-free implementation of single binary for sp/dp versions

piotrows:
This draft PR aims at demonstration of the possible technical solution to the challenge of selecting precision of computations at runtime.
Main code modifications are:

1) Renaming of JPRB to a desired precision at the library creation time (thus exploiting exisiting cmake ${prec} definition}  via PRIVATE_DEFINITIONS. Expecting removal of dependence on PARKIND1 module, it is renamed to EC_PARKIND the same way.

2) Creation of  stub ectrans module that imports actual routines from trans/external as submodules. aiming at modern Fortran importing. The ectrans module relies on the existing interface headers of trans/include/ectrans, slightly preprocessed to offer "MODULE PROCEDURE" if needed.

3) The benchmark program is augmented with a set of wrappers to setup_trans(0), dir_trans, inv_trans etc. that are imported by a a benchmark program instead of their legacy trans/external counterparts. The latter (apart from specnorm) are not even compiled directly, but only when embedded into ectrans_mod submodules. The wrappers are equipped with an extra argument that is being resolved at the wrapping level, so the original trans external routines are not modified in this respect. At the moment, the wrappers sit in the program directory to minimise mixing with the legacy trans code.

4) Minor technical adaptations, namely introducing JPRC to serve in assignments like a=1.0_JPRC were introduced to satisfy the needs of the NVHPC preprocessor, which apparently can't identify macro names being a part of such statement(s). On the occasion, JPRB was removed when imported but not used in several files.

5) Code specialisation is realised by cmake function **generate_trans_build** that:
     a) creates directory structure to host processed files mimicking the source tree
     b) executes **define_code_spec** cmake function that performs _sed_ edit on the source files using sed script **include/ectrans/sedrenames.txt** that annotates module names with the description of a desired variant, eg. _dp or _sp.
     c) creates modified source files list for add_library by prepending the name with path to the specialised source code tree in CMAKE_BINARY_DIRECTORY
     d) propagates modified source file list name to parent scope
6) Note that the sed script itself is edited to specify the designated variant (eg. _sp or _dp). 
7) The source code was sorted using extra subdirectories to separate fully precision-independent modules that do not need specialisation. They could potentially be cast in a separate "common" module.


piotrows:
The auxiliary construction that provides transparency of the new solution to the legacy code seems to me a bit complicated. For a given trans/external file "a", we have programs/wrappers/a.F90, a.inc; src/trans/modules/a_mod.F90 in addition to existing trans/external/a and trans/include/ectrans/a.h .This means that five files need to be synchronized and interface is actually a duplicated code. Debugging is not perhaps very hard, but still could be a bit confusing. If this solution is here to stay, perhaps a small  file generator could be helpful to keep the files in sync.

PR #57, opened 2024-01-25
Fix bug in PSPSC3A inverse Legendre transform packing

samhatfield:
Previously each field of PSPSC3A after the first was overwriting the previous one, and the packing index (IFIRST) wasn't advancing as each field was packed.


wdeconinck:
Nice catch, is this not a commonly used option?

samhatfield:
All credit to Nils! No I guess we don't usually call with multiple 3D scalar fields.

wdeconinck:
@samhatfield Can you also apply this change to the "develop" branch's CPU code? This will then make its way to the cpu version here automatically as well when merged.

samhatfield:
I don't think this bug affects the CPU version? https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/internal/ltinv_mod.F90#L180

wdeconinck:
Do we have a test to catch this bug?

wdeconinck (approved):

PR #58, opened 2024-01-29
Fix LDFOU2 loop

samhatfield:
The loop indices were missing from the PRIVATE list. Somehow this only started generating errors on LUMI-G today...

While there I took the chance to:

- Delete some unused variables.
- Consolidate the accelerated regions to reduce the amount of preprocessor interruption when reading the code.
- Replace the unfinished OpenMP offload declaration with an abort statement.
- Touch up the code a bit with whitespace to make it easier to read.


marsdeno (approved):
Sorry, had missed this. Looks fine to me

PR #60, opened 2024-02-07
Synchronise with CY49R1.0

samhatfield:
A number of commits were made in the past year or two to the contrib/ectrans directory of ifs-source. Some of these have already been cherry picked to develop here. This commit brings over the remaining differences.

Note that the MétéoFrance copyright statements are not always the same between this branch and contrib/ectrans. I have ignored differences of the form

```
! (C) Copyright 2000- Meteo-France. -> ! (C) Copyright 2013- Meteo-France.
```


samhatfield:
I will check this hasn't been left behind in my commit.

samhatfield:
All ft*_ctl*_mod.F90 files are in sync so there's no `DEALLOCATE` in the original either. Is it not the case that `ZGTF_HEAP` will be automatically deallocated when it goes out of scope?

reuterbal:
Right, sorry, I overlooked this!

PR #61, opened 2024-02-08
Unique symbols per precision

wdeconinck:
The libraries trans_sp and trans_dp have duplicated symbols.
This PR uses some kind of preprocessing of original files to generate files with a precision suffix like `_sp` and `_dp`, avoiding the same symbol names, so that in the future a library could link to both trans_sp and trans_dp.

The work in this PR is inherited from PR #55 by @piotrows, and tidied up a bit, but does not yet go so far to introduce wrappers that make a unified interface.

Note, commit 5329c07203270b972e75e14f9d0c5aece9687641 in this PR changes the API of setup_trans and setup_trans0 to use arguments with KIND=JPRD only.  This could be reverted and/or added in a follow-up PR if desired.

When this PR is merged and a new release is to be used in the IFS, an extra commit in ifs-source is required, which I have prepared and would be ready for review. I will update this Issue with ifs-source status.
ifs-source was using some ectrans internal modules or subroutines without including the interface.
So even without API change in setup_trans, the ifs-source commit would be required.


wdeconinck:
Note, preprocessing is based on `sed`. On macOS 13 this works but on the CI it does not work yet.
Either a workaround will be needed before merging, or we update the macOS runner. Further investigation needed.

src/trans/algor/interpol_decomp_mod.F90
```diff
@@ -23,7 +23,7 @@ MODULE INTERPOL_DECOMP_MOD
 ! Author: Mats Hamrud
 
 
-USE PARKIND1, ONLY : JPRB, JPIM, JPRD, JPIB
+USE EC_PARKIND, ONLY : JPIM, JPRD, JPIB
```
samhatfield, line 26: What's the purpose of `EC_PARKIND`?

wdeconinck: Excellent question. 
ec_parkind is part of fiat and does not contain JPRB so it shows this routine or module is precision independent and does not need to link to library parkind_sp or parkind_dp.

src/trans/internal/abort_trans_mod.F90
```diff
@@ -13,8 +13,7 @@ MODULE ABORT_TRANS_MOD
 SUBROUTINE ABORT_TRANS(CDTEXT)
 
 USE TPM_GEN  , ONLY : NOUT,NERR
-USE TPM_DISTR, ONLY : NPROC,MYPROC
```
samhatfield, line 16: This makes me wonder if we could remove all instances of `NPROC`, `MYPROC` from ecTrans.

src/trans/internal/dealloc_resol_mod.F90
```diff
@@ -41,7 +41,7 @@ SUBROUTINE DEALLOC_RESOL(KRESOL)
 
 !     ------------------------------------------------------------------
 
-USE PARKIND1  ,ONLY : JPIM     ,JPRB
+USE PARKIND1  ,ONLY : JPIM
```
samhatfield, line 44: What about using EC_PARKIND here?

src/trans/internal/dist_grid_32_ctl_mod.F90
```diff
@@ -43,7 +43,7 @@ SUBROUTINE DIST_GRID_32_CTL(PGPG,KFDISTG,KPROMA,KFROM,PGP)
 
 !     ------------------------------------------------------------------
 
-USE PARKIND1  ,ONLY : JPIM     ,JPRB  ,JPRM
+USE PARKIND1  ,ONLY : JPIM     ,JPRM
```
samhatfield, line 46: And here.

src/trans/internal/eq_regions_mod.F90
```diff
@@ -81,7 +81,7 @@ MODULE eq_regions_mod
 PUBLIC eq_regions,l_regions_debug,n_regions_ns,n_regions_ew,n_regions,my_region_ns,my_region_ew
 PUBLIC eq_regions_t, eq_regions_save, eq_regions_load, eq_regions_free
 
-real(kind=jprb) pi
+real(kind=jprd) pi
```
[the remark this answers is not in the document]
samhatfield: Should sort out the case (make upper) in this file at some point too.

src/trans/internal/field_split_mod.F90
```diff
@@ -58,7 +58,7 @@ SUBROUTINE FIELD_SPLIT(KBLK,KF_GP,KKF_UV_G,KVSETUV,KVSETSC,&
 !        Original : 01-01-03
 
 !     ------------------------------------------------------------------
-USE PARKIND1  ,ONLY : JPIM     ,JPRB
+USE PARKIND1  ,ONLY : JPIM
```
samhatfield, line 61: EC_PARKIND here?

src/trans/internal/inigptr_mod.F90
```diff
@@ -14,7 +14,7 @@ SUBROUTINE INIGPTR(KGPTRSEND,KGPTRRECV)
 
 !     Compute tables to assist GP to/from Fourier space transpositions
 
-USE PARKIND1  ,ONLY : JPIM     ,JPRB
+USE PARKIND1  ,ONLY : JPIM
```
samhatfield, line 17: EC_PARKIND here?

src/trans/external/specnorm.F90
```diff
@@ -64,11 +64,11 @@ SUBROUTINE SPECNORM(PSPEC,KVSET,KMASTER,KRESOL,PMET,PNORM)
 ! Declaration of arguments
 
 
+REAL(KIND=JPRB)             , INTENT(OUT) :: PNORM(:)
```
samhatfield, line 67: We might as well make the corresponding argument in SPNORM_CTL mandatory as well.

src/trans/internal/set_resol_mod.F90
```diff
@@ -11,7 +11,7 @@
 MODULE SET_RESOL_MOD
 CONTAINS
 SUBROUTINE SET_RESOL(KRESOL,LDSETUP)
-USE PARKIND1  ,ONLY : JPIM     ,JPRB
+USE PARKIND1  ,ONLY : JPIM
```
samhatfield, line 14: EC_PARKIND...?

src/trans/internal/setup_dims_mod.F90
```diff
@@ -12,7 +12,7 @@ MODULE SETUP_DIMS_MOD
 CONTAINS
 SUBROUTINE SETUP_DIMS
 
-USE PARKIND1  ,ONLY : JPIM     ,JPRB
+USE PARKIND1  ,ONLY : JPIM
```
samhatfield, line 15: EC_PARKIND...?

src/trans/internal/shuffle_mod.F90
```diff
@@ -55,7 +55,7 @@ SUBROUTINE SHUFFLE(KF_UV_G,KF_SCALARS_G,KSHFUV_G,KIVSETUV,KSHFSC_G,KIVSETSC,&
 !        Original : 01-01-03
 
 !     ------------------------------------------------------------------
-USE PARKIND1  ,ONLY : JPIM     ,JPRB
+USE PARKIND1  ,ONLY : JPIM
```
samhatfield, line 58: EC_PARKIND...?

src/trans/internal/sump_trans_mod.F90
```diff
@@ -17,7 +17,7 @@ SUBROUTINE SUMP_TRANS
 ! Modifications :
 ! P.Marguinaud : 11-Sep-2012 : Fix twice allocated pointer
 
-USE PARKIND1  ,ONLY : JPIM     ,JPRB,  JPRD
+USE PARKIND1  ,ONLY : JPIM     ,JPRD
```
samhatfield, line 20: EC_PARKIND?

src/trans/internal/sugaw_mod.F90
```diff
@@ -108,6 +107,7 @@ SUBROUTINE SUGAW(KDGL,KM,KN,PL,PW,PANM,PFN)
 
 ! computations in extended precision for alternative root finding
 ! which also works for associated polynomials (m>0)
+INTEGER, PARAMETER :: JPRH = JPRD
```
samhatfield, line 110: This one I need to think about a bit. It looks like we are removing [the option to compute the roots in extended precision](https://github.com/ecmwf-ifs/fiat/blob/main/src/parkind/parkind2.F90#L26). If we don't need this anymore, and double precision is fine, better to be transparent and just replace all instances of JPRH with JPRD.

src/trans/sedrenames.txt
```diff
@@ -0,0 +1,147 @@
...
+s/INTERPOL_DECOMP_MOD/INTERPOL_DECOMP_MOD_VARIANTDESIGNATOR/g
+s/INV_TRANS_CTL_MOD/INV_TRANS_CTL_MOD_VARIANTDESIGNATOR/g
+s/INV_TRANS_CTLAD_MOD/INV_TRANS_CTLAD_MOD_VARIANTDESIGNATOR/g
+s/inv_trans( *($|\(| |\*))/inv_trans_VARIANTDESIGNATOR\1/g
+s/INV_TRANS( *($|\(| |\*))/INV_TRANS_VARIANTDESIGNATOR\1/g
+s/inv_transad( *($|\(| |\*))/inv_transad_VARIANTDESIGNATOR\1/g
+s/INV_TRANSAD/INV_TRANSAD_VARIANTDESIGNATOR/g
+s/jprb/TYPEDESIGNATOR_LOWER/g
+s/JPRB/TYPEDESIGNATOR_UPPER/g
+s/JPRH/JPRD/g
```
samhatfield, line 69: This line confuses me a bit. With this sed, [this line](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/internal/suleg_mod.F90#L20C1-L20C35):
```
USE PARKIND2          ,ONLY : JPRH
```
will become this line
```
USE PARKIND2          ,ONLY : JPRD
```
which should not compile because `PARKIND2` does not contain `JPRD`. What am I missing, question for @piotrows I guess?

samhatfield: And as before, we need to decide first if we are happy to give up on the option for extended precision. The `REALHUGE` Fiat CPP definition appears to be defunct, so I guess we can go ahead.

samhatfield:
First read over completed. Here is my summary:
- Where possible, PARKIND1 replaced with EC_PARKIND
- PWEIGHT and PSTRET arguments to SETUP_TRANS and corresponding module variables in TPM_DISTR, TPM_GEOMETRY made double precision always
- PRAD argument to SETUP_TRANS0 made double precision always
- SPECNORM PNORM argument made mandatory - **Why is this necessary?**
- ectrans_common library created based on subset of JPRB-agnostic modules
- ectrans_{sp/dp} "backend libraries" created with JPRB-dependent modules

I'd like some more time to read the new CMake functionality for generating the ectrans_${prec} libraries, but feel free to respond to the above comments in the mean time.

src/trans/internal/abort_trans_mod.F90
wdeconinck: Yes I am not entirely sure. We can ask Mats. For an abort it is better with MPL. This was necessary to allow it to be part of ectrans_common so that other routines in this lib could use it.

src/trans/internal/dealloc_resol_mod.F90
samhatfield: Perhaps ignore my other comments then and we leave this for another PR.

src/trans/internal/sugaw_mod.F90
wdeconinck: Yes I don't think we need JPRH. At least parkind1 has not recently compiled JPRH differently.
If agreed by others as well we can follow your proposal.

src/trans/sedrenames.txt
wdeconinck: > This line confuses me a bit. With this sed, [this line](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/internal/suleg_mod.F90#L20C1-L20C35):
> 
> ```
> 
> USE PARKIND2          ,ONLY : JPRH
> 
> ```
> 
> will become this line
> 
> ```
> 
> USE PARKIND2          ,ONLY : JPRD
> 
> ```
> 
> which should not compile because `PARKIND2` does not contain `JPRD`. What am I missing, question for @piotrows I guess?

It works because SED will replace both PARKIND1 and PARKIND2 with EC_PARKIND

wdeconinck:
> * Where possible, PARKIND1 replaced with EC_PARKIND

I only applied this to some routines, but we could push this through everywhere JPRB is not needed.

> * PWEIGHT and PSTRET arguments to SETUP_TRANS and corresponding module variables in TPM_DISTR, TPM_GEOMETRY made double precision always

Correct. These are arguments that are only needed to set up the grid point domain decomposition.

> * PRAD argument to SETUP_TRANS0 made double precision always

Yes, this makes TPM_CONSTANTS module independent of JPRB.

> * SPECNORM PNORM argument made mandatory - **Why is this necessary?**

It is in fact not necessary for this PR, but disambiguates overload on this argument's KIND of this routine in the future. All optional arguments cannot be disambiguated. In practice this argument is always required. Otherwise what's the point. Trying this change in ifs-source proved that this argument is never omitted. This change could be removed from this PR and introduced later, but it's pretty harmless.

> * ectrans_common library created based on subset of JPRB-agnostic modules

It is also enlightening to see which routines and modules are not used for the actual transforms.

> * ectrans_{sp/dp} "backend libraries" created with JPRB-dependent modules

> I'd like some more time to read the new CMake functionality for generating the ectrans_${prec} libraries, but feel free to respond to the above comments in the mean time.

Noteworthy is that neither backend libraries ectrans_sp and ectrans_dp link with parkind_sp or parkind_dp because the preprocessor step replaces parkind1 with ec_parkind and JPRB with JPRD or JPRM.

src/trans/external/specnorm.F90
wdeconinck: done

wdeconinck:
Rebased on recently merged CY49R1 sync branch

wdeconinck:
I added commit [f77550c](https://github.com/ecmwf-ifs/ectrans/commit/f77550c15d5672560de0a9f5f0daf9fc19bee9c8) to develop branch which updates the macOS runner to macos-13. Having rebased this branch on develop now fixes the 'sed' issues previously present.

src/trans/internal/sugaw_mod.F90
samhatfield: I can just create an issue so we deal with this in a future PR.

src/trans/internal/trgtol_mod.F90
```diff
@@ -472,7 +472,7 @@ SUBROUTINE TRGTOL_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
 INTEGER(KIND=JPIM) :: ISEND, ITAG, JBLK, JFLD, JK, JL, IFLD, II, IFLDS, INS, INR
 INTEGER(KIND=JPIM) :: JJ,JI,IFLDT, J
 
-INTEGER(KIND=JPIA) :: JFLD64
+INTEGER(KIND=JPIB) :: JFLD64
```
samhatfield, line 475: This variable was introduced in my CY49R1 [sync commit](https://github.com/ecmwf-ifs/ectrans/commit/635f9bb1c36c40945077e8f7e1ca68649ad794e6#diff-d50752e10eeb0069e066a6ac0bd9fea3ed72921d7b40aa07cfb5edca47d6efd2R475). I think @RyadElKhatibMF picked `JPIA` for a reason but not sure why. Can we keep it?

src/trans/sedrenames.txt
samhatfield: Oh I see, gotcha.

RyadElKhatibMF:
JPIB is better. 
JPIA has become too ambiguous. 

----- Météo-France ----- 
EL KHATIB Ryad 
DESR/CNRM/GMAP/ALGO 
***@***.*** 
Fixe : +33 561078466

src/trans/CMakeLists.txt
piotrows, line 1: Why do we pass setup_trans0.h through configure? Is it meant to "do nothing" in this case?

src/trans/internal/sugaw_mod.F90
wdeconinck: Thanks @samhatfield for creating #62 for follow-up.

src/trans/CMakeLists.txt
wdeconinck: Good observation! I used this to copy files from source dir to build-dir, and indeed in this case changes nothing of the content.

samhatfield:
I finally read through the CMake changes and that looks good to me! I think I'm starting to like CMake.

Can `internal/sutrle_mod.F90` be added to the common library?

Also `external/setup_trans.F90`?

wdeconinck:
> I finally read through the CMake changes and that looks good to me! I think I'm starting to like CMake.
> 
> Can `internal/sutrle_mod.F90` be added to the common library?
> 
> Also `external/setup_trans.F90`?

Hi @samhatfield these files unfortunately cannot be added to the common library because they `USE` modules that have themselves JPRB. e.g. `TPM_FIELDS`

samhatfield:
Good to merge for me.

marsdeno (approved):
This looks good to me.

wdeconinck:
Merged in develop now! Thanks everyone, especially @piotrows for the majority of all work included in this PR. 🙏

src/trans/CMakeLists.txt
```diff
@@ -32,41 +32,207 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
...
+      get_filename_component(outfile_name_we ${file_i} NAME_WE)
+      get_filename_component(outfile_ext     ${file_i} EXT)
+      get_filename_component(outfile_dir     ${file_i} DIRECTORY)
+      set(outfile "${destination}/${file_i}")
+      ecbuild_debug("Generate ${outfile}")
+      generate_file(BACKEND ${backend} INPUT ${file_i} OUTPUT ${outfile})
+      list(APPEND outfiles ${outfile})
+    endif()
+  endforeach(file_i)
+  set(${_PAR_OUTPUT} ${outfiles} PARENT_SCOPE)
```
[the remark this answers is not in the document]
wdeconinck: I wasn't aware of `return(PROPAGATE)`. It is very neat but I see it is only available since cmake 3.25 which is too recent to be a requirement. This is a known idiom. Perhaps in the distant future this may be cleaned up in the entire stack, if it's worth it as it is only syntactic.

src/trans/CMakeLists.txt
```diff
@@ -32,41 +32,207 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
...
+    QUIET
+  )
+  if( NOT HAVE_FFTW )
+    ecbuild_list_exclude_pattern( LIST files REGEX tpm_fftw.F90 )
+  endif()
+
+
+  set(outfiles)
+  foreach(file_i ${files})
+    if(NOT (${file_i} IN_LIST ectrans_common_src))
```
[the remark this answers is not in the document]
wdeconinck: Too much boiler plate code for something that takes no time at all.

src/trans/CMakeLists.txt
```diff
@@ -32,41 +32,207 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
...
+  foreach(file_i ${files})
+    get_filename_component(outfile_name    ${file_i} NAME)
+    get_filename_component(outfile_name_we ${file_i} NAME_WE)
+    get_filename_component(outfile_ext     ${file_i} EXT)
+    get_filename_component(outfile_dir     ${file_i} DIRECTORY)
+    if (${file_i} IN_LIST ectrans_common_includes)
+      configure_file(${file_i} ${destination}/${outfile_name})
+    else()
+      set(outfile "${destination}/${outfile_name_we}_${backend}${outfile_ext}")
+      ecbuild_debug("Generate ${outfile}")
```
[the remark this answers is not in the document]
wdeconinck: I don't think that is a problem, this is also just an ecbuild debug statement.

src/trans/CMakeLists.txt
```diff
@@ -32,41 +32,207 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
...
     ecbuild_add_library(
-      TARGET           trans_${prec}
+      TARGET           ectrans_${prec}
       LINKER_LANGUAGE  Fortran
-      SOURCES          ${trans_src}
-      PUBLIC_INCLUDES  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/trans/include>
-                       $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/trans/include/ectrans>
-                       $<INSTALL_INTERFACE:include/ectrans>
+      SOURCES          ${ectrans_${prec}_src}
+      PUBLIC_INCLUDES  $<INSTALL_INTERFACE:include/ectrans>
```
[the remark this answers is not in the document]
wdeconinck: Yes it is still valid as the "common" includes should be there.

src/trans/CMakeLists.txt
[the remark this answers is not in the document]
piotrows: Sure, I didn't notice that is 3.25 up.

src/trans/CMakeLists.txt
```diff
@@ -32,41 +32,207 @@ if(CMAKE_Fortran_COMPILER_ID MATCHES "Intel")
...
+  set (options)
+  set (oneValueArgs BACKEND TARGET DESTINATION)
+  set (multiValueArgs)
+  cmake_parse_arguments(_PAR "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
+
+  set(destination ${_PAR_DESTINATION} )
+  set(backend     ${_PAR_BACKEND})
+
+  file(MAKE_DIRECTORY ${destination})
+  file(MAKE_DIRECTORY ${destination}/${backend})
```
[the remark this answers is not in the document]
wdeconinck: You are right, something fishy here, that should be 
```cmake
file(MAKE_DIRECTORY ${destination}/trans_${backend})
```
This is now added to develop with commit 91a51b1dbca7e0cb5b9ba789f6c2c24be4774d88

PR #63, opened 2024-02-20
Propagate pinned flag 

reuterbal:
To make use of pinned memory allocations, we add the `-gpu=pinned` flag for NVHPC to setup_trans.F90:

https://github.com/ecmwf-ifs/ectrans/blob/e454b3e01fdf585008a3f2921524b0ad569e4bb1/src/trans/gpu/CMakeLists.txt#L17-L20

For this to be effective, we also need to make sure this flag ends up on the link lines of all targets that depend on ectrans. This PR propagates the flag as part of the build interface for NVHPC.


wdeconinck (approved):
Looks good to me!

samhatfield (approved):

PR #65, opened 2024-02-29
Create BLAS interfaces for GEMM and GEMV

wdeconinck:
This PR creates a module "ectrans_blas_mod" with interfaces GEMM and GEMV, which delegate calls to the correct BLAS function DGEMM/SGEMM and DGEMV/SGEMV.

With this in many places constructs like
```fortran
IF( LLDOUBLE ) THEN
  CALL DGEMM( ... JPRD ... )
ELSE
  CALL SGEMM( ... JPRM ... )
ENDIF
```
can be replaced with a one liner where possible:
```fortran
CALL GEMM( ... JPRB ... )
```

For SGEMM it also encapsulates the IEEE floating point handling, which was copy-pasted in various places.

Also, and why I started this in the first place, is that now all the argument-mismatch warnings with gfortran > 10 are eradicated.


src/trans/internal/leinv_mod.F90
```diff
@@ -138,19 +130,7 @@ SUBROUTINE LEINV(KM,KMLOC,KFC,KIFC,KF_OUT_LT,KSL,KDGLU,PIA,PAOA1,PSOA1)
...
-    ELSE
-       IF (LL_IEEE_HALT) THEN
-          call ieee_get_halting_mode(ieee_invalid,LL_HALT_INVALID)
-          if (LL_HALT_INVALID) call ieee_set_halting_mode(ieee_invalid,.false.)
-       ENDIF
-       CALL SGEMM('N','N',KDGLU,KIFC,ILA,1.0_JPRM,S%FA(KMLOC)%RPNMA,KDGLU,&
-            &ZBA,ILA,0._JPRM,ZC,KDGLU)
-       if (LL_IEEE_HALT .and. LL_HALT_INVALID) call ieee_set_halting_mode(ieee_invalid,.true.)
-    ENDIF
-    IF (LHOOK) CALL DR_HOOK('LEINV_'//CLX//'GEMM_1',1,ZHOOK_HANDLE)
```
samhatfield, line 153: I think you need to keep this closing Dr Hook call.

samhatfield:
Apart from comment above, looks great to me! Much neater now.

marsdeno (approved):
Very nice, both tidier and *much* improved code/compilation behaviour!

src/trans/internal/leinv_mod.F90
wdeconinck: This is why an extra pair of eyes is so useful!!! Thanks!

wdeconinck: Fixed now.

Issue #73, opened 2024-03-29
FLUSH function non-standard


DJDavies2:
There are a few uses of the FLUSH intrinsic in the code.  This function is non-standard, and nowadays there is a FLUSH statement that has been standard since 2003 and is widely supported.

PR #74, opened 2024-03-29
Use FLUSH statement

DJDavies2:
See https://github.com/ecmwf-ifs/ectrans/issues/73.  This uses a flush statement instead of the intrinsic.


samhatfield:
This seems uncontroversial to me!

marsdeno (approved):

PR #76, opened 2024-03-29
Rearrange the declaration of some dummy arguments

DJDavies2:
This is because there are some dummy arguments declared like this:

  real(kind=jprd), intent(inout) :: a(n)
  integer(kind=jpim), intent(in) :: n

I think this counts as "n" being used before it is declared; the solution is to have "n" declared first, this is what this change does.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -1138,8 +1138,8 @@ end subroutine str2int
 
 subroutine sort(a, n)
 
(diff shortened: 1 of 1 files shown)


samhatfield:
Seems reasonable.

marsdeno (approved):
@DJDavies2 Out of curiosity, did you find this because a compiler objected? If so, which compiler?

DJDavies2:
Yes, it was the NAG compiler.

Issue #77, opened 2024-03-29
Failure to compile SIZEOF (NAG) in suleg_mod


DJDavies2:
NAG compiles are failing in suleg_mod like this:

Error: /home/h01/frwd/cylc-run/mi-be984/work/1/ecbuild_ectrans_nag/build/src/trans/generated/ectrans_dp/internal/suleg_mod.F90, line 1207: Implicit type for SIZEOF in SULEG
[NAG Fortran Compiler pass 1 error termination, 1 error]
Error: /home/h01/frwd/cylc-run/mi-be984/work/1/ecbuild_ectrans_nag/build/src/trans/generated/ectrans_sp/internal/suleg_mod.F90, line 1207: Implicit type for SIZEOF in SULEG
[NAG Fortran Compiler pass 1 error termination, 1 error]

wdeconinck:
This should now be fixed with PR #89

PR #78, opened 2024-03-29
Implement SIZEOF for NAG.

DJDavies2:
For https://github.com/ecmwf-ifs/ectrans/issues/77.  This implements the same macro as for _NEC_.


samhatfield:
Looks good to me.

src/trans/internal/suleg_mod.F90
```diff
@@ -12,6 +12,9 @@ MODULE SULEG_MOD
 #ifdef __NEC__
 #define SIZEOF(x) STORAGE_SIZE(x)/KIND(x)
 #endif
+#ifdef NAG
```
marsdeno, line 15: Looking at paragraph 4.5.4 in 
https://support.nag.com/nagware/np/r72_doc/manual/compiler_4_5.html#MACRO
suggests that this should perhaps be NAGFOR instead.
Thoughts?

DJDavies2: Yes, thanks, I didn't realize nag set that.  I have updated the branch.

samhatfield:
I think instead of adding a `SIZEOF` implementation we should just refactor the code so we don't need it at all. Consider [one of the two places](https://github.com/ecmwf-ifs/ectrans/blob/d78f880652d404c708ebac3fbae451c4a7aed1f3/src/trans/internal/suleg_mod.F90#L874C1-L883C20) where we used `SIZEOF`:

```fortran
IF( .NOT.ALLOCATED(ZCLONEA(IMLOC)%COMMSBUF) )THEN
    ICLONELEN=IKOUNT(JSETV)
    ALLOCATE(ZCLONEA(IMLOC)%COMMSBUF(ICLONELEN))
    ZCLONEA(IMLOC)%COMMSBUF(1:ICLONELEN) = ZRCVBUTFV(1:ICLONELEN,JSETV)
    CALL UNPACK_BUTTERFLY_STRUCT(S%FA(IMLOC)%YBUT_STRUCT_A,ZCLONEA(IMLOC))
ENDIF
IF(ALLOCATED(ZCLONEA(IMLOC)%COMMSBUF) ) THEN
    IF( SIZEOF(ZCLONEA(IMLOC)%COMMSBUF) > 0 ) DEALLOCATE(ZCLONEA(IMLOC)%COMMSBUF)
ENDIF
```

Let me rewrite this in pseudocode:

```fortran
 IF (.NOT. ALLOCATED(ARRAY)) THEN
     ICLONELEN = IKOUNT(JSETV)
     ALLOCATE(ARRAY(ICLONELEN))
    ! Some stuff 
ENDIF
! ARRAY must be allocated so why bother having this IF?
IF (ALLOCATED(ARRAY)) THEN
    ! Even so, we don't need SIZEOF because we can just check whether ICLONELEN is > 0
    IF( SIZEOF(ZCLONEA(IMLOC)%COMMSBUF) > 0 ) DEALLOCATE(ZCLONEA(IMLOC)%COMMSBUF)
ENDIF
```

I will have a go at this refactor and submit a new PR.

wdeconinck:
Alternative PR #89 was merged that solves this issue differently

PR #87, opened 2024-04-04
Add compile flags to match those applied to ifs-source/contrib/ectrans

samhatfield:
These flags are necessary to achieve bit reproducibility with ifs-source/contrib/ectrans (at least with double precision).


samhatfield:
I guess this can be closed now (unmerged) @wdeconinck ?

PR #89, opened 2024-04-11
Remove need for non-standard SIZEOF

### LINKED ISSUE #77: Failure to compile SIZEOF (NAG) in suleg_mod

samhatfield:
Builds on #78 and resolves #77.

I remove one redundant `IF` statement (right above there is logic which ensures the condition is always `.TRUE.`.

I also remove the `SIZEOF` by simply checking the value of the integer used to determine the size of the allocated array, rather than checking how many bytes were actually allocated.


src/trans/internal/suleg_mod.F90
```diff
@@ -874,10 +871,9 @@ SUBROUTINE SULEG
                 ZCLONEA(IMLOC)%COMMSBUF(1:ICLONELEN) = ZRCVBUTFV(1:ICLONELEN,JSETV)
                 CALL UNPACK_BUTTERFLY_STRUCT(S%FA(IMLOC)%YBUT_STRUCT_A,ZCLONEA(IMLOC))
               ENDIF
-              IF(ALLOCATED(ZCLONEA(IMLOC)%COMMSBUF) ) THEN
-                IF( SIZEOF(ZCLONEA(IMLOC)%COMMSBUF) > 0 ) DEALLOCATE(ZCLONEA(IMLOC)%COMMSBUF)
-                !              ZCLONEA(IMLOC)%COMMSBUF=>NULL()
-              ENDIF
+
+              IF( ICLONELEN > 0 ) DEALLOCATE(ZCLONEA(IMLOC)%COMMSBUF)
+
```
wdeconinck, line 876: I think that ICLONELEN could be unininitialized in case
`IF( .NOT.ALLOCATED(ZCLONEA(IMLOC)%COMMSBUF) )THEN` (line 868) is not triggered.
Probably the `IF(ALLOCATED(ZCLONEA(IMLOC)%COMMSBUF) ) THEN` was still correct.
Could it not also be that it needs to stay allocated across iterations? Otherwise why all the guards (line 868).

src/trans/internal/suleg_mod.F90
```diff
@@ -1157,10 +1153,9 @@ SUBROUTINE SULEG
                 ZCLONES(IMLOC)%COMMSBUF(1:ICLONELEN) = ZRCVBUTFV(1:ICLONELEN,JSETV)
                 CALL UNPACK_BUTTERFLY_STRUCT( S%FA(IMLOC)%YBUT_STRUCT_S,ZCLONES(IMLOC))
               ENDIF
-              IF( ALLOCATED(ZCLONES(IMLOC)%COMMSBUF) ) THEN
-                IF( SIZEOF(ZCLONES(IMLOC)%COMMSBUF) > 0 ) DEALLOCATE(ZCLONES(IMLOC)%COMMSBUF)
-                !              ZCLONES(IMLOC)%COMMSBUF=>NULL()
-              ENDIF
+
+              IF( ICLONELEN > 0 ) DEALLOCATE(ZCLONES(IMLOC)%COMMSBUF)
```
DJDavies2, line 1157: Ditto

DJDavies2:
I've tested here and it seems to work for me.

src/trans/internal/suleg_mod.F90
samhatfield: Good point @wdeconinck - we shouldn't use `ICLONELEN`.

The first `IF` has no purpose at all, as far as I can see. I can't see how it's possible for `ZCLONEA(IMLOC)%COMMSBUF` to not be allocated at line 877.

Instead of checking if `ICLONELEN > 0`, why don't we check `STORAGE_SIZE(ZCLONEA(IMLOC)%COMMSBUF) > 0`? `STORAGE_SIZE` is the same as `SIZEOF` but returns the bits allocated rather than bytes, and it appears to be part of the Fortran 2008 standard, unlike `SIZEOF`.

samhatfield:
@DJDavies2 - could you test the latest commit with NAG Fortran?

DJDavies2:
This branch seems to work for me as far as it goes.

DJDavies2:
I'm down as a reviewer for this but I don't think I have the permissions to do that?  I am happy to "approve" this.

wdeconinck (approved):
OK for me. Nice cleanup

PR #91, opened 2024-04-19
Set `LINKER_LANGUAGE` for benchmark executable

awnawab:
This PR comprises of a small fix needed for static IFS builds. As I understand it, the `develop` branch is no longer bit-identical with cy49 in single-precision. Can I therefore please request this commit be tagged as 1.3.2, rather than merging into `develop` and tagging the head? Many thanks 🙏


wdeconinck:
Thanks @awnawab , this is now released in version 1.3.2 as hotfix, cherry-picked from this branch.
I am now deleting this branch.

PR #97, opened 2024-05-08
General mostly performance-neutral improvements

samhatfield:
This PR will collect suggested changes that came as a result of the deep-dive into ecTrans with Nvidia colleagues.


samhatfield:
I think we might as well get this merged before it grows too big.

wdeconinck (approved):
This looks like all great improvements both in code and documentation.

marsdeno (approved):
Looks all good to me

Issue #95, opened 2024-04-30
Runtime failure in DIST_SPEC_CONTROL


DJDavies2:
I have a program which is failing without any sort of error message.  Looking through a core dump it appears to be failing in the call to DIST_SPEC_CONTROL, and this is confirmed by adding write statements.  I know this isn't much to go on, it will be hard for me to produce something to reproduce the problem.

The compiler is gnu 12 on a Cray.  It is related to https://github.com/JCSDA-internal/saber/issues/821.

samhatfield:
Hi @DJDavies2 - thanks for bringing this to our attention. Do you know which line the failure occurs on? Is it a segfault?

Also the link you provided seems to be to a private repository.

DJDavies2:
The line number I think is the call itself to DIST_SPEC_CONTROL from DIST_SPEC; write statements seem to indicate that the code gets as far as the call in DIST_SPEC but doesn't start DIST_SPEC_CONTROL.  I can do more checks though.

Sorry, yes the repo linked to is private, I have no control over that.

samhatfield:
Okay, let us know if you find out anything more. I guess there's nothing we can do on our side until we have a reproducer.

DJDavies2:
Yes, thanks.

Issue #93, opened 2024-04-27
Compile failure for Intel


DJDavies2:
I am getting this compile failure with ifort for the develop branch:

/home/h01/frwd/cylc-run/mo-bundle-build-base-latest/share/make_mo__spice_intel_debug/ectrans/src/trans/generated/ectrans_dp/internal/ftdir_ctl_mod.F90(147): error #5558: A pointer with the CONTIGUOUS attributes is being made to a non-contiguous target
  ZGTF => ZGTF_STACK(:,:)
--^
compilation aborted for /home/h01/frwd/cylc-run/mo-bundle-build-base-latest/share/make_mo__spice_intel_debug/ectrans/src/trans/generated/ectrans_dp/internal/ftdir_ctl_mod.F90 (code 1)
gmake[2]: *** [ectrans/src/trans/CMakeFiles/ectrans_dp.dir/generated/ectrans_dp/internal/ftdir_ctl_mod.F90.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
gmake[1]: *** [ectrans/src/trans/CMakeFiles/ectrans_dp.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....

samhatfield:
Hi @DJDavies2 - which version of ifort are you using? I am using
```
ifort (IFORT) 2021.4.0 20210910
```
and I don't see this error.

samhatfield:
You're right to raise this though. The [documentation](https://www.ibm.com/docs/en/xffbg/121.141?topic=attributes-contiguous-fortran-2008) is pretty clear:

> An array pointer with the CONTIGUOUS attribute can only be pointer associated with a contiguous target.

The `CONTIGUOUS` statements were introduced in commit 635f9bb. This commit contains changes imported from the IFS, and I think @RyadElKhatibMF wrote this originally. I noticed that `ZGTF` in `FTDIR_CTL` has `CONTIGUOUS` but not `ZGTF` in `FTINV_CTL`. So I wonder if we actually need `CONTIGUOUS` at all. Any thoughts @RyadElKhatibMF?

DJDavies2:
I am using Ifort 2019, a bit old unfortunately.

RyadElKhatibMF:
Yes, I am used to signing my changes in the source code itself ;-) 
[...]

PR #98, opened 2024-05-10
Remove CONTIGUOUS from dist_spec_control_argument

DJDavies2:
I am not entirely sure what is happening here, but this change does seem to fix the issue at least for the failing job in question.  The change just removes the contiguous attribute from one of the arguments (PSPECG).  The other argument with the contiguous attribute, PSPEC, seems to be fine, but I don't know why there would be a difference.  The only obvious difference is the intent.


samhatfield:
These `CONTIGUOUS`s were added in the [same commit](https://github.com/ecmwf-ifs/ectrans/commit/635f9bb1c36c40945077e8f7e1ca68649ad794e6) as the other ones. I'll see what Météo-France thinks about having these removed as well.

DJDavies2:
A bit more info:

The job concerned is a 2PE job where the failure occurs on one of the PEs but not the other (the code then stops in the mpl_wait in dist_spec_control).  The PSPECG argument is present on one PE and not present on the other.  The PE that crashes is the one where the argument isn't present.  The origin of this difference is in trans_distspec (in transi_module.F90) where there are two calls to dist_spec, one which passes PSPECG and the other with doesn't.

wdeconinck:
Is this for a specific compiler only, or is this generally a problem?

DJDavies2:
I am seeing failures in different downstream systems: saber (as noted above) and also lfric-lite-jedi.  The saber failures are with gnu 9.2.  For lfric-lite-jedi the failures occur with gnu 9.2 and gnu 12.2.

It seems to work with gnu 7.3.0.  Other systems would take some time to test as they are not fully ported.

DJDavies2:
This is related to #95.

DJDavies2:
Would a possibility be to add a compile time option to have the CONTIGUOUS attribute or not?  It would make the code a bit ugly, but would allow people to have the attribute or not as works best for them.

wdeconinck:
@marsdeno please could you liaise with author(s) of the CONTIGUOUS change?

samhatfield:
@DJDavies2 do you think this is related to https://github.com/ecmwf-ifs/ectrans/issues/93?

marsdeno:
> Would a possibility be to add a compile time option to have the CONTIGUOUS attribute or not? It would make the code a bit ugly, but would allow people to have the attribute or not as works best for them.

I guess doing something like 

```
#ifdef CONTIG_BUGGY_COMPILER
#define CONTIG_STATUS ,CONTIGUOUS
#else
#define CONTIG_STATUS 
#endif

```
and then 

`REAL(KIND=JPRB) ,OPTIONAL, INTENT(IN) CONTIG_STATUS :: PSPECG(:,:)`

might be acceptable?
The downside is syntax would no longer be valid without preprocessing, but at least it would be be pretty clear what is being achieved.

wdeconinck:
So a known working compiler is gfortran 7.3.
I guess we could disable CONTIGUOUS statement here for gfortran >= 8 then.

marsdeno:
@DJDavies2 do you have a simple reproducer I could try with latest GNU on our machine to see if the issue is still there?

DJDavies2:
> @DJDavies2 do you have a simple reproducer I could try with latest GNU on our machine to see if the issue is still there?

I don't have a simple reproducer I'm afraid.  I can try to produce one if you like, but I can't guarantee I will be able to do so.

DJDavies2:
> > Would a possibility be to add a compile time option to have the CONTIGUOUS attribute or not? It would make the code a bit ugly, but would allow people to have the attribute or not as works best for them.
> 
> I guess doing something like
> 
> ```
> #ifdef CONTIG_BUGGY_COMPILER
> #define CONTIG_STATUS ,CONTIGUOUS
> #else
> #define CONTIG_STATUS 
> #endif
> ```
> 
> and then
> 
> `REAL(KIND=JPRB) ,OPTIONAL, INTENT(IN) CONTIG_STATUS :: PSPECG(:,:)`
> 
> might be acceptable? The downside is syntax would no longer be valid without preprocessing, but at least it would be be pretty clear what is being achieved.

I was thinking more like:

#ifdef CONTIG_BUGGY_COMPILER
REAL(KIND=JPRB), OPTIONAL, INTENT(IN) :: PSPECG(:,:)
#else
REAL(KIND=JPRB), OPTIONAL, INTENT(IN), CONTIGUOUS :: PSPECG(:,:)
#endif

but either way would be fine.  The file already ends in .F90 so I think it is already being "pre-processed" even if their aren't any cpp statements at the moment.

DJDavies2:
> @DJDavies2 do you think this is related to #93?

It is a different part of the code and a different compiler so I don't know.how connected these two issues are.  It would certainly be helpful if #93 could be handled.

marsdeno:
> > > Would a possibility be to add a compile time option to have the CONTIGUOUS attribute or not? It would make the code a bit ugly, but would allow people to have the attribute or not as works best for them.
> > 
> > 
> > I guess doing something like
> > ```
> > #ifdef CONTIG_BUGGY_COMPILER
> > #define CONTIG_STATUS ,CONTIGUOUS
> > #else
> > #define CONTIG_STATUS 
> > #endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then
> > `REAL(KIND=JPRB) ,OPTIONAL, INTENT(IN) CONTIG_STATUS :: PSPECG(:,:)`
> > might be acceptable? The downside is syntax would no longer be valid without preprocessing, but at least it would be be pretty clear what is being achieved.
> 
> I was thinking more like:
> 
> #ifdef CONTIG_BUGGY_COMPILER REAL(KIND=JPRB), OPTIONAL, INTENT(IN) :: PSPECG(:,:) #else REAL(KIND=JPRB), OPTIONAL, INTENT(IN), CONTIGUOUS :: PSPECG(:,:) #endif
> 
> but either way would be fine. The file already ends in .F90 so I think it is already being "pre-processed" even if their aren't any cpp statements at the moment.

The advantage of defining something like CONTIG_STATUS is it could be done once, and used wherever needed. In fact it could be done in one of the CMake files.

wdeconinck:
So what's missing now is a CMake modification to add the preprocessor definition `CONTIG_BUGGY_COMPILER` for particular compilers...

DJDavies2:
> So what's missing now is a CMake modification to add the preprocessor definition `CONTIG_BUGGY_COMPILER` for particular compilers...

I wasn't quite sure how to approach this.  I don't know exactly which compilers/versions would need this option to be set and it could get messy and error-prone to try and handle it in the cmake file.  An alternative would be to just let users set the option themselves via CMAKE_Fortran_FLAGS if they need it.  What do you think?

marsdeno:
Something like 

```
if( CMAKE_Fortran_COMPILER_ID MATCHES "Intel"  )
  if( CMAKE_Fortran_COMPILER_VERSION VERSION_LESS_EQUAL 19)
    set( COMPILER_HAS_CONTIG_ISSUES True)
  endif()
elseif( CMAKE_Fortran_COMPILER_ID MATCHES "GNU"  )
  if( CMAKE_Fortran_COMPILER_VERSION VERSION_EQUAL "9.2|12.2")
    set( COMPILER_HAS_CONTIG_ISSUES True)
  endif()
endif()

```
in the main CMakeLists.txt 

and adding 
```
if( COMPILER_HAS_CONTIG_ISSUES ) 
  target_compile_definitions( ectrans_${prec} CONTIG_BUGGY_COMPILER) 
endif()
```
after the ecbuild_add_library() call in src/trans/CMakeLists.txt ?

wdeconinck:
Good @marsdeno except I'd perhaps use the variable `ECTRANS_HAS_CONTIGUOUS_ISSUE` instead.
Also let's make sure to insert a link to this Github issue in comments.
I'd also wrap the first snippet with:
`if (NOT DEFINED ECTRANS_HAS_CONTIGUOUS_ISSUE)` so that it could be overwritten (turn OFF) via a command line argument.

wdeconinck:
@DJDavies2 I applied my suggestion on top of your latest change in 
2fdc7ff

DJDavies2:
Thanks, I have tested the latest version of this branch and it seems to work for us.

wdeconinck (approved):
Great! This looks like ready to merge then.

PR #94, opened 2024-04-29
Some changes related to CONTIGUOUS

DJDavies2:
This is for #93, it seems to fix the Intel issue.

I don't know if this is acceptable as it removes the CONTIGUOUS from a dummy argument.  I think this is a problem with the compiler.

PR #99, opened 2024-05-14
Make use of CONTIGUOUS for ZGTF/PREEL more consistent

samhatfield:
Potential resolution of #93. Renders #94 unnecessary.

@DJDavies2 could you try this with your compiler? I am still not sure what to do about the dummy arguments `ZGTF` is passed to, e.g. `PGLAT` of `TRGTOL`. Maybe that argument needs to be declared as `CONTIGUOUS` as well.


samhatfield:
Workflow failed. Of course you can't put `CONTIGUOUS` on a `TARGET` array. Nevermind.

PR #101, opened 2024-05-22
Remove nspecresmin

samhatfield:
This variable serves no purpose. It is supposed to replicate the process for initialising `NPRTRW` and `NPRTRV` in the IFS (`SUMPINI`). There one can specify in the IFS namelist a maximum allowed value for `NPRTRW`, though I'm not sure the reason you would need this. In ecTrans we don't need this feature, so it's best to just delete this.

This PR was prompted by Lucian Anton at ECMWF trying to run ecTrans on 8000 tasks without specifying `--nprtrw` and `--nprtrv`. The automatic logic does not allow running with so many tasks due to the check on `nspecresmin`.


marsdeno:
Looks fine to me

marsdeno (approved):

wdeconinck (approved):

PR #104, opened 2024-06-04
Remove fft992 and Bluestein FFT functionality

samhatfield:
This makes FFTW a required package.

@marsdeno @ioanhadade - this is what it looks like without fft992 and Bluestein. It's a more invasive change than I expected!


samhatfield:
Note: API change on `SETUP_TRANS`, with removal of `LDUSEFFTW` argument.

samhatfield:
Given the recent interest in developments like https://github.com/ecmwf-ifs/ectrans/pull/100 I think we should take the plunge and merge this PR. 

Any objections @wdeconinck @marsdeno ?

marsdeno:
No objections from me, I agree we should go for it as this unblocks subsequent changes.

samhatfield:
I forgot to appoint official reviewers!

wdeconinck:
OK, I agree as well then. We could always make an adaptor library of bluestein / fft992 (if needed) in the future that follows the FFTW API, at which point.
Definitely we need a CPU ifs-bundle test with this though before merging into main.
I would request we keep the outward facing API unchanged (optional argument LDUSEFFTW) but print a deprecation message (non-severe warning) if it is provided, so that downstream packages are aware to remove this argument in the call-site at their own pace.

src/trans/external/setup_trans.F90
```diff
@@ -10,7 +10,7 @@
 
 SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
 &KTMAX,KRESOL,PWEIGHT,LDGRIDONLY,LDUSERPNM,LDKEEPRPNM,LDUSEFLT,&
-&LDSPSETUPONLY,LDPNMONLY,LDUSEFFTW,&
```
wdeconinck, line 13: As mentioned in PR comments, please leave LDUSEFFTW for the moment but print deprecation message when `present(LDUSEFFTW)`.

src/trans/include/ectrans/setup_trans.h
```diff
@@ -11,7 +11,7 @@
 INTERFACE
 SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
 &KTMAX,KRESOL,PWEIGHT,LDGRIDONLY,LDUSERPNM,LDKEEPRPNM,LDUSEFLT,&
-&LDSPSETUPONLY,LDPNMONLY,LDUSEFFTW,&
```
wdeconinck, line 14: And here to match interface.

wdeconinck (changes requested):

DJDavies2:
Just a question, I tried building this and got this cmake failure:

CMake Error at /home/h01/david.davies/cylc-run/ectrans-104-build-stack/share/mobbs/util/exz_gnu/share/ecbuild/cmake/ecbuild_log.cmake:190 (message):
  [1;35mCRITICAL - ectrans FAILED to find REQUIRED package FFTW

  find_package(FFTW) used a FindFFTW module to find FFTW

    Please check file `/home/h01/david.davies/cylc-run/ectrans-104-build-stack/share/mobbs/util/exz_gnu/share/ecbuild/cmake/FindFFTW.cmake` for help on setting variables to find this package.[m
Call Stack (most recent call first):
  /home/h01/david.davies/cylc-run/ectrans-104-build-stack/share/mobbs/util/exz_gnu/share/ecbuild/cmake/ecbuild_find_package.cmake:342 (ecbuild_critical)
  ectrans/CMakeLists.txt:58 (ecbuild_find_package)

It looks like there is a change of dependencies regarding fftw?  Is this correct?

samhatfield:
That is correct. FFTW would become a mandatory feature if this is merged. Does this error mean that up until now you have not been using FFTW with ecTrans? If so that's valuable feedback! We were under the assumption that everybody uses FFTW.

DJDavies2:
I hadn't really thought about it.  Most of our platforms have MKL and the FFTW is picked up from there, but one of our platforms doesn't have MKL so we will have to do something else.  It isn't a problem, we will sort something out.

wdeconinck:
@DJDavies2 , FFTW (or MKL's implementation) is really improving performance quite a bit for most of the supported grids. Perhaps it is good that this was caught. 
As usual, also many thanks for testing the PRs on more architectures and configurations!

wdeconinck (approved):
Thanks @samhatfield !

PR #105, opened 2024-06-05
Make GEM? calls in FLT subroutines precision agnostic

samhatfield:
I noticed we're not making use of the new precision-agnostic GEMM interface layer in `BUTTERFLY_ALG_MOD`.


samhatfield:
Woops! I forgot I already made this change in https://github.com/ecmwf-ifs/ectrans/pull/97.

PR #106, opened 2024-06-13
Merge redgreen-optimized into develop

samhatfield:
This will bring GPU capability into the develop branch.

@wdeconinck and I discussed again the issue of how to do this merge yesterday. As @wdeconinck pointed out before, if we just merge develop into redgreen-optimized before submitting a PR of the latter, we should be able to do the merge in one go rather than first preparing develop with the cpu/gpu subdirectory structure. That's what I've tried here. It seems to have worked, so if we merge this PR we no longer need https://github.com/ecmwf-ifs/ectrans/pull/103.

This is a draft as I want to carry out tests on both cpu and gpu versions first. According to Meld there are (literally) no differences between src/trans in develop and src/trans/cpu in rgo-develop. So I think this PR shouldn't screw anything up on the CPU side at least.

Note that this branch is still not stable on AMD platforms due to an outstanding issue with HIPFFT. Solving this is on the todo list. In the meantime, for AMD support you have to fall back on redgreengpu, which is the unoptimised but more portable branch.


src/programs/ectrans-benchmark.F90
```diff
@@ -581,16 +585,22 @@ program transform_test
...
 ztloop = timef()
 
 !===================================================================================================
 ! Do spectral transform loop
 !===================================================================================================
 
-do jstep = 1, iters
+gstats_lstats = .false.
+
+do jstep = 1, iters+2
```
samhatfield, line 601: @marsdeno any ideas why your branch does 2 extra iterations?

lukasm91: hope it is fine if I also put some comments :) see the line below, the idea is to turn on gstats in the thrid timestep to exclude the setup time. creating the fft plans can be expensive and is not really something we should include in "per-timestep" timers.

samhatfield: That's reasonable. Maybe we should add a print statement to explain this briefly as I was confused for a while why I got 102 iterations when I asked for 100.

samhatfield:
Tests failing for unknown reasons. Investigating...

samhatfield:
Tests were failing because the arrays containing time step stats were not resized for the extended (by two elements) time step loop in the benchmark program. Intel and NVHPC were happy with this for some reason.

marsdeno:
I think we probably want to aim for a single (unified) interface folder, rather than have separate ones for cpu and gpu?

CMakeLists.txt
```diff
@@ -48,22 +69,108 @@ set( HAVE_sp ${HAVE_SINGLE_PRECISION} )
 
 ecbuild_add_option( FEATURE MKL
                     DESCRIPTION "Use MKL for BLAS and/or FFTW"
-                    DEFAULT ON
+                    DEFAULT OFF
```
marsdeno, line 72: Should this really be default off?

CMakeLists.txt
```diff
@@ -48,22 +69,108 @@ set( HAVE_sp ${HAVE_SINGLE_PRECISION} )
...
+# following also needs cuda arch sm80
+ecbuild_add_option( FEATURE 3XTF32
+                    DEFAULT OFF
+		    CONDITION HAVE_CUTLASS
+		    DESCRIPTION "Support for 3xTF32 with Cutlass (>= 2.8)"
+                    REQUIRED_PACKAGES "NvidiaCutlass" )
+
+
+ecbuild_add_option( FEATURE GPU_AWARE_MPI
+                    DEFAULT OFF
```
marsdeno, line 156: Default ON might make more sense at this point.

marsdeno:
> I think we probably want to aim for a single (unified) interface folder, rather than have separate ones for cpu and gpu?

I guess it could be done after this PR

src/programs/ectrans-benchmark.F90
wdeconinck: I think a comment should definitely be added yes.
Also, I think we should refactor the code (followup) to have setup costs in `setuptrans` and not during the time stepping loop, and then this could be reverted.

wdeconinck:
> > I think we probably want to aim for a single (unified) interface folder, rather than have separate ones for cpu and gpu?
> 
> I guess it could be done after this PR

We need to work on a unified interface for any combination of double, single, cpu, gpu. Part of this work is already in develop on avoiding duplicate symbols.

src/programs/ectrans-benchmark.F90
lukasm91: I agree with adding a comment (or maybe a print statement saying "100 perf it + 2 warmup it").  The second part I disagree with. It might be possible for this benchmark to move set up costs into setup, but not in general. Also, it was never the case, because you would need to know beforehand all sizes for which spectral transforms are being called. This is possible for the benchmark here, but not representative for the real world. I think, we could evolve ectrans in the direction of creating handles similar to fft plans, so the users could create the "ectrans" handles during setup if they want (which would be a very nice direction because it would add better encapsulation of many global module variables we have today). But in the end, it is just a fact that the first 1-2, maybe more timesteps will be slower, be it due to memory registration during MPI, be it because of allocation overhead, be it because of FFT plan creation. It is always saver to exclude those.

wdeconinck: Right "warmup" iterations are always useful. I was not thinking of memory/cache/registration, ... and spoke too soon. I was only considering the extra computations required on first use.

Still having the FFT plan creation should in principle be feasible in setuptrans (not setuptrans0 !) because all runtime options (grid, resolution, domain decomposition) should be known at that point.

lukasm91: No, I don't think so. The problem is that FFT plan creation requires knowledge of number of levels, which depends on the configuration with which we call the spectral transform; so this is a property of dir_trans/inv_trans, not of setup_trans (if we call it once with a single level, and then with lvordiv=true, and once with lvordiv=false, three different sets of transforms are generated). Even worse, the graphs that are built on top, using the FFT plans, require the pointers of the intermediate/internal buffer. If the pointers change, graphs are destroyed and recreated, which can happen until we found the largest intermediate buffer. What I was using to avoid this in "production runs" is that I just have an env variable to set the minimum size of the internal buffer. This you only have to figure out once, and then you can set the env variable. Any graph will be created only once, and you don't pay for it many times. But maybe we can postpone this discussion and just add the comment for now. We should probably not strive for the perfect solution in the first step, otherwise we will never merge (and I am sure you agree with this :D )

samhatfield: I've added a print statement to clarify the number of iterations.

wdeconinck: I see. I also never intended this to be a blocker for this PR, just a follow-up discussion. Thanks for your insights @lukasm91 !

samhatfield:
I can confirm that all norms produced by the tests are identical between the CPU version in this branch and develop. To check this I had the norms before and after the benchmark loop dumped in hexadecimal to a text file and compared this between the full CTest suite between both branches. That's a nice feature we might like to implement properly at some point.

src/trans/gpu/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+        if( HAVE_GPU_STATIC )
+            set( GPU_LIBRARY_TYPE STATIC )
+        endif()
+
+        ectrans_add_library(
+          TARGET               trans_gpu_${prec}
+          TYPE                 ${GPU_LIBRARY_TYPE}
+          SOURCES              ${trans_src}
+          LINKER_LANGUAGE      Fortran
+          PUBLIC_INCLUDES      $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/trans/include>
```
marsdeno, line 81: This path does not exist : needs either to be updated to the include folder in the cpu folder, or we move the cpu includes up a level

src/trans/gpu/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
...
+                               $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/trans/include/ectrans>
+                               $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/algor/interface>
+                               $<INSTALL_INTERFACE:include/ectrans>
```
marsdeno, line 84: Same as above

samhatfield:
The include directories are definitely borked in this branch. Until we have unified interfaces, for now I suggest we aim for
```
install
|
 --> include
     |
      --> ectrans
          |
           --> cpu
           --> gpu
```
I tried implementing this first for the CPU version but I got lost and ended up creating an cpu/cpu directory at some point. Need a bit of help there.

wdeconinck:
Looking great so far. Some conflicts need resolving. I suggest merging 'develop' into this branch once more to resolve conflicts.

wdeconinck:
Is something substantial holding this back? (perhaps holidays :) )

samhatfield:
Apologies, I should have given an update here. One big thing which makes me hesitant to merge is the fact that this branch is not yet usable within the IFS even on Nvidia platforms. When I use this branch within IFS-RAPS, I get numerical instabilities. This issue only appeared a couple weeks ago when I gave the branch a spin in RAPS on JSC/JURECA-DC (A100). This was a surprise to me as I successfully used an earlier version of rgo-develop within RAPS with (GPU enabled). I'm sure we can get to the bottom of it, but it's hard to say how long it will take.

We could just get this PR over the line now with big warning signs that it is unstable. Or we could nail this bug beforehand. Any thoughts from others?

reuterbal:
I'd say as long as CPU and GPU works in the standalone benchmark and doesn't add regression over current behaviour (i.e., CPU version fully functional in IFS) there's nothing that should prohibit merging this and fixing incrementally going forward.

wdeconinck:
Great, so all good for integration.
Could I please propose a squash-merge of this branch into develop to keep history manageable?

wdeconinck (approved):

samhatfield:
By all means, do whatever you think will give us the cleanest history.

samhatfield:
I will attempt to merge the latest develop into rgo-develop now.

samhatfield:
Oh ffs. I have to open a new PR now?

PR #107, opened 2024-06-19
[redgreengpu] Fixes for CCE 17

reuterbal:
These have been obtained from vendor collaboration and supposedly fix compilation issue with  with ROCm 5.7.1, CCE 17.0.0, OMP and GPU aware MPI.

ROCm 6.1.1 seems to cause additional issues with the discovery of include paths for hipblas, hipfft and the rocm equivalents.
A potential fix is this:

```patch
diff --git a/cmake/ectrans_find_hip.cmake b/cmake/ectrans_find_hip.cmake
index 90fca11..b4142f5 100644
--- a/cmake/ectrans_find_hip.cmake
+++ b/cmake/ectrans_find_hip.cmake
@@ -110,10 +110,13 @@ macro( ectrans_find_hip )
     if( HAVE_HIP )
       list( APPEND ECTRANS_GPU_HIP_LIBRARIES ${hipblas_LIBRARIES} ${hipfft_LIBRARIES})
       list( APPEND ECTRANS_GPU_HIP_LIBRARIES ${rocblas_LIBRARIES} ${rocfft_LIBRARIES})
+      list( APPEND ECTRANS_GPU_HIP_INCLUDES ${hipblas_INCLUDE_DIRS}/hipblas ${hipfft_INCLUDE_DIRS}/hipfft)
+      list( APPEND ECTRANS_GPU_HIP_INCLUDES ${rocblas_INCLUDE_DIRS}/rocblas ${rocfft_INCLUDE_DIRS}/rocfft)
     endif()
   endif()
   ecbuild_info("HIP libraries: ${ECTRANS_GPU_HIP_LIBRARIES}")
+  ecbuild_info("HIP includes: ${ECTRANS_GPU_HIP_INCLUDES}")
 endmacro()
 macro( ectrans_declare_hip_sources )
diff --git a/src/trans/gpu/CMakeLists.txt b/src/trans/gpu/CMakeLists.txt
index 77d4e24..ad4c254 100644
--- a/src/trans/gpu/CMakeLists.txt
+++ b/src/trans/gpu/CMakeLists.txt
@@ -91,6 +91,7 @@ foreach( prec dp sp )
                                $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/algor/interface>
                                $<INSTALL_INTERFACE:include/ectrans>
                                $<INSTALL_INTERFACE:include>
+                               ${ECTRANS_GPU_HIP_INCLUDES}
           PUBLIC_LIBS          parkind_${prec}
                                fiat
```

However, I'm hoping there are proper library targets that could/should be used instead, and which needs to be examined once we have access to the newer software stack versions.

### DIFF
--- a/src/trans/gpu/internal/uvtvd_mod.F90
+++ b/src/trans/gpu/internal/uvtvd_mod.F90
@@ -1,6 +1,6 @@
 ! (C) Copyright 1991- ECMWF.
 ! (C) Copyright 1991- Meteo-France.
-! 
+!
 ! This software is licensed under the terms of the Apache Licence Version 2.0
 ! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
 ! In applying this licence, ECMWF does not waive the privileges and immunities
@@ -102,8 +102,7 @@ I_DIV_OFFSET = 2 * KFIELD
 !$OMP& MAP(ALLOC:ZN) &
 !$OMP& MAP(TO:D_MYMS,D_NUMP,R_NTMAX) &
 !$OMP& MAP(TO:F_RN) &
-!$OMP& MAP(ALLOC:ZEPSNM) &
-!$OMP& SHARED(I_DIV_OFFSET)
+!$OMP& MAP(ALLOC:ZEPSNM,ZOA1,ZOA2)
 #endif
 
 #ifdef OMPGPU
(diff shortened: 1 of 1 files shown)


wdeconinck:
Perhaps we can wait a bit for this after #106 is merged in develop?
Or merge here and keep a mental note that this will also need another PR to develop in addition, as 'redgreengpu' will not get merged to develop.

reuterbal:
Yes, I fully agree. In fact, we don't have any means of testing this ourselves, yet, and therefore might want to keep this open until confirmed functional.
It would still be useful to include this into redgreengpu even though it is not going to find its way into the main branches, since it continues to be the fallback for Cray+AMD platforms.

wdeconinck:
We can merge the source code change then.
I am also okay for you to add the cmake changes, but perhaps with the name "ECTRANS_GPU_HIP_INCLUDE_DIRS" instead of "...INCLUDES". This could be added to this PR as well.

samhatfield:
I guess we can close this @reuterbal?

reuterbal:
Yes, I think this has been superseded.

PR #110, opened 2024-07-01
Enable building ectrans against fftw

### LINKED ISSUE #109: Compile ectrans against libfftw
Following #104 I would like to be able to compile ectrans against libfftw (https://www.fftw.org/).

DJDavies2:
Resolves #109.

This attempts to enable compilation of ectrans against fftw while also preserving the ability to compile against other fftw implmentations.  In practice the only other system I have tested against MKL.  This seems to work for both but I am happy to look at other ways if you think they might be better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,11 +57,6 @@ endif()
 
 ecbuild_find_package( NAME FFTW REQUIRED COMPONENTS double ${single} )
 
-if( NOT HAVE_MKL )
-  set( fftw3 "fftw3" )
-  set( fftw3f "fftw3f" )
-endif()
-
 ecbuild_add_option( FEATURE TRANSI
                     DEFAULT ON
                     DESCRIPTION "Compile TransI C-interface to trans"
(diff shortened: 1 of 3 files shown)


samhatfield:
I take it you're still working on this?

DJDavies2:
I am looking into it, but at the moment I am testing using a Cray-installed fftw module that seems to work without ectrans modifications.  If that can be made to work then perhaps I won't need this branch.

The previous changes are due to me trying to compile FFTW myself and linking to it.  Do you want me to try and add support for that anyway?  I think the difference between the self compile and the Cray module is that the Cray module seems to have a some cmake support whereas the FFTW install system is still configure based.

samhatfield:
fftw does come with a top-level CMakeLists.txt if I remember right. If you build using that instead of Autoconf, perhaps it can be referenced by ecTrans without any modifications? If so that should be our official advice.

DJDavies2:
I rebuilt fftw using cmake and ectrans seems to link against that without modification by setting FFTW_ROOT.  Given that I am inclinced to close this; given that the cmake build of fftw seems to work we will have no use for the automake build.  Are you happy with that?

samhatfield:
Yes that sounds good to me :) feel free to close.

PR #111, opened 2024-07-02
Fix bug in median calculation

### LINKED ISSUE #108: Out of array boundaries accesses
Hello, 

I was investigating some segfault when ectrans is built Intel ifort, without the `-heap-arrays 32` flags. After building both fiat & ectrans in debug mode, the benchmark utility crashes with the following error : 
```
forrtl: severe (408): fort: (3): Subscript #1 of the array ZTSTEP has value 0 which is less than the lower bound of 1

Image              PC                Routine            Line        Source
ectrans-benchmark  000000000042B8E4  MAIN__                    859  ectrans-benchmark.F90
ectrans-benchmark  000000000040F48D  Unknown               Unknown  Unknown
libc-2.28.so       00007F497AC71D85  __libc_start_main     Unknown  Unknown
ectrans-benchmark  000000000040F3AE  Unknown               Unknown  Unknown
Command exited with non-zero status 152
```

This indicates that some accesses are made out of array boundaries. 

To reproduce, simply give `-DCMAKE_BUILD_TYPE=Debug` to the CMake of fiat and ectrans.  With an Intel suite, this leads to the following flags being used : 
```
  c compiler      : Intel 2021.10.0.20230609
    flags         :  -O0 -g -traceback
  fortran compiler: Intel 2021.10.0.20230609
    flags         :  -fast-transcendentals -fp-model precise -fp-speculation=safe -O0 -g -traceback -heap-arrays 32 -check all
```
Note the `-check all`on the fortran line : https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-2/check.html

And the command line to start the benchmark to reproduce the error above :
```
export OMP_NUM_THREADS=1
mpirun -n 1 ectrans-benchmark-sp -n 1 -l 137 -t 319
```

Some earlier experiments, before adding the check flags, did show backtraces originating from other places, so I suscpect the one I report above is not the only one.

I tested this with 
* ectrans 1.3.2; fiat 1.4.1; ecbuild 3.8.5; 
* Intel ifort version 2021.10. (from OneAPI 2023.2.0) and 2020.4.304 (+mkl & mpi from these versions)

Best.

samhatfield:
With the old style, the calculation of the median was broken for runs with a single iteration (accessing the zeroth element of the time step storage array) and did not give a conventional value for runs with two iterations (it should be the average of the two).

Resolves #108.

Dear colleagues @wdeconinck and @marsdeno, I hope this isn't controversial!

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -849,21 +849,15 @@ endif
 ztstepavg = (ztstepavg/real(nproc,jprb))/real(iters,jprd)
 ztloop = ztloop/real(nproc,jprd)
 ztstep(:) = ztstep(:)/real(nproc,jprd)
-
-call sort(ztstep,iters)
-ztstepmed = ztstep(iters/2)
+ztstepmed = get_median(ztstep)
 
 ztstepavg1 = (ztstepavg1/real(nproc,jprb))/real(iters,jprd)
 ztstep1(:) = ztstep1(:)/real(nproc,jprd)
-
-call sort(ztstep1, iters)
-ztstepmed1 = ztstep1(iters/2)
+ztstepmed1 = get_median(ztstep1)
 
 ztstepavg2 = (ztstepavg2/real(nproc,jprb))/real(iters,jprd)
 ztstep2(:) = ztstep2(:)/real(nproc,jprd)
-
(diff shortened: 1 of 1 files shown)


marsdeno (approved):
Looks very uncontroversial to me!

PR #112, opened 2024-07-03
Fix bug in zmaxerrg calculation

samhatfield:
Previously we completely ignored zmaxerr(4)!

Thanks @antoine-morvan for spotting this.


marsdeno (approved):
Nice catch!

PR #114, opened 2024-07-08
Remove NCOMBFLEN parameter

samhatfield:
This wasn't used anywhere.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -147,7 +147,6 @@ real(kind=jprd) :: zra = 6371229._jprd
 
 integer(kind=jpim) :: nmax_resol = 37 ! Max number of resolutions
 integer(kind=jpim) :: npromatr = 0 ! nproma for trans lib
-integer(kind=jpim) :: ncombflen = 1800000 ! Size of comm buffer
 
 integer(kind=jpim) :: nproc ! Number of procs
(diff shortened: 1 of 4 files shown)


wdeconinck:
Interesting find. I wonder when it was removed...

Similar to a previous cleanup, I would ask to keep the API stable but deprecate the use of kcombflen in setup_trans0 because this parameter is still used in e.g. [ifs-source/arpifs/setup/sutrans0.F90](https://git.ecmwf.int/projects/IFS/repos/ifs-source/browse/arpifs/setup/sutrans0.F90?at=refs%2Fheads%2Fmaster)

Definitely however we should have another ifs-source commit to fix the use of these deprecated parameters.

wdeconinck (changes requested):
Please keep external API stable with deprecation warning for the time being

samhatfield:
It was removed [16 years ago](https://github.com/ecmwf-ifs/ectrans/commit/8aaf3ec28d4b41a22135a8acd98b4901cbe5421f#diff-cec309202e471a4e0fe8adc60e8c9dcecd4034195e8e3fba9e728e67b8be7183L102) :p

But yes I will keep the argument but issue a warning. Just a second.

samhatfield:
PR updated.

wdeconinck:
16!!! years? Good riddance.

wdeconinck (approved):

PR #119, opened 2024-07-09
Reset N_MAX to 0 in DESTROY_PLANS_FFTW

### LINKED ISSUE #118: Atlas ctest failure in ectrans
One of the atlas ctests (atlas_test_functionspace) is failing in ectrans:

142: Runtime Error: /home/h01/frwd/cylc-run/mi-be984/work/1/ecbuild_ectrans_nag/build/src/trans/generated/ectrans_dp/internal/tpm_fftw.F90, line 231: Reference to disassociated POINTER TW%FFTW_PLANS
142: Program terminated by fatal error

DJDavies2:
Resolves #118.  In the test in question DESTROY_PLANS_FFTW is called twice; the first time is fine and the deallocations are done, but the second time is not okay because when the loop iterates there is an attempt to dereference TW%FFTW_PLANS when it is deallocated.  The fix sets N_MAX to 0 so that the loop doesn't execute.

What do you think?


src/trans/internal/tpm_fftw.F90
```diff
@@ -242,6 +242,7 @@ SUBROUTINE DESTROY_PLANS_FFTW
   IF( ASSOCIATED(TW%FFTW_PLANS) )  DEALLOCATE(TW%FFTW_PLANS)
   IF( ALLOCATED(TW%N_PLANS) )     DEALLOCATE(TW%N_PLANS)
 ENDIF
+TW%N_MAX=0
```
wdeconinck, line 245: That addition was definitely missing. Thanks for finding this @DJDavies2 !
I would just ask to insert the change instead within the preceding `IF( ASSOCIATED(TW )` statement.

DJDavies2: Thanks, I have done that.

wdeconinck (approved):

PR #124, opened 2024-07-31
Qualify all USE statements in GPU version

samhatfield:
I had issues building this branch on LUMI (AMD GPUs) because the Cray compiler was resolving the kind parameters (`JPIM` etc.) in a weird way. One module might import some variables from `PARKIND1` but also some from `ECTRANS_PARKIND`. But the latter also imports everything from `PARKIND1`, so I got errors about variables already being imported. The confusing setup with `PARKIND1`, `ECTRANS_PARKIND`, and `EC_PARKIND` should be resolved at some point. But for now we need to qualify those `USE` statements. This PR resolves this.

I also snuck in a commit fixing all `PRINT` statements.


wdeconinck:
In the CPU version parkind1 should no longer be a dependency.
A preprocessing step there  replaces PARKIND1 + JPRB with EC_PARKIND + (JPRD or JPRM).

Removing unqualified use_statements is definitely a first requirement and a great cleanup.

wdeconinck (approved):

samhatfield:
I'll have a chat with @marsdeno about `ectrans_parkind` at some point.

PR #127, opened 2024-08-01
Add missing USE statements

samhatfield:
I developed the previous PR using the Cray compiler, but it missed these which were flagged by the Nvidia compiler.


reuterbal:
You can try using this trick instead: https://github.com/ecmwf-ifs/fiat/blob/fa7a8b46b65568bd9e54bceebbff956c3f9da445/.github/workflows/build.yml#L123

samhatfield:
How come this only became a problem now though?

samhatfield:
Also, [a430e5c](https://github.com/ecmwf-ifs/ectrans/pull/127/commits/a430e5c9bd74cbe4ddb03b21490a28ba291c8edc) was a reaaaaly weird one. I did not appreciate spending my entire Friday tracking this down.

marsdeno:
> Also, [a430e5c](https://github.com/ecmwf-ifs/ectrans/pull/127/commits/a430e5c9bd74cbe4ddb03b21490a28ba291c8edc) was a reaaaally weird one. I did not appreciate spending my entire Friday tracking this down.

Does smell like a compiler bug, doesn't it

lukasm91:
> Also, [a430e5c](https://github.com/ecmwf-ifs/ectrans/pull/127/commits/a430e5c9bd74cbe4ddb03b21490a28ba291c8edc) was a reaaaally weird one. I did not appreciate spending my entire Friday tracking this down.

Looking into it ...

src/trans/gpu/internal/ext_acc.F90
```diff
@@ -328,7 +328,7 @@ subroutine ext_acc_copyout(ptrs, stream)
     enddo
   end subroutine
   subroutine ext_acc_delete(ptrs, stream)
-    use openacc, only: acc_async_sync
+    use openacc
```
lukasm91, line 331: @samhatfield instead of deleting all those, just add `acc_delete` here. not sure which acc_delete it is trying to link when it is not given here, but it makes sense that it is crashing.

```suggestion
    use openacc, only: acc_async_sync, acc_delete
```

src/trans/gpu/internal/ext_acc.F90
```diff
@@ -301,7 +301,7 @@ subroutine ext_acc_copyin(ptrs, stream)
     enddo
   end subroutine
   subroutine ext_acc_copyout(ptrs, stream)
-    use openacc, only: acc_async_sync, acc_copyout
+    use openacc
```
lukasm91, line 304:
```suggestion
    use openacc, only: acc_async_sync, acc_copyout
```

src/trans/gpu/internal/ext_acc.F90
```diff
@@ -274,7 +274,7 @@ subroutine ext_acc_create(ptrs, stream)
     enddo
   end subroutine
   subroutine ext_acc_copyin(ptrs, stream)
-    use openacc, only: acc_async_sync
+    use openacc
```
lukasm91, line 277:
```suggestion
    use openacc, only: acc_async_sync
```

src/trans/gpu/internal/ext_acc.F90
samhatfield: Ah nice spot :) Instead of actually studying the code, I mistakenly relied on the compiler to tell me which symbols needed to be imported. Any idea why the compiler doesn't complain when `acc_delete` isn't imported from `openacc`?

lukasm91: I am not a fortran standard expert, maybe you would need to add an `implicit none` somehwere to let it complain...  (maybe just inside the subroutine?). So it then assumes there is a acc_delete, and I think this will just link the acc_delete from the C API which is not compatible with acc_delete from the module.

marsdeno: The module has a global "implicit none" at line 10, so this should cover all the included subroutines should it not?

samhatfield: It should yes.

lukasm91: I'll see if I can create a reproducer and submit it to our compiler devs.

lukasm91: I thought about it again. implicit none doesn't cover externals, those can still be implicit. Only F18 adds `implicit none (external,type)`. The external spec would cover this problem, but it is not supported in NVHPC.

samhatfield: Interesting. So `implicit none` by itself only checks for variables (equivalent to `implicit none (type)`. The compile won't throw an error if an arbitrary routine is called. The symbol will probably not be resolved at link time of course, but in this case it was being incorrectly resolved from the C API.

Issue #126, opened 2024-08-01
The fate of _32 routines


samhatfield:
There are currently 10 files (total 1844 lines) with a "_32" suffix:
- trans/cpu/external/dist_grid_32.F90
- trans/cpu/external/gath_grid_32.F90
- trans/cpu/internal/dist_grid_32_ctl.F90
- trans/cpu/internal/gath_grid_32_ctl.F90
- trans/gpu/external/dist_grid_32.F90
- trans/gpu/exeternal/gath_grid_32.F90
- trans/gpu/internal/dist_grid_32_ctl.F90
- trans/gpu/internal/gath_grid_32_ctl.F90
- trans/include/ectrans/dist_grid_32.h
- trans/include/ectrans/gath_grid_32.h

It seems that these are intended to be versions of the respective non-\_32 subroutines that accept single-precision arguments. However they haven't been kept in sync with the non_-32 versions and they only seem to be called in the IFS ENKF, which has been out of use for 8 years now. So I'd like to find a way to remove these subroutines as I suspect they are not required by any recent or future users of ecTrans.

If I was to delete the internal subroutines, and replace the external ones with a deprecation warning and an abort, then we can remove them completely in a future release? How does that sound @wdeconinck? This would allow us to delete most of the code while still allowing ifs-source to compile against this version of ecTrans.

wdeconinck:
Semantically that would not a deprecation warning as the function would no longer runs through. Rather it is a hard error. If you are sure this code path is no longer tested I am good with this for sure.

samhatfield:
Good point yes.

I'm just checking now with Massimo and Mats that we really don't need to keep EnKF code maintained.

samhatfield:
Response from Mats Hamrud:
> I think they can be depreciated. They were written before the general use of single precision in the IFS, EnKF is all single precision only.

wdeconinck:
They are still being referenced in current `ifs-source/enkf/module/xb_state_mod.F90`
So I don't think we can remove it.

PR #128, opened 2024-08-01
Remove _32 subroutines

samhatfield:
Addresses #126.


samhatfield:
Issue not urgent. Will look to removing all uses of _32 routines in the IFS in future, then this can be revisited.

PR #131, opened 2024-08-06
GPU: Fix data transfers if already on device

lukasm91:
(requested by Judicael Grasset)

If data is already on device, we should not make an extra copy (similar to the openacc semantics: copy means copy if not present)

Should go after #127


samhatfield:
@lukasm91 could you rebase this branch against latest develop?

lukasm91:
> @lukasm91 could you rebase this branch against latest develop?

done

src/trans/gpu/internal/trgtol_mod.F90
```diff
@@ -120,7 +120,7 @@ SUBROUTINE TRGTOL(ALLOCATOR,HTRGTOL,PREEL_REAL,KF_FS,KF_GP,KF_UV_G,KF_SCALARS_G,
     USE BUFFERED_ALLOCATOR_MOD, ONLY: BUFFERED_ALLOCATOR, ASSIGN_PTR, GET_ALLOCATION
     USE OPENACC_EXT,            ONLY: EXT_ACC_ARR_DESC, EXT_ACC_PASS, EXT_ACC_CREATE, &
       &                               EXT_ACC_DELETE
-    USE OPENACC,                ONLY: ACC_HANDLE_KIND
+    USE OPENACC,                ONLY: ACC_IS_PRESENT
```
samhatfield, line 123: Should we still be importing `ACC_HANDLE_KIND`? It's referenced further down.

lukasm91, line 123:
```suggestion
    USE OPENACC,                ONLY: ACC_IS_PRESENT, ACC_HANDLE_KIND
```

lukasm91: I think so, yes. Fixing...

src/trans/gpu/internal/trltog_mod.F90
```diff
@@ -121,7 +121,7 @@ SUBROUTINE TRLTOG(ALLOCATOR,HTRLTOG,PREEL_REAL,KF_FS,KF_GP,KF_UV_G,KF_SCALARS_G,
     USE ISO_C_BINDING,          ONLY: C_SIZE_T
     USE OPENACC_EXT,            ONLY: EXT_ACC_ARR_DESC, EXT_ACC_PASS, EXT_ACC_CREATE, &
       &                               EXT_ACC_DELETE
-    USE OPENACC,                ONLY: ACC_HANDLE_KIND
+    USE OPENACC,                ONLY: ACC_IS_PRESENT
```
samhatfield, line 124:
```suggestion
    USE OPENACC,                ONLY: ACC_HANDLE_KIND, ACC_IS_PRESENT
```

samhatfield: Same thing as in the other file.

lukasm91: I should have realized... sorry for that :)

samhatfield:
One thing I don't get about this change: first time we call `DIR_TRANS`, `ACC_IS_PRESENT(PGPUV)` (say) is `.FALSE.`. So a host -> device update happens for `PGPUV`. But the second time, `ACC_IS_PRESENT(PGPUV)` is `.TRUE.`, so the update does not happen. That means the second call to `DIR_TRANS` is not operating on the latest values in `PGPUV` on the host side.

In other words, don't we _always_ have to do a host->device update for all `PRESENT` (Fortran, not OpenACC) arrays in `TRGTOL`? And vice versa for `TRLTOG`.

lukasm91:
Let me explain in more detail: I can think of using ectrans in two modes of operations. 

 - Currently, the underlying assumption is that when you enter ectrans, the inputs/outputs are never on device. ectrans has to take care of copying the data onto device, and in the end has to make sure that the data is also updated back. The spectral space is fine, this is handled in ltdir_mod/ltinv_mod, because those are acc data copyout. On the grid-point space it is also fine because we do the updates
 - On the other hand ,we can use ectrans with PGP, ... already on device. ectrans should not make any copies, otherwise it would overwrite the data on the device.  The spectral space is fine again, because acc data copyin/out only makes the copy if the data is not present. The grid-point space should only do the copy if the data is not present before we potentially create it.

In the current main branch, ectrans is behaving very weirdly: On the grid-point space it does the copy in any case, in the spectral space it only copies if the data is not present. This PR changes the behaviour for both spaces to "only copy if not present".

samhatfield:
Okay, let me put what you're saying in other terms to see if I understand:
### Current behaviour
- `DIR_TRANS`: 
  - Inputs (`PGP` etc.) *always* copied from host to device (`TRGTOL`)
  - Outputs (`PSPVOR` etc.) *always* copied from device to host (`LTDIR`)
- `INV_TRANS`:
  - Inputs (`PSPVOR etc.) *always* copied from host to device (`LTINV`)
  - Outputs (`PGP` etc.) *always* copied from device to host (`TRLTOG`)
  
### New behaviour
- `DIR_TRANS`: 
  - Inputs (`PGP` etc.) only copied from host to device when not already present on device (`TRGTOL`)
  - Outputs (`PSPVOR` etc.) *always* copied from device to host (`LTDIR`)
- `INV_TRANS`:
  - Inputs (`PSPVOR etc.) *always* copied from host to device (`LTINV`)
  - Outputs (`PGP` etc.) only copied from device to host when not already present on device (`TRLTOG`)

But writing out all that only raises more questions...
1. `DIR_TRANS`: `PGP` etc. are arguments, not module variables. Hence `PGP` etc. cannot be made present on device by anyone other than an earlier call to `TRGTOL`. In that case, they would be present, but would contain stale values. So an update would be required, always.
2. For `INV_TRANS`, the first time you call it, `PGP` etc. won't be present, and `PGP` etc. will be copied from device to host. The second time, `PGP` etc. will be present, so `PGP` etc. won't be copied from device to host. Why would you not want to copy the output back to host? Again, `PGP` etc. are arguments, so you cannot access them anywhere else.

lukasm91:
Let me modify your explanation and correct it (current behaviour is *not* consistent):

> ### Current behaviour
> 
>     * `DIR_TRANS`:
>       
>       * Inputs (`PGP` etc.) _always_ copied from device to host (`TRGTOL`)
>       * Outputs (`PSPVOR` etc.) only copied from device to host when not already present on device (`LTDIR`)
> 
>     * `INV_TRANS`:
>       
>       * Inputs (`PSPVOR` etc.) only copied from host to device when not already present on device (`LTINV`)
>       * Outputs (`PGP` etc.) _always_ copied from device to host (`TRLTOG`)
> 
> 
> ### New behaviour
> 
>     * `DIR_TRANS`:
>       
>       * Inputs (`PGP` etc.) only copied from host to device when not already present on device (`TRGTOL`)
>       * Outputs (`PSPVOR` etc.) only copied from device to host when not already present on device (`LTDIR`)
> 
>     * `INV_TRANS`:
>       
>       * Inputs (`PSPVOR` etc.) only copied from host to device when not already present on device (`LTINV`)
>       * Outputs (`PGP` etc.) only copied from device to host when not already present on device (`TRLTOG`)

To answer the other questions:

>     1. `DIR_TRANS`: `PGP` etc. are arguments, not module variables. Hence `PGP` etc. cannot be made present on device by anyone other than an earlier call to `TRGTOL`. In that case, they would be present, but would contain stale values. So an update would be required, always.

Yes, they are arguments. PGP can be made present by the callee:

!$ACC DATA CREATE(PGP, ...)
call dir_trans(...)
call inv_trans(...)
!$ACC END DATA


>     2. For `INV_TRANS`, the first time you call it, `PGP` etc. won't be present, and `PGP` etc. will be copied from device to host. The second time, `PGP` etc. will be present, so `PGP` etc. won't be copied from device to host. Why would you not want to copy the output back to host? Again, `PGP` etc. are arguments, so you cannot access them anywhere else.

Same reason. This is not about the first/second/third call, it is about what happens if the data is managed outside of ectrans. Assume the whole code of IFS is ported to GPU, in this case all data always live on the GPU, including PGP, PSPVOR, ...  There should be 0 copies, which is why the "current behaviour" is not what we want.

wdeconinck:
@samhatfield is @lukasm91 explanation sufficient? Is there anything else blocking, besides conflicts that need to be resolved (rebase please)?

samhatfield:
To the best of my memory, I didn't feel confident we could merge this as what Lukas is proposing (which is reasonable of course!) has implications outside of just ecTrans, and influences our overall strategy for device-side memory handling in the IFS. I need to refresh my memory though - let me take a look again.

wdeconinck:
So what Sam understood is that PGP could be present (acc_mapped) on device just because of previous calls to inv_trans or dir_trans. 
So a new call to dir_trans will then see that PGP is acc_mapped and assume its values are up to date.

The IFS currently has no notion of any device copy and only works with host data, treating the GPU as an offload-accelerator for ectrans.
The IFS should become itself GPU aware to track multiple memory spaces and provide info to ectrans if a field is up-to-date on the GPU or not. I guess that is the real problem. A new inv_trans / dir_trans API would need to be created for this.

Notwithstanding there are inconsistencies discovered by @lukasm91 about the treatment of PSPVOR that should be corrected. Probably to the safe side where the copy has to be made __always__ ?
A new inv_trans/dir_trans API could then cater for the __copy if not present__?

awnawab:
Hi all. I am interested in a similar development, is there a plan for this to be merged?

I agree with @wdeconinck's comments, in that simply checking for the existence of those addresses in the openacc map is probably not enough. With FIELD_API we only use openacc primitives once upon allocation and then deallocation, so the openacc reference counter isn't updated for example if I call `FIELD%SYNC_HOST_RDWR`. `ACC_IS_PRESENT` in that situation would return true, even though the device copy is stale, and an equivalent `!$acc exit data copyout` would have made the right change to the reference counter.

We should instead have an optional argument that indicates whether the relevant arrays are on device. This would allow us to control with runtime flags within the IFS whether data is persisted on device across calls to `INV/DIR_TRANS`. Considering the wide variety of ways in which we call `INV/DIR_TRANS` from within the IFS (e.g. in the setup and norms), that functionality would be highly desirable, if for nothing more than ease of debugging.

I would be happy to file a PR that adds optional args like `LPGP_ON_GPU` and uses those to achieve the above logic, but I think this PR should either be merged first or closed.

samhatfield:
I deferred this issue because I thought it depended on the wider GPU strategy for the IFS, which wasn't clear to me two years ago. At the time, the only use case for GPU ecTrans was as the *only* part of the IFS running on GPUs. Of course we've made a lot of progress since then and I can see how this is now become important.

@awnawab are you essentially proposing a new argument for `INV/DIR_TRANS` such as `LHOSTSYNC` which would copy inputs from host->device before and from device->host after? If this is false, data is both assumed to be already on device on input and is left on device on output, with no update to the host.

In other words, if ecTrans is the only part of the model running on GPUs, one should pass `LHOSTSYNC=.true.`. If everything before and after ecTrans is running on GPUs, one should pass `LHOSTSYNC=.false.`.

Something like that?

awnawab:
Yes in essence.

For example in `TRGTOL`, we could have an optional argument called `LPGP_ON_GPU`. This would default to false, in which case the data offload logic would stay as is. If it's present and true, then we skip the update device calls for all the `PGP*` arrays. In `TRLTOG`, we would then skip the update host calls if `LPGP_ON_GPU` is present and true. This would allow us to control per `INV/DIR_TRANS` call in the IFS whether data is persisted on device across that call.

awnawab:
For reference, what I'm proposing would look like this: https://github.com/ecmwf-ifs/ectrans/compare/develop...awnawab:ectrans:naan-device-resident-pgp

I'm happy to file that against develop once we decide what to do with the current PR, since my branch will conflict with the changes herein.

samhatfield:
#404 merged so this PR is now redundant.

PR #132, opened 2024-08-12
[GPU] Fix potential NaNs when using cutlass

lukasm91:
This fix is some potential NaNs. 

I didn't thoroughly checked performance on this, but this can be necessary in some corner cases: The legendre coefficients are zero padded, so we can in principle safely keep the padding data uninitialized, *except* if the padding data is NaN (can be simulated in the buffered_allocator_mod.F90  by removing the SET_VALUE_EFF check such that all data is initialized to NaN).

I remember that this happened in some case in IFS, when some data was previously used in another data type. This is only needed when we are using 3XTF32, because only this case is using padded data.

I added a warning for this in the code where we require the padding. Note that to trigger this issue, we must be in single precision, using cutlass and 3xtf32. This should be kept in mind that if at some point we see sudden NaNs in the output, we might have missed a corner case (which I think we didn't). 

@marsdeno  You already have the first of those commits in your branch. The second commit is doing the same in the other direction.


samhatfield:
I think @marsdeno is more familiar with the CUTLASS capability.

marsdeno (approved):
This looks good to me, and will get branch with matching commit in other direction ready asap

wdeconinck:
So, consensus to merge this @marsdeno ?

PR #133, opened 2024-08-12
[GPU] Add missing ASYNC statements

samhatfield:
Without these the UPDATE can be carried out before the device arrays have been properly filled by the preceding ASYNC(1) unpacking DO loop.

This was causing an obscure error during integration testing of develop in the IFS.


marsdeno (approved):
Well spotted!

PR #139, opened 2024-08-27
Add capability to call GPU version via transi C interface

wdeconinck:
We also want to be able to call the GPU version of ectrans via the C interface to be used from Atlas.

This PR simply creates a new transi_gpu_dp CMake target. Programs previously using transi_dp now need to link to transi_gpu_dp.

Note that a program can and should not link to both transi_gpu_dp and transi_dp due to symbol conflicts. This will be resolved in the future with a unified API for all backends.

In transi (also CPU) the restriction is currently that NPRTRV == 1. Hence it is also not needed to pass the optional variables KVSETUV, KVSETSC etc. to 'dir_trans'.

The logic in TRGTOL (gpu edition) is currently wrongly relying on these. This PR fixes that as well.

There is a new test `ectrans_test_transi_program_gpu` which one should be running as follows on our atos:

```sh
export TRANS_USE_MPI=1       # just to enable MPI within the test
srun -q dg -n 1 --gpus-per-task=1 tests/ectrans_test_transi_program_gpu
```

This currently fails with an error:
```
CUDA error: ****/hicfft_execute_plan.cuda.cu, line 209, hipDeviceSynchronize(): an illegal memory access was encountered
```
The Fortran dir_trans call looks like this (from [here](https://github.com/ecmwf-ifs/ectrans/blob/main/src/transi/transi_module.F90#L1521))
```
    call DIR_TRANS( KRESOL=trans%handle, &
      &             KPROMA=args%nproma, &
      &             PGP=RGPM, &
      &             PSPVOR=RSPVOR,PSPDIV=RSPDIV,PSPSCALAR=RSPSCALAR ) ! unused args: KVSETUV,KVSETSC
```
@marsdeno or @lukasm91 @samhatfield can you try to reproduce and have any idea why?

Note, [similar changes](https://github.com/ecmwf-ifs/ectrans/compare/redgreengpu...wdeconinck:ectrans:redgreengpu_transi) applied to the redgreengpu branch give a similar error:
```
GPU runtime error: Failed to synchronize
GPU runtime error: Failed to synchronize
GPU runtime error at 1
GPU runtime error in file '*****/hicfft_execute_plan.cuda.cu'
GPU runtime error at 2
GPU runtime error line '55'
GPU runtime error at 3
GPU runtime error 1: CUFFT_INVALID_PLAN
terminating!
ABOR1     [PROC=1,THRD=1] from [*****/hicfft.h +45] : Error in FFT
```


samhatfield:
I'm able to reproduce your error, but I don't know the cause. The error's thrown by an innocuous-looking synchronize, so likely it originates elsewhere. Unfortunately we only have the ability at present to disable graphing in the GEMM, not in the FFT. We really ought to be able to disable both. Not sure how to fix this at the moment.

lukasm91:
@wdeconinck First of all, I think the issue that you see is simply that the binaries are not linked with `-acc`. You can check with PGI_ACC_DEBUG=1, and check the first line if it actually runs on device (if it says PGI native, this is CPU, it should rather be ACC detected 1 device, and tell you what device it is). It then just starts running and suddenly passes pointers to cublas/cufft that have never been allocated.

If I add -acc to the link flags, I believe we run into a separate issue. The error is about an array already being allocated, which might be a much more fundamental issue: I don't think the GPU version ever supported different resolutions properly. We might need to start looking into where should ZAA/ZAS/... be stored, because it seems like this test tries to reallocate them, or if it is the same resolution, it should just not try reallocate them. 
On the other hand this would also be a good opportunity to finally do a cleanup. There are all this extra copies of a bunch of arrays (D_NSTAGT0B copied from D%NSTAGT0B, ......). I think might have been a time this was required, long before I started working on ectrans, in the ESCAPE time, but it is not that important anymore to make those extra copies; we might then only need few arrays to be moved into the data structures, maybe ZAA/ZAS,ZAA0/ZAS0, and ZEPSNM.

wdeconinck:
Thanks @lukasm91 !
Indeed I can also now proceed further when linking with "-acc".
Do you know if "-gpu=pinned" is also a required link flag or is that only required on compiling the ".o"s ?

I managed to run the test program in question can also be run to completion if I comment out some parts.
i.e.:
- https://github.com/ecmwf-ifs/ectrans/blob/5e920dddadde8db44a18a2c519724a559425429e/tests/transi/transi_test_program.c#L150-L160
- https://github.com/ecmwf-ifs/ectrans/blob/5e920dddadde8db44a18a2c519724a559425429e/tests/transi/transi_test_program.c#L231-L242
These 'vordiv_to_UV' routines only operate in spectral space and don't really require the full spectral transforms (and thus any stored coefficients). A special "lightweight" KRESOL (or trans_handle) is internally created and destroyed for this when calling this routine that does not hold Legendre coefficients or FFTs.

Finally I also had to comment out 
https://github.com/ecmwf-ifs/ectrans/blob/5e920dddadde8db44a18a2c519724a559425429e/tests/transi/transi_test_program.c#L313-L315

It seems there are problems with deallocating some structures. In the Fortran `ectrans-benchmark` codes we never have called these routines and I guess we let the OS and runtimes clean everything up. I think we should also be able to observe this problem in the Fortran `ectrans-benchmark` code by adding  `trans_release(KRESOL)` (mirroring `setup_trans(KRESOL)`) and `trans_end` (mirroring `setup_trans0`) in the `ectrans-benchmark`.
Probably it would be a good idea to add it and fix that with those tests first.

lukasm91:
This is a good suggestion that we should ad dtrans_release and trans_end, and yes this can very well be that this is not properly cleaned up. @samhatfield  @marsdeno  for vis (I might do it, eventually, but if you happen to start working on this, let me know). Nonetheless I think cleaning up all this extra D_XXX arrays might be a good thing anyway.

>  Do you know if "-gpu=pinned" is also a required link flag or is that only required on compiling the ".o"s ?
It is absolutely required in the linked line too, so it should be put into the PUBLIC section of the link flags.

reuterbal:
Just in case: it could be that the flags should be added to the `INTERFACE` section rather than `PUBLIC`, to make sure they appear in the link lines but to avoid the issue of the flag being propagated to all downstream components. The same occured with field api before it was fixed in https://github.com/ecmwf-ifs/field_api/pull/56.

wdeconinck:
This branch should pass the tests of using transi with GPU DP backend.
I have disabled the VorDivToUV code path. That needs to be fixed to work in a further PR.
The last commit disables trans_delete and trans_finalize from the transi test. That commit can be reverted once PR #151 is merged. I have verified that works, having rebased this branch on PR #151. That is also a good verification that #151 works, at least with NVHPC.

Issue #138, opened 2024-08-27
GPU feature cannot be compiled without MPI


wdeconinck:
The current GPU code-path cannot be compiled without MPI.
This is an issue if we want to use ectrans with GPU in a MPI-serial context.

samhatfield:
The solution is to move to MPL for calls like [this](https://github.com/ecmwf-ifs/ectrans/blob/main/src/trans/gpu/internal/trgtol_mod.F90#L583).

I think @marsdeno has done some work on this?

reuterbal:
Indeed. This was blocked by support for GPU-aware MPI communication in MPL, which in turn I believe requires using mpi_f08 in MPL.

wdeconinck:
Arguably in a context of ENABLE_MPI=OFF (serial) there's nothing to send or receive and mpi_f08 would not be needed either.
Preprocessor guards could be added where needed temporarily, even if it's ugly, until MPL can handle the GPU-aware comms.

wdeconinck:
Resolved with #140

PR #140, opened 2024-08-27
Enable GPU feature without MPI requirement

wdeconinck:
This solves #138 by introducing preprocessor guards around MPI includes and symbols


marsdeno (approved):
Looks ok to me.
Indentation of the fortran is getting harder to follow in some places with the number of preprocessor guards we're using, but that is not the fault of this PR!

wdeconinck:
Once MPI is replaced by MPL we should have access to all symbols again, and the preprocessor guards can be removed.

samhatfield:
Looks good. However note that although you can now build ecTrans/GPU without MPI, you still can't run it. So MPI is still effectively required for running on GPUs, even with a single MPI task. Since this is (hopefully) temporary, this is nitpicking.

wdeconinck:
> Looks good. However note that although you can now build ecTrans/GPU without MPI, you still can't run it. So MPI is still effectively required for running on GPUs, even with a single MPI task. Since this is (hopefully) temporary, this is nitpicking.

You will be happy to learn I was successful at running GPU code without linking to any MPI library this way. This is because the calls to MPI are guarded at runtime with `IF`s that relate to send/receive counts that amount to zero.

samhatfield:
That's great! I assumed that TRLTOG/GTOL were not clever enough to handle NPROC=1 without issuing MPI calls, unlike TRLTOM/MTOL which has a separate code path. But clearly I'm wrong.

samhatfield (approved):

PR #145, opened 2024-09-03
Fix corner cases with many ranks

lukasm91:
Fixing a few corner cases:
- The barriers are causing a hang if not all ranks have data in the spectral space, hence we should use the right communicators for those.
- Various potential "not-present" errors, some of them can be triggered by extreme corner cases (e.g. if dir_trans comes first, and there are more ranks than w-sets). We have actually seen this with Meteo France codes.

No performance impact expected.

@marsdeno Please make sure you carefully check trltog/trgtol in your branch, in the end. You might have an extra case because you might have an additional acc present clause added due to the issue with use_device...

I don't expect this to cause any difficult conflicts for anyone.


src/trans/gpu/internal/inv_trans_ctl_mod.F90
```diff
@@ -196,12 +196,14 @@ SUBROUTINE INV_TRANS_CTL(KF_UV_G,KF_SCALARS_G,KF_GP,KF_FS,KF_OUT_LT,&
     IF (IF_FOURIER /= KF_FS) CALL ABORT_TRANS('Size mismatch: Wrong computation KF_FS')
 
     ALLOCATOR = MAKE_BUFFERED_ALLOCATOR()
-    HLTINV = PREPARE_LTINV(ALLOCATOR,KF_UV,KF_SCALARS,LVORGP,LDIVGP,LSCDERS)
-    HTRMTOL_PACK = PREPARE_TRMTOL_PACK(ALLOCATOR,IF_LEG)
-    HTRMTOL = PREPARE_TRMTOL(ALLOCATOR,IF_LEG)
-    HTRMTOL_UNPACK = PREPARE_TRMTOL_UNPACK(ALLOCATOR,IF_FOURIER)
-    HFSC = PREPARE_FSC(ALLOCATOR)
-    HFTINV = PREPARE_FTINV(ALLOCATOR)
+    IF (KF_FS > 0) THEN
```
lukasm91, line 199: This is just replicating the code structure below.

marsdeno (approved):
Looks good to me, and noted for the warning, will check my pending branch carefully.

samhatfield:
Branch tested in the IFS on JEDI (GH200). Works with 64 tasks/GPUs.

marsdeno:
[like]  Olivier Marsden reacted to your message:
________________________________

PR #146, opened 2024-09-12
Tidy GEMMs

samhatfield:
I've deleted some unused interface functions and underlying C code.

Note: would be good to add `INTENT` declarations for the subroutines in hicblas_mod.F90.


wdeconinck:
At first sight it looks like the teardown of the GPU GEMMS is not being used (the finalise), and you can delete this piece.
But perhaps it's wrong not to call any sort of finalise.
Instead it should be called during trans_relase or trans_end or something of this sort.

samhatfield:
For the GEMMs, is there anything that requires finalisation? You don't have plans etc. like the FFTs.

lukasm91:
I suggest we delete it here because it is from the old implementation. I already planned adding destructing the GEMM plans in the implementation of the trans_end/trans_release that is required for transi.

wdeconinck:
OK good to merge :) Thanks Sam !

wdeconinck (approved):

PR #147, opened 2024-09-12
Fix calculation of zstepavg

samhatfield:
We collect `iters+2` statistics, but use `iters` to calculate the average. We should use `iters+2` to calculate the average.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -876,16 +876,16 @@ if (luse_mpi) then
   call mpl_allreduce(ztstepmin2, 'min', ldreprod=.false.)
 endif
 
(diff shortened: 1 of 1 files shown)


lukasm91:
Good catch, though wouldn't it be better if we were to report the timings only after the third timestep (do the same warm up phases we also do for gstats labels)? I think this should be more representative for a real run. On the other hand we can also just focus on median times, but I believe having a representative mean time is also useful.

wdeconinck (approved):

wdeconinck:
It's still useful to output the timing and norms for the warmup steps, but perhaps we should **warn** in the output that they are not counted towards the average time.

samhatfield:
PR updated. Just testing now.

samhatfield:
Seems to work nicely.

wdeconinck (approved):

PR #148, opened 2024-09-12
Move more TPM modules and setup routines to common

wdeconinck:
Another small step in direction of using multiple handles of different backend and resolution.

Moving more TPM modules and setup routines to a common precision/backend independent library means that there is no code duplication and/or multiple compilation and a possible reuse of the globals that are part of these modules across the different backends. 

I managed to move all but one TPM modules.
The missing one for now is TPM_FLT because that essentially contains the Legendre coefficients for each backend in a specific precision. The CPU version also contains the fast-legendre coefficients and butterfly structures which are not ported to the GPU.
Because of this, it will be still some task to see how we can reuse the Legendre coefficients of same resolution across different precisions and backends without recomputing them.

Most work was spent in making TPM_FIELDS , and the flattened GPU arrays have been moved to a new TPM_FIELDS_FLAT module, and I understand that the code will be refactored soon to remove the use of the flattened arrays.

Note that results will not be bit-identical, but hopefully performance-neutral. Could someone benchmark/verify this for me?


src/trans/cpu/internal/ledir_mod.F90
```diff
@@ -189,7 +189,7 @@ SUBROUTINE LEDIR(KM,KMLOC,KFC,KIFC,KSL,KDGLU,KLED2,PAIA,PSIA,POA1,PW)
   DO JK=1,KFC,ISKIP
     IFLD=IFLD+1
     DO J=1,KDGLU
-      ZB(J,IFLD)=PSIA(JK,ISL+J-1)*PW(ISL+J-1)
+      ZB(J,IFLD)=PSIA(JK,ISL+J-1)*REAL(PW(ISL+J-1),JPRB)
```
samhatfield, line 192: Since the weights `F%RW` have been promoted from `JPRB` to `JPRD`, do we have any idea about the cost of the necessary casting here? Could it be significant enough to be a noticeable performance penalty?

samhatfield (commented):
Looks good, but I'd like a bit more assurance that the additional casting won't impact performance. Maybe it's insignificant, but we should demonstrate that with evidence. I'll try running some benchmarks.

samhatfield (approved):

marsdeno (approved):
Looks good, and nice testing from @samhatfield

PR #149, opened 2024-09-13
Configurable warmup iterations for ectrans-benchmark

wdeconinck:
This PR should not be as difficult to review and includes:
- Add command line argument --niter-warmup to ectrans-benchmark
- Modify default to 3 as after this, experimentally the timings stay more or less constant
- Some cosmetics in printing and removal of spurious prints.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -65,7 +65,7 @@ integer(kind=jpim) :: nsmax   = 79  ! Spectral truncation
 integer(kind=jpim) :: iters   = 10  ! Number of iterations for transform test
 integer(kind=jpim) :: nfld    = 1   ! Number of scalar fields 
 integer(kind=jpim) :: nlev    = 1   ! Number of vertical levels
-integer(kind=jpim) :: iters_warmup = 2 ! Number of warm up steps (for which timing statistics should be ignored)
+integer(kind=jpim) :: iters_warmup = 3 ! Number of warm up steps (for which timing statistics should be ignored)
 
 integer(kind=jpim) :: nflevg
 integer(kind=jpim) :: ndgl ! Number of latitudes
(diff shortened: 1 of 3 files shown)


src/programs/ectrans-benchmark.F90
```diff
@@ -571,16 +569,16 @@ program ectrans_benchmark
 
 if (verbosity >= 0 .and. myproc == 1) then
   write(nout,'(" ")')
-  write(nout,'(a,i6,a,f9.2,a)') "ectrans_benchmark initialisation, on",nproc,&
+  write(nout,'(a,i0,a,f9.2,a)') "ectrans_benchmark initialisation, on",nproc,&
```
samhatfield, line 572: Haven't seen `i0` before. Does that automatically scale the width of the printed integer?

src/programs/ectrans-benchmark.F90
```diff
@@ -706,8 +704,8 @@ program ectrans_benchmark
     if (myproc == 1) then
       zmaxerr(:) = -999.0
       do ifld = 1, 1
-        write(nout,*) "znormsp", znormsp
-        flush(nout)
+        !write(nout,*) "znormsp", znormsp
+        !flush(nout)
```
samhatfield, line 708: I think we just delete these.

samhatfield (changes requested):
Looks good!

src/programs/ectrans-benchmark.F90
marsdeno: Correct

src/programs/ectrans-benchmark.F90
wdeconinck: Done

marsdeno (approved):
All good for me

samhatfield (approved):

PR #151, opened 2024-09-16
Enable multiple resolutions - first step

lukasm91:
This PR is a first step towards enabling running multiple resolutions (required for transi).

- properly implementing `trans_end`/`trans_release`.
- remove all the flattened arrays (in a first step replace them with `ASSOCIATE` to make changes less intrusive, we can clean the associates up in a second step)
- propagate `ncur_resol` to fft and gemms to be able to cache multiple plans
- clean up graphs/plan caches for FFT and gemms

**We have to revert the ectrans-benchmark program before merge.**


lukasm91:
Present table in my sample program (currently "ectrans-benchmark", to be removed again before merge) after trans_release and before trans_end:

host:0x73d2840 device:0x331a00000 size:1180032 presentcount:0+1 line:46 name:alloc%ptr(:)
allocated block device:0x331a00000 size:1433600 thread:1

and after trans_end
Present table dump for device[1]: NVIDIA Tesla GPU 0, compute capability 9.0, threadid=1
Hint: specify 0x800 bit in NV_ACC_DEBUG for verbose info.
...empty...

samhatfield:
Nice work @lukasm91! This will probably take a while to review properly. In the mean time, has something gone wrong with a rebase? I see duplicated commits here from develop.

lukasm91:
I think it is because I was merging with master. Do you prefer rebases? If yes, I might restructure the commits into three new commits or so. And yes, there are conflicts after the merge today, but I can easily resolve those.

samhatfield:
Personally I have no preference of rebasing over merging, as long as commits aren't duplicated. It looks like that's the case, but perhaps I just don't understand how Github presents merging.

lukasm91:
Good question, I am not sure what it was showing; maybe it was because I was starting from Willems branch (which was merged in the meantime), and then I merged develop later. In any case, I quickly restructured the commits into three commits (from which the last commit we will remove before merge; this is just for show-casing the usage). It is still correct that Willems transi branch can just be merged into this and it will work (except vordivtouv/uvtovordiv, which will we have to fix separately).

src/programs/ectrans-benchmark.F90
```diff
@@ -194,6 +194,7 @@ program ectrans_benchmark
 integer(kind=jpim) :: jend_uder_EW = 0
 integer(kind=jpim) :: jbegin_vder_EW = 0
 integer(kind=jpim) :: jend_vder_EW = 0
+integer(kind=jpim) :: nresol1, nresol2
```
lukasm91, line 197: just adding a comment here such that we don't forget to revert this file before merging.

src/programs/ectrans-benchmark.F90
```diff
@@ -397,12 +400,601 @@ program ectrans_benchmark
...
+deallocate(ztstep)
+deallocate(ztstep1)
+deallocate(ztstep2)
+deallocate(nloen)
+
+!===================================================================================================
+
+call trans_release(nresol1)
+call trans_release(nresol2)
+call acc_present_dump
```
wdeconinck, line 981: `acc_present_dump` is probably a debugging leftover

wdeconinck (commented):
This is a huge first step! It looks really nicely done!

Question on the ASSOCIATE statements. Are they required for OpenACC to work or is it to introduce minimal changes?

The ectrans-benchmark looks a little bit too inflated to my liking, though I understand the reasoning. Could we not create a smaller test that just tests that the resolutions are properly deallocated or uniquely using their own memory?

src/programs/ectrans-benchmark.F90
lukasm91: yes, ignore this file. I will completely revert it in the end, this is just to show how I tested it (I essentially copy pasted the whole file)

lukasm91:
> This is a huge first step! It looks really nicely done!
> 
> Question on the ASSOCIATE statements. Are they required for OpenACC to work or is it to introduce minimal changes?
> 
> The ectrans-benchmark looks a little bit too inflated to my liking, though I understand the reasoning. Could we not create a smaller test that just tests that the resolutions are properly deallocated or uniquely using their own memory?

Yes, I will revert the ectrans-benchmark. I just did that to test and show how it could be used. I think we will have to test it properly, through the transi tests, so that will happen. But I wanted to have you a look at how this can now be used.

The ASSOCIATE at this point are really to keep the changes to a minimum. In a second step, we can remove then, and we should then also check if they have a performance impact.

wdeconinck:
What would be the performance cost of having arrays like ZAA in JPRD only, and reusable for single and double precision?
Then TPM_FIELDS_GPU could be in gpu_common.
I want to gauge how much of the memory can be reused (and avoided) between all CPU/GPU/single/double precision backends for the same grid-and-truncation.

wdeconinck:
So then opposite approach, could it be in JPRM and still give desired results for double precision?

lukasm91:
For the legendre coefficients, this could indeed be worth a test. We store one wave number in double precision for a reason, but I don't know if a "mixed precision" gemm would be sufficient in that case, too (Andreas or Sam might know the details). We do the single precision GEMM at a even lower precision, so who knows, it could be sufficient. For sure not something we should do in this PR, but I can only encourage experiments of doing certain things in lower precision.

wdeconinck:
I am happy to hear you think it's an avenue worth exploring. That's what I wanted to get to. This is of course out of scope of this PR.
@samhatfield do you think you would have time to explore this avenue?
Certainly it won't be bit-identical for double precision anymore, but much accuracy is lost...?

samhatfield:
I think the question being asked is: what would be the implications of doing DGEMM for m=0 and SGEMM for m>0 in the dp version of ecTrans, like is already done in the sp version? Is that right?

src/programs/ectrans-benchmark.F90
lukasm91: reverted that commit, then we still have it in the history after merging. now it is passing the tests.

src/trans/gpu/external/gpnorm_trans_gpu.F90
```diff
@@ -273,8 +274,8 @@ SUBROUTINE GPNORM_TRANS_GPU(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL)
   DO JGL=IBEG,IEND
     IGL = D_NPTRLS(MYSETW) + JGL - 1
     DO JF=1,IF_FS
-      ZAVE(JF,JGL)=ZAVE(JF,JGL)*F_RW(IGL)/G_NLOEN(IGL)
-      !write(iunit,*) 'aver inside ',JF,IF_FS,IGL,ZAVE(JF,JGL), F_RW(IGL), G_NLOEN(IGL),ZMINGPN(JF),ZMAXGPN(JF)
+      ZAVE(JF,JGL)=ZAVE(JF,JGL)*F%RW(IGL)/G_NLOEN(IGL)
```
samhatfield, line 277: Do we need to `COPY(F%RW)` somewhere above?

src/trans/gpu/internal/fsc_mod.F90
```diff
@@ -182,6 +183,7 @@ SUBROUTINE FSC(ALLOCATOR,HFSC,PREEL_COMPLEX, KF_FS, KF_UV, KF_SCALARS, KUV_OFFSE
 
 !*       2.1      U AND V.
 
+ILOEN_MAX = MAXVAL(G_NLOEN)
```
samhatfield, line 186: Do we need to `COPY` this in for the parallel region below?

lukasm91: yes, I think so, either we have to drop DEFAULT(NONE), then it is firstprivate, or we have to add it explicitly, otherwise it won't compile on Cray (which I think is the standard compliant way)

samhatfield: Okay, let me suggest the change so you can consider it for merging.

src/trans/gpu/internal/spnsde_mod.F90
```diff
@@ -103,7 +105,7 @@ SUBROUTINE SPNSDE(KF_SCALARS,PEPSNM,PF,PNSD)
 #ifdef OMPGPU
   !$OMP TARGET TEAMS DISTRIBUTE PARALLEL DO
   !! DEFAULT(NONE) PRIVATE(IJ) &
-  !!$OMP&   SHARED(KM,F,ZN,ZEPSNM,KMLOC)
+  !!$OMP&   SHARED(KM,ZN,ZEPSNM,KMLOC)
```
samhatfield, line 108: `ZEPSNM` isn't used -> I guess you left this in because you thought it might be needed for OpenMP offload for some reason?

src/trans/gpu/internal/trmtol_pack_unpack.F90
```diff
@@ -251,16 +258,17 @@ SUBROUTINE TRMTOL_UNPACK(ALLOCATOR,HTRMTOL_UNPACK,FOUBUF,PREEL_COMPLEX,KF_CURREN
...
 #ifdef ACCGPU
 !$ACC PARALLEL LOOP PRIVATE(IGLG,IOFF_LAT,ISTA,RET_REAL,RET_COMPLEX) FIRSTPRIVATE(KF_CURRENT,&
-!$ACC&              KF_TOTAL,OFFSET_VAR,G_NLOEN_MAX) DEFAULT(NONE) &
+!$ACC&              KF_TOTAL,OFFSET_VAR) DEFAULT(NONE) &
 !$ACC&              ASYNC(1) TILE(32,16,1)
 #endif
 DO KGL=1,D_NDGL_FS
   DO JF=1,KF_CURRENT
-    DO JM=0,G_NLOEN_MAX/2
+    DO JM=0,ILOEN_MAX/2
```
samhatfield, line 271: Same thing here as above. I'm concerned we might break things on AMD if we start missing out variables when `DEFAULT(NONE)` is specified.

src/trans/gpu/internal/dealloc_resol_mod.F90
```diff
@@ -68,7 +78,27 @@ SUBROUTINE DEALLOC_RESOL(KRESOL)
...
+#endif
+#ifdef OMPGPU
+#endif
+
+  ! Empty all fields (none of them has pointers; allocatable arrays implicitly deallocate)
+  D = D_
+  F = F_
+  FG = FG_
+  R = R_
+  G = G_
```
samhatfield, line 99: Would it not be simpler to nullify these pointers? I.e. `R => NULL()`?

src/trans/gpu/internal/fsc_mod.F90
lukasm91: fixed, will commit it in a moment.

src/trans/gpu/internal/spnsde_mod.F90
lukasm91: I believe this is an artifact from earlier times. Turns out that PEPSNM is passed as an argument, I think there was a time when this was probably not passed as an argument. I removed it and added PEPSNM to the clause, same as OpenACC.

src/trans/gpu/external/gpnorm_trans_gpu.F90
lukasm91: Good point. Let me do it the same way as for the other variables for now and just put F_RW here, add it to the ASSOCIATES and the initial data clause. I think in general this should only be in a present clause because it is set up in setup_trans.

src/trans/gpu/internal/dealloc_resol_mod.F90
lukasm91: This wouldn't be that easy. Yes, you can do that, but you have to go through the arguments of D and deallocate the allocatable arrays, because they are pointing to the DISTR_RESOL and friends class, which is an allocatable of many such D. When you just nullify the pointers here, it will nullify this pointer, but not the content.
I think, this way is easier, especially if we think that any of the classes might get new allocatables, it is one place less to modify.

lukasm91:
==> took care of the comments so far.

src/trans/gpu/internal/dealloc_resol_mod.F90
samhatfield: So what happens to e.g. `D%MYMS` here? In the CPU version, there's an explicit `DEALLOCATE(D%MYMS)`, for example. Do we not need to keep these deallocations also in the GPU version so that host-side versions are correctly destroyed?

Also I noticed `D%MYMS` is referenced below here on line 107 -> is that safe?

lukasm91: You are right, we have to do it after the if condition.  But yes, the deallocate is not needed, and my impression is that we also want to avoid it if possible. 
https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/common/internal/tpm_distr.F90
All members distr_type are allocatable, that means they deallocate automatically.

samhatfield: Okay cool, I think I get it now. Equality is a bit weird with Fortran pointers. So as I understand it, `D = D_` is saying "set whatever `D` is pointing to to `D_`. In other words it does `DISTR_RESOL(NCUR_RESOL) = D_`. The old value of `DISTR_RESOL(NCUR_RESOL)` contained allocated members. But because those are now out of scope, and are `ALLOCATABLE`, they are automatically deleted.

wdeconinck:
> I think the question being asked is: what would be the implications of doing DGEMM for m=0 and SGEMM for m>0 in the dp version of ecTrans, like is already done in the sp version? Is that right?

That's right.

samhatfield:
> > I think the question being asked is: what would be the implications of doing DGEMM for m=0 and SGEMM for m>0 in the dp version of ecTrans, like is already done in the sp version? Is that right?
> 
> That's right.

For forecasting, it would probably be fine. The real concern is assimilation. With such a change, the tangent-linear/adjoint correspondence would be degraded, and the minimisation would potentially be compromised. I actually tried this experiment a few years ago (double-precision 4D-Var but where tangent-linear/adjoint has single-precision Legendre transforms (except for m = 0 in direct transform)). The results were ... inconclusive. It might be possible, but it needs an assimilation scientist to give the thumbs up.

So in short, no I don't think we can make this change as it would have scientific consequences for data assimilation.

wdeconinck:
I agree with you this definitely is a strong deviation. I am happy you explored this in the past. Let's park this idea somewhere far away :)

src/trans/gpu/internal/dealloc_resol_mod.F90
wdeconinck: `= /= =>` 😜

src/trans/gpu/algor/hicblas_gemm.hip.cpp
```diff
@@ -274,8 +274,8 @@ void hipblas_dgemm_wrapper(char transa, char transb, int m, int n, int k,
     op_t2 = HIPBLAS_OP_T;
 
   if (!hip_alreadyAllocated_dgemm_handle) {
-    hipblasCreate(&handle_hip_dgemm);
-    hip_alreadyAllocated_dgemm_handle = true;
+    HICBLAS_CHECK(hipblasCreate(&handle_hip_dgemm));
+    hip_alreadyAllocated_sgemm_handle = true;
```
wdeconinck, line 278: should this not be `dgemm` instead of `sgemm`?

lukasm91: of course! good catch... I will overwrite the merge commit. Isn't there a rule of thumb to never use two variables that only differ in one character?

src/trans/gpu/algor/hicfft.hip.cpp
```diff
@@ -43,57 +46,100 @@ template <class Type, hipfftType Direction> class hicfft_plan {
...
   }
   void set_stream(hipStream_t stream) {
-    fftSafeCall(hipfftSetStream(handle, stream));
+    fftSafeCall(hipfftSetStream(*handle_ptr, stream));
   }
   hicfft_plan(hipfftHandle handle_, int offset_)
-      : handle(handle_), offset(offset_) {}
+      : handle_ptr(new hipfftHandle{handle_},
+                   [](auto ptr) {
+                     fftSafeCall(cufftDestroy(*ptr));
```
samhatfield, line 63:
```suggestion
                     fftSafeCall(hipfftDestroy(*ptr));
```

src/trans/gpu/algor/growing_allocator_mod.F90
```diff
@@ -89,4 +83,20 @@ SUBROUTINE REGISTER_FREE_C(ALLOC_C, FREE_FUNC_C) BIND(C, NAME="growing_allocator
 
   END SUBROUTINE
 
+  SUBROUTINE DESTROY_GROWING_ALLOCATOR(ALLOC)
+    USE ISO_C_BINDING, ONLY: C_SIZE_T
+    IMPLICIT NONE
+    TYPE(GROWING_ALLOCATION_TYPE) :: ALLOC
+    INTEGER :: I
+    IF (ALLOCATED(ALLOC%PTR)) THEN
```
samhatfield, line 91: I don't think this is strictly compliant - only `ALLOCATABLE` objects can be passed to `ALLOCATED`. Somehow nvfortran doesn't mind, but this line fails when compiling with Cray.

You can of course check whether a pointer is `ASSOCIATED`. I suppose this is what you really want to do here?

samhatfield:
In addition to the comments above there are two remaining issues with this branch (both the Cray compiler's fault - not @lukasm91's):
- When `D_member` (where `D_member => D%member`) is referenced in an OpenACC region, the parent object must also be present on device. I.e., you must add `D` to the list of `PRESENT` variables in the enclosing data region. This affects files
  - ledir_mod.F90
  - leinv_mod.F90
  - prfi1b_mod.F90
  - spnsde_mod.F90
  - trltom_pack_unpack.F90
  - trmtol_pack_unpack.F90
  - updspb_mod.F90
  - uvtvd_mod.F90
  - vdtuv_mod.F90
- Internal compiler error when building ectrans-benchmark.F90.o:
  ```
  [ 97%] Building Fortran object ectrans/src/programs/CMakeFiles/ectrans-benchmark-gpu-dp.dir/ectrans-benchmark.F90.o
  Global is external, but doesn't have external or weak linkage!
  ptr @"str2int$ectrans_benchmark_" 
  Global is external, but doesn't have external or weak linkage! 
  ptr @"print_help$ectrans_benchmark_"
  LLVM ERROR: Broken module found, compilation aborted!
  ftn-2116 ftn: INTERNAL
  ```

samhatfield:
The second error above only occurs when building the GPU version, and can be resolved by moving declarations of `str2int` and `print_help` above their first usage. Weird...

samhatfield:
Final issue compiling with Cray on AMD:
```
ectrans/src/programs/ectrans-benchmark.F90, Line = 1184
Unsupported OpenACC construct INIT
```
This is actually an issue with PR #166. As mentioned there, `ACCGPU` wasn't defined previously so this code was skipped. Now that we use `_OPENACC`, the line is passed to the compiler, and it fails. Cray doesn't support `!$ACC INIT`?

wdeconinck:
https://www.openacc.org/sites/default/files/inline-files/OpenACC.2.7.pdf section 2.14.1 (p62) is about !$acc init

samhatfield:
> https://www.openacc.org/sites/default/files/inline-files/OpenACC.2.7.pdf section 2.14.1 (p62) is about !$acc init
> 
> If CrayFortran does not support this, then probably following guard should be used? `#if defined(_OPENACC) && !defined(_CRAYFTN)`

Yeah that should do the trick.

src/trans/gpu/external/setup_trans.F90
```diff
@@ -481,244 +468,83 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
 #ifdef OMPGPU
+  WRITE(NOUT,*) 'Using OpenMP offloading'
 #endif
-#ifdef ACCGPU
-!$ACC ENTER DATA COPYIN(D_OFFSETS_GEMM1,D_OFFSETS_GEMM2)
-#endif
-
-D_NUMP=D%NUMP
-D_NDGL_FS=D%NDGL_FS
+WRITE(NOUT,'(A10,":",I9,"B")') 'FG%ZAS', SIZEOF(FG%ZAS)
```
samhatfield, line 509: `SIZEOF` isn't portable so must be replaced. Would printing the `SIZE` be a good substitute for this diagnostic?

reuterbal:
> > https://www.openacc.org/sites/default/files/inline-files/OpenACC.2.7.pdf section 2.14.1 (p62) is about !$acc init
> > If CrayFortran does not support this, then probably following guard should be used? `#if defined(_OPENACC) && !defined(_CRAYFTN)`
> 
> Yeah that should do the trick.

According to `man intro_openacc`, Cray supports `acc_init`. Would that be an alternative?

wdeconinck:
> Cray supports acc_init. Would that be an alternative?
If it works, then yes :)

samhatfield:
`acc_init` must take a device type argument. Can I just do
```
call acc_init(acc_get_device_type())
```
?

src/trans/gpu/algor/hicblas_mod.F90
```diff
@@ -63,10 +63,17 @@ SUBROUTINE HIP_SGEMM_BATCHED( &
     TYPE(C_PTR), INTENT(IN), VALUE :: ALLOC
   END SUBROUTINE HIP_SGEMM_BATCHED
 END INTERFACE
+INTERFACE
+  PUBLIC CLEAN_GEMM
```
samhatfield, line 67: I don't think this is a valid location to specify `PUBLIC`. In fact I think you can just delete this line.

src/trans/gpu/algor/growing_allocator_mod.F90
lukasm91: Thanks, nice find. Yes this should be associated (as it is done in the other parts of the code :) ). Will fix it in a moment ..

src/trans/gpu/external/setup_trans.F90
lukasm91: I grepped for SIZEOF, but I guess only in the internal folder. If you don't mind, I change it to SIZE(...)*C_SIZEOF(...) because I think it is good to have bytes here, because not all prints are the same precision.

lukasm91:
> The second error above only occurs when building the GPU version, and can be resolved by moving declarations of `str2int` and `print_help` above their first usage. Weird...

This sounds weird that it only shows up now, but we ICE can be caused as surprizing side effects ... I moved the code now.

src/trans/gpu/external/setup_trans.F90
samhatfield: Sounds good!

lukasm91:
Following changes have been done now:
- The suggestions are applied
- !$acc init => acc_init is done (I am checking if this is equivalent but I am almost sure it is)
- I went through the files @samhatfield  suggested and added the D/R/... as I have seen usage, this needs another test of course.

One general suggestion: If you have more comments, even if they apply to multiple places, can you just add  a comment in one case, then I can reply one-by-one and we have threads to discuss. This might be easier.

src/trans/gpu/external/setup_trans.F90
```diff
@@ -481,244 +468,83 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
 #ifdef OMPGPU
+  WRITE(NOUT,*) 'Using OpenMP offloading'
 #endif
-#ifdef ACCGPU
-!$ACC ENTER DATA COPYIN(D_OFFSETS_GEMM1,D_OFFSETS_GEMM2)
-#endif
-
-D_NUMP=D%NUMP
-D_NDGL_FS=D%NDGL_FS
+WRITE(NOUT,'(A10,":",I9,"B")') 'FG%ZAS', C_SIZEOF(FG%ZAS(1,1,1))*SIZE(FG%ZAS)
```
samhatfield, line 509: Could you add `C_SIZEOF` to the list of imported variables from `ISO_C_BINDING` above (line 106)?

src/trans/gpu/internal/fsc_mod.F90
```diff
@@ -104,7 +105,7 @@ SUBROUTINE FSC(ALLOCATOR,HFSC,PREEL_COMPLEX, KF_FS, KF_UV, KF_SCALARS, KUV_OFFSE
 
 #ifdef ACCGPU
 !$ACC DATA &
-!$ACC& PRESENT(D_NPTRLS,D_NSTAGTF,PREEL_COMPLEX,F_RACTHE,G_NMEN,G_NLOEN, G_NLOEN_MAX, R_NSMAX)
+!$ACC& PRESENT(D_NPTRLS,D_NSTAGTF,PREEL_COMPLEX,F_RACTHE,G_NMEN,G_NLOEN,  R_NSMAX)
```
samhatfield, line 108:
```suggestion
!$ACC& PRESENT(D,D_NPTRLS,D_NSTAGTF,PREEL_COMPLEX,F,F_RACTHE,G,G_NMEN,G_NLOEN,R,R_NSMAX)
```

src/trans/gpu/internal/uvtvd_mod.F90
```diff
@@ -79,13 +79,14 @@ SUBROUTINE UVTVD(KF_UV,PU,PV,PVOR,PDIV)
...
+ASSOCIATE(D_NUMP=>D%NUMP, R_NTMAX=>R%NTMAX, D_MYMS=>D%MYMS, ZEPSNM=>FG%ZEPSNM)
 
 !*       1.    COMPUTE U V FROM VORTICITY AND DIVERGENCE.
 !              ------------------------------------------
 
 #ifdef ACCGPU
 !$ACC DATA &
-!$ACC& PRESENT(D_MYMS,D_NUMP,R_NTMAX) &
+!$ACC& PRESENT(D,D_MYMS,D_NUMP,R,R_NTMAX) &
 !$ACC& PRESENT(ZEPSNM,PU,PV,PVOR,PDIV) ASYNC(1)
```
samhatfield, line 90:
```suggestion
!$ACC& PRESENT(FG,ZEPSNM,PU,PV,PVOR,PDIV) ASYNC(1)
```

samhatfield:
Just a couple more tweaks. PR #169 has slightly screwed up this branch - sorry about that @lukasm91. You'll have to rebase against develop again.

samhatfield:
Unfortunately this branch (with the tweaks I just suggested) doesn't run on MI250x (LUMI-G) so we'll have to rectify that before we can merge this. Firstly I see many instances of this warning:
```
ACC: libcrayacc/acc_datatype.c:1799 CRAY_ACC_WARNING - Asynchronous transfer not supported for deep transfers - falling back to a synchronous transfer
```
and eventually the benchmark program crashes with `HIPFFT_PARSE_ERROR`. My suspicion is that the Cray compiler still doesn't like it when you put derived types in `DATA` statements. We encountered this before. Of course this branch does away with flattened derived types so it's natural this issue would come up again. I'll have a think if there's a minimally invasive workaround we can implement for Cray/AMD.

lukasm91:
I will rebase, yes. But regarding the problem with Cray Compilers - isn't this just a warning, could we put ASYNC into ifdefs for the problematic instances? I can't reallz help with the parse error, but I don't think this can be related to derived types? Have you checked if it is actually using the not inplace FFT, because I think this caused the same error?

samhatfield:
Yeah, I'll take some time this week to look closer and see what's going on.

samhatfield:
And LUMI just announced a 4-day service break :)

samhatfield:
Sorry, I'm not making any progress understanding the issue on Cray systems. I'm using Frontier to debug this. If I build ecTrans and FIAT "manually" through CMake invocations etc., everything seems to work. If I build both using an ecBundle-based system, as I usually do, I get the warning mentioned above and either this error message:
```
ACC: libcrayacc/acc_present.c:868 CRAY_ACC_ERROR - Error placing variable 'f' (0x1a5bf80 - 0x1a5c480) from ectrans/src/trans/gpu/generated/ectrans_gpu_sp/external/setup_trans.F90:525 on the device. Another instance of the address range (0x1a5bf80 - 0x1a5c480) is already present.
```
OR it does "work" but gives spectral error norms of ~ 1.0, which looks like at some point the spectral fields are just being zeroed. I've seen both these behaviours while debugging but I don't understand fully the conditions under which each one occurs. Annoying.

src/trans/gpu/external/setup_trans.F90
```diff
@@ -481,244 +468,83 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
-DO I=1,SIZE(F%RACTHE)
-  F_RACTHE(I)=F%RACTHE(I)
-END DO
-
 #ifdef ACCGPU
-!$ACC ENTER DATA COPYIN(R_NSMAX,R_NTMAX,R_NDGL,R_NDGNH,D_NSTAGT0B,D_NSTAGT1B,&
-!$ACC&                  D_NPNTGTB1,D_NPROCL,D_NUMP,D_NDGL_FS,D_MYMS,D_NASM0,D_NSTAGTF,D_MSTABF,&
-!$ACC&                  D_NPNTGTB0,D_NPROCM,D_NPTRLS,G_NDGLU,G_NMEN,G_NMEN_MAX,G_NLOEN,&
-!$ACC&                  G_NLOEN_MAX,F_RW,F_RLAPIN,F_RACTHE)
+!$ACC ENTER DATA COPYIN(R) ASYNC(1)
```
samhatfield, line 524:
```suggestion
!$ACC ENTER DATA COPYIN(R,R%NSMAX,R%NTMAX,R%NDGL,R%NDGNH) ASYNC(1)
```

src/trans/gpu/external/setup_trans.F90
```diff
@@ -481,244 +468,83 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
+!$ACC ENTER DATA COPYIN(F,F%RLAPIN,F%RACTHE,F%RW) ASYNC(1)
+!$ACC ENTER DATA COPYIN(FG,FG%ZAA,FG%ZAS,FG%ZEPSNM) ASYNC(1)
+!$ACC ENTER DATA COPYIN(D,D%MYMS,D%NPNTGTB0,D%NPNTGTB1,D%NSTAGT0B,D%NSTAGT1B,D%NSTAGTF,D%NPROCM)&
+!$ACC&           COPYIN(D%NPTRLS,D%MSTABF,D%NASM0,D%OFFSETS_GEMM1,D%OFFSETS_GEMM2) ASYNC(1)
+!$ACC ENTER DATA COPYIN(G,G%NDGLU,G%NMEN,G%NLOEN) ASYNC(1)
```
samhatfield, line 529:
```suggestion
!$ACC ENTER DATA COPYIN(D,D%NUMP,D%MYMS,D%NPNTGTB0,D%NPNTGTB1,D%NSTAGT0B,D%NSTAGT1B,D%NSTAGTF,D%NPROCM,D%NPROCL)&
!$ACC&           COPYIN(D%NPTRLS,D%MSTABF,D%NASM0,D%OFFSETS_GEMM1,D%OFFSETS_GEMM2,D%NDGL_FS) ASYNC(1)
!$ACC ENTER DATA COPYIN(G,G%NDGLU,G%NMEN,G%NMEN,G%NLOEN) ASYNC(1)
```

samhatfield:
Finally seem to be making some progress with the Cray compiler. It seems Cray doesn't do the same thing as NVHPC when copying derived types. I think you have to explicitly copy all the members, as well as the parent itself, hence the suggestions above. With these changes I can at least run the following command successfully on MI250x:
```
ECTRANS_USE_MPI=0 ./ectrans/build/bin/ectrans-benchmark-gpu-sp --norms --niter 100 --vordiv
```
which believe me I am very happy about...

I'll report back on results from proper testing soon.

samhatfield:
Yes 🎉 just ran on 8 MI250x. Good error norms. As before, I'm seeing thousands of these warnings
```
ACC: libcrayacc/acc_datatype.c:1799 CRAY_ACC_WARNING - Asynchronous transfer not supported for deep transfers - falling back to a synchronous transfer
```
I was able to track down a few. Surprisingly the warnings are not emitted on data statements but rather on parallel regions, even ones that don't have any explicit copys in or out. Both parallel regions in LEINV emit this warning for example, which go away if you remove `ASYNC(1)`.

We can't have thousands of identical warnings polluting STDERR, so I'm afraid we'll have to guard all of the ones responsible with `#ifdef _CRAYFTN`s to remove `ASYNC(1)`, until we understand properly what is causing the warning.

lukasm91:
Thanks a lot for debugging this and glad you made progress! :tada:   Weird, that it happen for the parallel regions, but agree puting ifdefs seems like a good workaround. Unfortunately it is up to you to add those ifdefs, because I can't do that - I will rebase this branch onto master next week.

samhatfield:
Absolutely - forget those suggestions above: I will instead submit a PR to cleanup-part1 for you to merge. Then I think we will finally be able to merge this.

samhatfield:
See PR https://github.com/lukasm91/ectrans/pull/1 @lukasm91.

BTW if you rebase this branch against develop again you will pick up our new GPU-based CI tests.

lukasm91:
I merged your branch and master. Let's see if this passes tests...  @samhatfield  may you be able to help me run the build-hpc tests? 

As additional tests, I also tested doing multiple alloc/dealloc; the final present table is empty. I also enable the cleanup in transi for GPU (@wdeconinck). 

@samhatfield Note I did a few minor tweaks to your changes: I removed copyin scalars in setup trans for non-Cray because they do modifications in the present table that are not ideal. Also, I modified the ifdefs to really only have the ASYNC(1) inside.

src/trans/gpu/internal/ledir_mod.F90
```diff
@@ -248,10 +248,11 @@ SUBROUTINE LEDIR(ALLOCATOR,ZINPS,ZINPA,ZINPS0,ZINPA0,ZOUT,ZOUT0,POA1,KF_FS)
...
 #endif
 #ifdef ACCGPU
-#ifdef _CRAYFTN
-    !$ACC PARALLEL LOOP COLLAPSE(2) PRIVATE(KM,IA,J) FIRSTPRIVATE(KF_FS,IOUT_STRIDES0,IOUT0_STRIDES0) DEFAULT(NONE)
+    !$ACC PARALLEL LOOP COLLAPSE(2) PRIVATE(KM,IA,J) FIRSTPRIVATE(KF_FS,IOUT_STRIDES0,IOUT0_STRIDES0) DEFAULT(NONE) &
+#ifndef _CRAYFTN
+    !$ACC& ASYNC(1)
 #else
-    !$ACC PARALLEL LOOP COLLAPSE(2) PRIVATE(KM,IA,J) FIRSTPRIVATE(KF_FS,IOUT_STRIDES0,IOUT0_STRIDES0) DEFAULT(NONE) ASYNC(1)
+    !$ACC&
```
samhatfield, line 255: Is this necessary in order to terminate the `&` on the preceding line?

samhatfield:
I'll take a look at that failing build-hpc suite, and also re-test everything on Cray.

src/trans/gpu/internal/ledir_mod.F90
lukasm91: unfortunately yes.

samhatfield:
Everything works with Cray (testing on Frontier now). The failing build-hpc action above is simply because we haven't yet authorised PRs from external branches to run on our HPC yet. I can trigger this action manually once the branch is merged.

Great work @lukasm91! This was a real marathon 😅 Many thanks as always for your diligence.

@wdeconinck any final objections before I merge into develop?

PR #156, opened 2024-09-24
CMake fixes for test linker and OpenACC search

reuterbal:
Two small CMake fixes:
- Use the Fortran linker for ectrans_test_setup_trans0
- Only apply the OpenACC search workaround for CMake < 3.25 if the corresponding option has been requested.


wdeconinck (approved):
Thank you for ironing out the corner cases!

PR #155, opened 2024-09-19
Make in-place FFT optional

samhatfield:
It seems that hipFFT has a problem with in-place FFTs. Running on LUMI-G (MI250x), this error is thrown at runtime:

```
GPU runtime error at 1
GPU runtime error in file '/pfs/lustrep3/scratch/project_465000454/hatfield/jube-suites/ectrans/ectrans_gpu/000005/000000_clone/work/ectrans-bundle/source/ectrans/src/trans/gpu/algor/hicfft.hip.cpp'
GPU runtime error at 2
GPU runtime error line '48'
GPU runtime error at 3
GPU runtime error 6: HIPFFT_EXEC_FAILED
terminating!
ABOR1     [PROC=1,THRD=1] from [/pfs/lustrep3/scratch/project_465000454/hatfield/jube-suites/ectrans/ectrans_gpu/000005/000000_clone/work/ectrans-bundle/source/ectrans/src/trans/gpu/algor/hicfft.h +48] : Error in FFT
```

Some investigations with @PaulMullowney showed that if you create a separate buffer for the output, the error vanishes.

With this change we are now finally able to run the optimised GPU version of ecTrans (now _the_ GPU version of ecTrans) on AMD GPUs, at least with a single MI250x GPU on LUMI-G. Multi-GPU runs are still a work in progress.

This PR makes in-place FFT compile-time adjustable. For now, in-place is enabled for cuFFT, but disabled for hipFFT.

@PaulMullowney and I will see what we can do to flag this to hipFFT developers so we can eventually remove this option.

Note that earlier `HIPFFT_PARSE_ERROR`s no longer appear with ROCm 6.0.2, which is the new default for LUMI-G.

Note that this requires testing on an Nvidia GPU. Hopefully we have not ruined things there.

This PR builds on https://github.com/ecmwf-ifs/ectrans/pull/150 so best to merge that one first.

PR #157, opened 2024-10-02
Add support for AMD GPUs

samhatfield:
Finally, we are able to run the develop branch on AMD GPUs.

Many many thanks to @PaulMullowney for getting this done.

Note that #155 should be merged first.

### DIFF
--- a/src/trans/gpu/internal/trltom_mod.F90
+++ b/src/trans/gpu/internal/trltom_mod.F90
@@ -94,7 +94,8 @@ CONTAINS
     USE TPM_DISTR,              ONLY: D, NPRTRW, NPROC, MYPROC, MYSETW
     USE TPM_GEN,                ONLY: LSYNC_TRANS, NERR
 #if ECTRANS_HAVE_MPI
-    USE MPI_F08,                ONLY: MPI_COMM, MPI_FLOAT, MPI_DOUBLE, MPI_ALLTOALLV
+    USE MPI_F08,                ONLY: MPI_COMM, MPI_FLOAT, MPI_DOUBLE
+    ! Missing: MPI_ALLTOALLV on purpose due to cray-mpi bug (see https://github.com/ecmwf-ifs/ectrans/pull/157)
 #endif
     USE TPM_STATS,              ONLY: GSTATS => GSTATS_NVTX
     USE BUFFERED_ALLOCATOR_MOD, ONLY: BUFFERED_ALLOCATOR, ASSIGN_PTR, GET_ALLOCATION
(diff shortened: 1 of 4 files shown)


samhatfield:
Note that graph support has *not* been tested on AMD GPUs yet.

src/trans/gpu/internal/trltom_mod.F90
```diff
@@ -94,7 +94,7 @@ SUBROUTINE TRLTOM(ALLOCATOR,HTRLTOM,PFBUF_IN,PFBUF,KF_FS)
     USE TPM_DISTR,              ONLY: D, NPRTRW, NPROC, MYPROC, MYSETW
     USE TPM_GEN,                ONLY: LSYNC_TRANS, NERR
 #if ECTRANS_HAVE_MPI
-    USE MPI_F08,                ONLY: MPI_COMM, MPI_FLOAT, MPI_DOUBLE, MPI_ALLTOALLV
```
PaulMullowney, line 97: This might need __HIP_PLATFORM_AMD__ protection?

samhatfield: I did consider that. Then I noticed that we don't explicitly import `MPI_ALLTOALLV` in `TRMTOL`, so it doesn't seem to be an issue.

wdeconinck:
For future reference and my understanding of this bizarre MPI related change.

Removing `MPI_ALLTOALLV` from the `USE` statement means that not the MPI_F08-module-mangled symbol (e.g. `_mpi_f08_mod_mpi_alltoallv` is being used, but rather an interface-less MPI_F77 `_mpi_alltoallv` symbol, which by luck(?) happens to be linked in as well, but that is not guaranteed necessarily.

Also there is **no** checking on arguments because of any missing interface, meaning we can pass whatever and any number of arguments. It so happens (again by luck?) that the signature matches on a byte-level, meaning that `TYPE(MPI_COMM)` argument is compatible with the `INT` argument expected for the MPI_F77 API.

I don't consider this as a final solution, but good enough for now, seeing that it works experimentally on some of our tested platforms / compilers / mpi libraries, and seeing that the GPU adaptations are very much in flux.

samhatfield:
Exactly. Also, the implication is that there is something wrong with `_mpi_f08_mod_mpi_alltoallv` on the only Cray platform we have access to and care about: LUMI-G. When you _do_ `USE` `MPI_ALLTOALLV`, presumably the `_mpi_f08_mod_mpi_alltoallv` symbol is used, as you say, and this version somehow doesn't work correctly when device pointers are passed for the buffer arguments. Hence, GPU-aware communication isn't possible -> the outcoming data is all NaNs.

What would be great would be to confirm @wdeconinck's hypothesis by inspecting the object code generated when you import and don't import `MPI_ALLTOALLV`.

PR #158, opened 2024-10-03
Pin buffers in benchmark

lukasm91:
We should pin buffers to have fast transfers on x86 machines.


wdeconinck:
Right I was just thinking about this the other day that the removal of the "-pinned" flag in the library must have disabled this for the benchmark as well. This is an elegant solution.

marsdeno (approved):
All good from my point of view, previously tested as working.

PR #161, opened 2024-10-10
Remove non-standard SIZEOF from GPU subtree

### LINKED ISSUE #160: Remove non-standard SIZEOF from gpu subtree
There are numerous instances of the non-standard `SIZEOF` function in the GPU source tree, as identified in #155. As discussed in https://github.com/ecmwf-ifs/ectrans/pull/89, these need to be removed.

samhatfield:
Resolves #160.

Needs testing on Nvidia and AMD platforms.


src/trans/gpu/algor/ext_acc.F90
```diff
@@ -268,7 +268,7 @@ subroutine ext_acc_create(ptrs, stream)
     num_ranges = get_common_pointers(ptrs, common_ptrs)
 
     do i = 1, num_ranges
-      call c_f_pointer(common_ptrs(i)%ptr, pp, shape=[common_ptrs(i)%sz/sizeof(pp(1))])
+      call c_f_pointer(common_ptrs(i)%ptr, pp, shape=[common_ptrs(i)%sz/storage_size(pp(1))/8])
```
wdeconinck, line 271:
```suggestion
      call c_f_pointer(common_ptrs(i)%ptr, pp, shape=[common_ptrs(i)%sz/(storage_size(pp(1))/8)])
```
And similar edits required further where `storage_size` is in the denominator

wdeconinck (changes requested):
Thanks for taking this up!

src/trans/gpu/algor/ext_acc.F90
samhatfield: Very good point. Somehow I forgot that `(x/y)/z` isn't the same as `x/(y/z)`.

src/trans/gpu/internal/ftdir_mod.F90
```diff
@@ -98,7 +98,7 @@ SUBROUTINE FTDIR(ALLOCATOR,HFTDIR,PREEL_REAL,PREEL_COMPLEX,KFIELD)
     PREEL_COMPLEX => PREEL_REAL
 #else
     CALL ASSIGN_PTR(PREEL_COMPLEX, GET_ALLOCATION(ALLOCATOR, HFTDIR%HREEL_COMPLEX),&
-      & 1_C_SIZE_T, INT(KFIELD*D%NLENGTF*SIZEOF(PREEL_COMPLEX(1)),KIND=C_SIZE_T))
+      & 1_C_SIZE_T, INT(KFIELD*D%NLENGTF*STORAGE_SIZE(PREEL_COMPLEX(1))/8,KIND=C_SIZE_T))
```
lukasm91, line 101:
```suggestion
      & 1_C_SIZE_T, 1_C_SIZE_T*KFIELD*D%NLENGTF*STORAGE_SIZE(PREEL_COMPLEX(1))/8)
```
It might be fine, not sure about the type of NLENGTF, but make sure you are not introducing new potential overflows. The value you compute is going to be 8X larger than before. Since you are changing this lines anyway, might be a good point to make those computations in JPIB or C_SIZE_T.

src/trans/gpu/internal/ftdir_mod.F90
```diff
@@ -36,7 +36,7 @@ FUNCTION PREPARE_FTDIR(ALLOCATOR,KF_FS) RESULT(HFTDIR)
     REAL(KIND=JPRBT) :: DUMMY
 
 #ifndef IN_PLACE_FFT
-    HFTDIR%HREEL_COMPLEX = RESERVE(ALLOCATOR, INT(KF_FS*D%NLENGTF*SIZEOF(DUMMY), KIND=C_SIZE_T))
+    HFTDIR%HREEL_COMPLEX = RESERVE(ALLOCATOR, INT(KF_FS*D%NLENGTF*STORAGE_SIZE(DUMMY)/8, KIND=C_SIZE_T))
```
lukasm91, line 39: also here

src/trans/gpu/internal/ftdir_mod.F90
wdeconinck: note that the iso_c_binding function c_sizeof returns c_size_t and storage_size also accepts a KIND argument alternatively

lukasm91: I think, what you are missing is my initial thought process, which was, that I thought that it already fixes the potental overflows (#162), which it does not, so I am fine with any solution. But: What I wanted to say initially is that the intermediate `KFIELD*D%NLENGTF*STORAGE_SIZE(...)` is 8 times larger than before, so a overflow is more likely than it was before, so we have to make sure that this expression does not overflow, by casing to C_SIZE_T early enough. But if you don't see overflows, you probably can just leave it like this and we can discuss it in #162.

samhatfield:
Okay, I have switched from `STORAGE_SIZE` to `C_SIZEOF`, since the latter returns an answer in bytes and is closer to what we were doing before. I've added some casting here and there to try and keep things consistent. Let me know what you think @lukasm91 @wdeconinck.

src/trans/gpu/internal/ftdir_mod.F90
samhatfield: I've switched to `C_SIZEOF` which I hope will at least not introduce more chances for overflow than currently exist.

wdeconinck (approved):

samhatfield:
Note: still to be tested. I'm on it.

samhatfield:
Successfully tested on LUMI-G up to 8 nodes.

samhatfield:
And successfully tested on JEDI up to 8 nodes.

Note that PR #158 has broken the benchmark program on both Nvidia platforms I use currently: HPC2020/AC (A100) and JEDI (GH200). I have to revert that commit for now. Something we need to investigate @lukasm91. I'll post the error message here in a bit.

samhatfield:
This error
```
+ srun --cpus-per-task=72 /p/project1/jureap5/jube-suites/ectrans/ectrans_gpu/000004/000002_execute/work/build/build/bin/ectrans-benchmark-gpu-sp --norms --truncation 316 --niter 100 --nlev 137 --nfld 1 --vordiv --uvders --scders --meminfo
line 156: cudaLaunchKernel returned status 98: invalid device function

line 156: cudaLaunchKernel returned status 98: invalid device function

line 156: cudaLaunchKernel returned status 98: invalid device function

line 156: cudaLaunchKernel returned status 98: invalid device function

srun: error: jpbot-001-01: tasks 1-2: Exited with exit code 1
srun: error: jpbot-001-01: tasks 0,3: Exited with exit code 1
```
@marsdeno is also aware and is investigating.

PR #162, opened 2024-10-16
Fix overflows up to at least tco1279

lukasm91:
This is one more change that Olivier ported from my branch to his branch, and now I am taking it to develop.

The upper end of the sizes that this allows is certainly more "theoretical" because we can't run simulations with it, but even for real use-cases, we might see overflows without this, and those can be very difficult to spot. With one rank, we can grow to very large domains, with multiple ranks, at some point we run into overflows of the MPI buffers, those are properly diagnosed.

This will conflict with #161 , but should be trivial to resolve.


src/trans/gpu/algor/ext_acc.F90
```diff
@@ -323,12 +321,11 @@ subroutine ext_acc_copyout(ptrs, stream)
 
     do i = 1, num_ranges
       call c_f_pointer(common_ptrs(i)%ptr, pp, shape=[common_ptrs(i)%sz/c_sizeof(pp(1))])
-      !!call acc_copyout_async(pp, common_ptrs(i)%sz, async=stream_act)
-      call acc_copyout(pp, int(common_ptrs(i)%sz))
+      !$acc exit data copyout(pp) async(stream_act)
     enddo
   end subroutine
   subroutine ext_acc_delete(ptrs, stream)
-    use openacc, only: acc_async_sync, acc_delete
```
wdeconinck, line 331: Why remove `acc_delete`? It is still called below.

lukasm91: that is a stupid typo... it should not be called. I had to switch to clauses because acc_delete only takes int32, not int64. This code deallocated nothing :D  I will test if everything properly deallocates after #151 has been merged).

lukasm91: this is fine now. acc_delete is not called - we are using pragmas because they allow for size_t.

lukasm91:
@samhatfield  This is the second last PR from the original GPU branch that is missing and I propose for integration. This should be much more straight forward than the previous one. I am using this for a while now without issues, and it is very useful because the overflows are very hard to find, and it is nice to be able experimenting on a single node/GPU.

samhatfield:
Noted, thanks Lukas. I'll try to find time to look at this in the next week.

src/trans/gpu/external/setup_trans.F90
```diff
@@ -98,7 +98,7 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
 !        R. El Khatib 07-Mar-2016 Better flexibility for Legendre polynomials computation in stretched mode
 !     ------------------------------------------------------------------
 
-USE PARKIND1,        ONLY: JPIM, JPRB, JPRD
+USE PARKIND1,        ONLY: JPIM, JPRB, JPRD, JPIB
```
samhatfield, line 101:
```suggestion
USE PARKIND1,        ONLY: JPIM, JPRB, JPRD
```

Do we need `JPIB` in this scope?

src/trans/gpu/algor/buffered_allocator_mod.F90
```diff
@@ -69,10 +69,11 @@ FUNCTION MAKE_BUFFERED_ALLOCATOR()
     MAKE_BUFFERED_ALLOCATOR%NEXT_BUF = 0
   END FUNCTION MAKE_BUFFERED_ALLOCATOR
 
-  FUNCTION RESERVE(ALLOCATOR, SZ)
+  FUNCTION RESERVE(ALLOCATOR, SZ, WHO)
     IMPLICIT NONE
     TYPE(BUFFERED_ALLOCATOR), INTENT(INOUT) :: ALLOCATOR
     INTEGER(KIND=C_SIZE_T), INTENT(IN) :: SZ
+    CHARACTER(*), INTENT(IN), OPTIONAL :: WHO
```
samhatfield, line 76: Is this argument needed?

src/trans/gpu/internal/ftdir_mod.F90
```diff
@@ -36,7 +36,7 @@ FUNCTION PREPARE_FTDIR(ALLOCATOR,KF_FS) RESULT(HFTDIR)
     REAL(KIND=JPRBT) :: DUMMY
 
 #ifndef IN_PLACE_FFT
-    HFTDIR%HREEL_COMPLEX = RESERVE(ALLOCATOR, INT(KF_FS*D%NLENGTF,KIND=C_SIZE_T)*C_SIZEOF(DUMMY))
+    HFTDIR%HREEL_COMPLEX = RESERVE(ALLOCATOR, 1_JPIB*KF_FS*D%NLENGTF*C_SIZEOF(DUMMY), "HFTDIR%HREEL_COMPLEX")
```
samhatfield, line 39: `JPIB` needs to be imported from `PARKIND_ECTRANS`.

src/trans/gpu/internal/ftinv_mod.F90
```diff
@@ -35,7 +35,7 @@ FUNCTION PREPARE_FTINV(ALLOCATOR,KF_FS) RESULT(HFTINV)
     REAL(KIND=JPRBT) :: DUMMY
 
 #ifndef IN_PLACE_FFT
-    HFTINV%HREEL_REAL = RESERVE(ALLOCATOR, INT(D%NLENGTF*KF_FS,KIND=C_SIZE_T)*C_SIZEOF(DUMMY))
+    HFTINV%HREEL_REAL = RESERVE(ALLOCATOR, 1_JPIB*D%NLENGTF*KF_FS*C_SIZEOF(DUMMY),"HFTINV%HREEL_REAL")
```
samhatfield, line 38: Ditto, need an import for `JPIB` here.

src/trans/gpu/internal/fsc_mod.F90
```diff
@@ -132,7 +132,7 @@ SUBROUTINE FSC(ALLOCATOR,HFSC,PREEL_COMPLEX, KF_FS, KF_UV, KF_SCALARS, KUV_OFFSE
     DO JM=0,R_NSMAX !(note that R_NSMAX <= G_NMEN(IGLG) for all IGLG)
       IGLG    = OFFSET_VAR+KGL-1
       IF (JM <= G_NMEN(IGLG)) THEN
-        IOFF_LAT = KF_FS*D_NSTAGTF(KGL)
+        IOFF_LAT = 1_JPIB*KF_FS*D_NSTAGTF(KGL)
```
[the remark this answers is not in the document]
lukasm91: In the end, I think I am fine with everything, and I can change the casts if needed, as long as it is as consistent as possible. Your suggestions won't work, though. It would need to be
```
        IOFF_LAT = INT(KF_FS,JPIB)*INT(D_NSTAGTF(KGL), JPIB)
```
Or, if there is a left-to-right guarantee, then
```
        IOFF_LAT = INT(KF_FS,JPIB)*D_NSTAGTF(KGL)
```
which then again is very close to my syntax. My preference for 1_JPIB*a*b*c is that if you have several lines, the actual multiplications are not cluttered with integer conversions. 

Which one would you prefer?

src/trans/gpu/algor/buffered_allocator_mod.F90
samhatfield: Okay we can leave it in then.

src/trans/gpu/internal/fsc_mod.F90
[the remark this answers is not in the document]
samhatfield: I think my concern is academic in any case. Whatever happens on the right-hand side of the assignment operator, it is casted to `JPIB` when assigning to `IOFF_LAT` anyway. So I think we can just leave this as it is.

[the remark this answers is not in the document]
lukasm91: I think my assumption is that multiplications are evaluated left-to-right, if that is the case, then we are good, and everything is always multiplied as JPIB. If this is not the case, then we have to cast each individual operand to JPIB before we execute the multiplication, which is possible, but doesn't help readability. I hope this argument makes sense :D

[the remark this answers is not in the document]
samhatfield: Let's just merge but keep this in mind!

samhatfield (approved):

PR #163, opened 2024-10-16
Dumping whole layers

lukasm91:
As discussed, using gather_grid before dumping. The output can easily be parsed through python.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -660,9 +660,9 @@ do jstep = 1, iters+iters_warmup
       allocate(global_field(ngptotg,1))
     endif
     call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgp2(:,1:1,:), 's', noutdump)
-    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,1:nflevg,1,:), 'u', noutdump)
-    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,1:nflevg,2,:), 'v', noutdump)
-    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgp3a(:,1:nflevg,1,:), 't', noutdump)
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,nflevg:nflevg,1,:), 'u', noutdump)
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,nflevg:nflevg,2,:), 'v', noutdump)
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgp3a(:,nflevg:nflevg,1,:), 't', noutdump)
     if (myproc == 1) then
       deallocate(global_field)
     endif
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -210,6 +210,8 @@ character(len=16) :: cgrid = ''
 
 integer(kind=jpim) :: ierr
 
+real(kind=jprb), allocatable :: global_field(:,:)
+
 !===================================================================================================
 
 #include "setup_trans0.h"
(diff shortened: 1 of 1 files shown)


src/programs/ectrans-benchmark.F90
```diff
@@ -652,12 +655,17 @@ program ectrans_benchmark
...
-    ! dump a field to a binary file
-    call dump_gridpoint_field(jstep, myproc, nproma, ngpblks, zgp2(:,1,:),         'S', noutdump)
-    call dump_gridpoint_field(jstep, myproc, nproma, ngpblks, zgpuv(:,nflevg,1,:), 'U', noutdump)
-    call dump_gridpoint_field(jstep, myproc, nproma, ngpblks, zgpuv(:,nflevg,2,:), 'V', noutdump)
-    call dump_gridpoint_field(jstep, myproc, nproma, ngpblks, zgp3a(:,nflevg,1,:), 'T', noutdump)
+  if (ldump_values .and. mod(jstep,10) == 1) then
+    if (myproc == 1) then
+      allocate(global_field(ngptotg,1))
+    endif
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgp2(:,1:1,:), 's', noutdump)
```
samhatfield, line 662: In my [benchmark refactor PR](https://github.com/ecmwf-ifs/ectrans/pull/122/files#diff-bfd9d146dfab7708c079d896a09f7c52c70fa17572bd2c12fddaf6316ff93427R636), I decided to only dump one field. All fields basically go through the same code path so I don't think it's necessary to dump every field. What do you think?

samhatfield:
Testing now.

samhatfield:
This works very nicely! Here's an example snapshot of U after 1,000 iterations on 16 tasks at TCO255. Nice work @lukasm91!

Screenshot 2024-10-18 at 15 41 02

samhatfield (approved):
Approved subject to discussion above on whether to dump all fields or just one.

src/programs/ectrans-benchmark.F90
lukasm91: I want dump_gridpoint_field to take a 3d array, such that it is easy to change when there is need, but I agree that it is a good default to only dump one per field. Let me change to default last level only

src/programs/ectrans-benchmark.F90
```diff
@@ -652,12 +655,17 @@ program ectrans_benchmark
...
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,1:nflevg,1,:), 'u', noutdump)
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,1:nflevg,2,:), 'v', noutdump)
+    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgp3a(:,1:nflevg,1,:), 't', noutdump)
```
lukasm91, line 665:
```suggestion
    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,nflevg:nflevg,1,:), 'u', noutdump)
    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgpuv(:,nflevg:nflevg,2,:), 'v', noutdump)
    call dump_gridpoint_field(jstep, myproc, nproma, global_field, zgp3a(:,nflevg:nflevg,1,:), 't', noutdump)
```

src/programs/ectrans-benchmark.F90
lukasm91: => done. can you agree with this solution?

lukasm91: oh, I think it makes sense to dump them separately, especiall U/V/T because they go through separate code paths (U/V are transformed into VOR/DIV). Also the errors for U/V behave different compared to T: They increase faster in general.

samhatfield: Ah yes, good point. U/V are derived from vor/div so go through a different procedure.

PR #164, opened 2024-10-16
Compact legendre polynomials

lukasm91:
Legendre polynomials don't need to be stored zero-padded, we can just concatenate them with proper zero padding.

E.g. on tco2559, this saves up to almost 40GB per rank (used to be 64 GB, now it is 26 GB). Initially, we did this for Leonardo, because legendre coefficients used almost the whole device memory.


lukasm91:
(convert to draft for the moment ==> it is fnished but we should first merge the other PR, then rebase and review)

lukasm91:
Beside replacing ext_acc with the proper "copy module", we are only left with this PR from the old GPU branch which I now rebased on top of develop. This PR is compacting the legendre polynomials and with this removing zero padding. 

This PR is not expected to interfere with any other PR, as it only touches the GEMMs.  Since I am touching the CUDA interfaces anyway, I also added const to pointers in the interface.

- [x]  Applied clang-format
- [x] Tested that everything is properly deallocated
- [x] Tested large sizes for no overflows.

wdeconinck (approved):
I think this looks all fine and what a nice saving!
Perhaps @samhatfield  can run it in few places to double check all is in order ?

samhatfield:
If I can find the time this week, I will take a look. It might have to wait until next year.

src/trans/gpu/external/setup_trans.F90
```diff
@@ -98,7 +98,7 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
 !        R. El Khatib 07-Mar-2016 Better flexibility for Legendre polynomials computation in stretched mode
 !     ------------------------------------------------------------------
 
-USE PARKIND1,        ONLY: JPIM, JPRB, JPRD
+USE PARKIND1,        ONLY: JPIM, JPRB, JPRD, JPIB
```
samhatfield, line 101: I don't think you need this do you?
```suggestion
USE PARKIND1,        ONLY: JPIM, JPRB, JPRD
```

lukasm91: You are right, this is a leftover from before we moved to the right C_SIZEOF. I like that you always keep an eye on unnecessary "use" statements!

samhatfield:
Sorry, I am gradually getting through this finally :) Will add some more comments soon.

src/trans/gpu/external/setup_trans.F90
```diff
@@ -470,49 +468,56 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
-WRITE(NOUT,'(A10,":",I9,"B")') 'FG%ZAA0', C_SIZEOF(FG%ZAA0(1,1))*SIZE(FG%ZAA0)
-WRITE(NOUT,'(A10,":",I9,"B")') 'FG%ZEPSNM', C_SIZEOF(FG%ZEPSNM(1,1))*SIZE(FG%ZEPSNM)
+WRITE(NOUT,'(A10,":",I11,"B")') 'FG%ZAS', C_SIZEOF(FG%ZAS(1))*SIZE(FG%ZAS)
+WRITE(NOUT,'(A10,":",I11,"B")') 'FG%ZAA', C_SIZEOF(FG%ZAA(1))*SIZE(FG%ZAA)
+WRITE(NOUT,'(A10,":",I11,"B")') 'FG%ZAS0', C_SIZEOF(FG%ZAS0(1,1))*SIZE(FG%ZAS0)
+WRITE(NOUT,'(A10,":",I11,"B")') 'FG%ZAA0', C_SIZEOF(FG%ZAA0(1,1))*SIZE(FG%ZAA0)
+WRITE(NOUT,'(A10,":",I11,"B")') 'FG%ZEPSNM', C_SIZEOF(FG%ZEPSNM(1,1))*SIZE(FG%ZEPSNM)
 
-IF (IMLOC0(1) > 0) THEN
+IF (IMLOC0 > 0) THEN
```
samhatfield, line 520: If I understand correctly, `KMLOC` has the role of "are any of the m's on this rank equal to 0?" Why not make this more transparent, by deleting `IMLOC0`, and replacing this if statement with
```
IF (ANY(D%MYMS == 0)) THEN
```
What do you think?

lukasm91: Yes, exactly, it tries to see if there is a m==0 (because those are computed in double precision). In other places I do `IMLOC0 = FINDLOC(D%MYMS,0)` because oftentimes we need the actual position of M==0, but agree in setup_trans we can use the way you propose. I will change this. (IMLOC0 used to be a global variable KMLOC0 until very recently when I removed it to enable various resolutions).

samhatfield:
Sorry @lukasm91, #180 has introduced a merge conflict, but it looks quite simple to resolve.

lukasm91:
no worry - I merged it with develop.

src/trans/gpu/algor/hicblas_gemm.hip.cpp
```diff
@@ -334,12 +333,14 @@ void hipblas_sgemm_wrapper_grouped(int resol_id, int blas_id, char transa,
...
 
 void hipblas_dgemm_wrapper_grouped(int resol_id, int blas_id, char transa,
-                                   char transb, int m, int *n, int *k,
-                                   double alpha, const double *A, int lda,
-                                   int64_t *offsetsA, const double *B, int ldb,
-                                   int64_t *offsetsB, double beta, double *C,
-                                   int ldc, int64_t *offsetsC, int batchCount,
-                                   size_t stream, void *growing_allocator) {
+                                   char transb, int m, const int *n,
+                                   const int *k, double alpha, double const *A,
```
samhatfield, line 337: The SGEMM equivalent has `const float *A` but here we have `double const *A`. This are different right?

src/trans/gpu/algor/hicblas_cutlass.cuda.h
```diff
@@ -154,11 +154,11 @@ class cutlass_sgemm_grouped<CutlassType::cutlass_fp32, TransA, TransB> {
 
 } // namespace detail
 template <cublasOperation_t TransA, cublasOperation_t TransB>
-void cutlass_sgemm_wrapper_grouped_op(int resol_id, int blas_id, int m, int *n, int *k,
+void cutlass_sgemm_wrapper_grouped_op(int resol_id, int blas_id, int m, const int *n, const int *k,
                                       float alpha, const float *A, int lda,
-                                      int64_t *offsetsA, const float *B, int ldb,
-                                      int64_t *offsetsB, float beta, float *C,
-                                      int ldc, int64_t *offsetsC, int batchCount,
+                                      const int64_t *offsetsA, const float *B, const int *ldb,
```
samhatfield, line 159: More generally, what's the rationale for adding the `const` decorator to the integer pointer arguments?

src/trans/gpu/algor/hicblas_gemm.hip.cpp
lukasm91: no they are the same (const applies to whatever is on the left, except there is nothing to the left, then it is to the right). but it makes sense to be consistent, will check for inconsistencies.

src/trans/gpu/algor/hicblas_cutlass.cuda.h
lukasm91: to make sure we cannot write into the data that we are pointing to, and also just consistency. It is natural that offsetsA cannot be modified, so I added it to some inner function, but that certainly has to be propagated outwards up to the entry function. Note that scalars are somewhat different - one could also add const for those, but modifying them does not propagate, so it is less harmful. I hope that makes sense.

samhatfield: It does! Many thanks. I am still very much a C++ n00b.

samhatfield:
If I understand right, you have basically flattened the Legendre polynomial work arrays (`ZAA`, `ZAS`), which contain the polynomials for every zonal wavenumber, from 3D to 1D, borrowing the same offset indexing style of the other 1D work arrays. The 3D representation was wasteful because we allocated the same amount of memory for each zonal wavenumber despite that the size of the polynomial array decreases as you go up in wavenumber. Very nice work indeed.

I am happy to merge this pending the conversation above and once you give me the go ahead @lukasm91.

lukasm91:
I just modified the consts to be "consistent" now. Your description makes sense! I think from my side we are ready to merge.

PR #166, opened 2024-10-22
Fix define to execute acc init

lukasm91:
As discussed separately, `ACCGPU` is not defined because it is PRIVATE.

`_OPENACC` is the right define to use here because it is enabled whenever the code is compiled with OpenACC support.


wdeconinck (approved):
Thanks!

PR #171, opened 2024-11-13
Merged versions from CY49T2 and CY49R1 for CY50

AlexandreMary:
This series of commits results from changes commited in ectrans for CY49T2, merged with the version of ectrans from CY49R1 (in the contrib/ directory of ifs-source).

The final commit 00cdb6ea, tagged `for-CY50` in ACCORD-NWP fork of ectrans, is the version bundled for CY50 at first (a version merged with a more recent version of ectrans, could be later bundled).

Note: past CY50, contributions from ACCORD consortium or MF to ectrans will be PR'ed directly and without so much delay to the main branch of this repo.

PR #172, opened 2024-11-13
ACCORD-NWP CY50 contributions

samhatfield:
This PR takes the branch from PR #171 and rebases it against develop for a clean merge.

This is a draft as further changes are required, e.g. adding an import for `TW` into `FTINV` etc.


samhatfield:
@ddegrauwe - I amended your commits which had ddegrauw@login05.leonardo.local as the "email address". Now you only need to sign the CLA once :)

src/programs/CMakeLists.txt
```diff
@@ -51,6 +51,20 @@ foreach( program ectrans-benchmark )
...
 
+if ( HAVE_ETRANS )
+  foreach( prec sp dp )
+    if( HAVE_${prec} )
+      ecbuild_add_executable(TARGET  ectrans-lam-benchmark-${prec}
+                             SOURCES ectrans-lam-benchmark.F90
+                             LIBS
+                               fiat
+                               parkind_${prec}
+                               trans_${prec}
```
marsdeno, line 62: I think something like 
`OpenMP::OpenMP_Fortran`
is needed here, as omp_lib is used in the driver

src/ectrans4py/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+if(HAVE_ETRANS)
```
samhatfield, line 1: I think this check is redundant because it's done in the CMakeLists.txt one level up.

wdeconinck:
Copyright statements are missing in every file I quickly browsed through.

wdeconinck:
There are tests missing for etrans and ectrans4py.
This code is guaranteed to break at some point without any notice.
The benchmark code for etrans could be transformed into a test.

src/trans/cpu/internal/tpm_fftw.F90
```diff
@@ -45,6 +46,7 @@ MODULE TPM_FFTW
   TYPE(FFTW_PLAN),POINTER :: FFTW_PLANS(:) => NULL()
   INTEGER(KIND=JPIM) :: N_MAX=0         ! maximum number of latitudes
   INTEGER(KIND=JPIM) :: N_MAX_PLANS=4   ! maximum number of plans for each active latitudes
+  LOGICAL            :: LALL_FFTW=.FALSE. ! T=do kfields ffts in one batch, F=do kfields ffts one at a time
```
wdeconinck, line 49: This new logical variable `LALL_FFTW` is configurable via a new optional argument to `SETUP_TRANS`: `LD_ALL_FFTW`
This variable is then passed to `EXEC_FFTW(..., LD_ALL)`
The routine `EXEC_FFTW` itself however apparently does not make use of this variable, i.e. an unused argument.

So I propose to revert all changes related to this variable in `TPM_FFTW`, and `SETUP_TRANS` as it brings nothing.
Changing the API to external facing routines like `SETUP_TRANS` is something we should not do lightly.
It would involve also adding tests to exercise the new API, and one also needs to consider the GPU codepath, where FFTW is not used.

CMakeLists.txt
```diff
@@ -156,6 +156,20 @@ ecbuild_add_option( FEATURE GPU_STATIC
                     DEFAULT ${GPU_STATIC_DEFAULT}
                     DESCRIPTION "Compile GPU library as static library")
 
+ecbuild_add_option( FEATURE ETRANS
+                    DEFAULT OFF
+                    DESCRIPTION "Include Limited-Area-Model Transforms" )
+
+ecbuild_add_option( FEATURE PROGRAMS
+                    DEFAULT ON
+                    DESCRIPTION "Build src/programs" )
```
wdeconinck, line 165: Why do we need an option to disable the programs?

If there is a good reason to do so, we should be consistent with other packages (e.g. eccodes) which use the [feature "BUILD_TOOLS" ](https://github.com/ecmwf/eccodes/blob/develop/CMakeLists.txt#L107-L109 )for this.

src/trans/cpu/internal/tpm_fftw.F90
wdeconinck: Apologies, I see now that LD_ALL is used in EXEC_FFTW after all.

wdeconinck:
Configuration with -DENABLE_ETRANS=ON seems broken. Possibly something went wrong with the rebase?

```
CMake Error at src/etrans/CMakeLists.txt:22 (target_include_directories):
  target_include_directories may only set INTERFACE properties on INTERFACE
  targets
```

I noticed that this PR patches a posteriori targets like `trans_dp` and `trans_sp` using `target_sources`. These targets are recently INTERFACE targets that wrap `ectrans_dp` and `ectrans_sp` which have precision suffixes to all modules and subroutines. The INTERFACE libraries `trans_dp` and `trans_sp` are there for backwards compatibility where the interface #include files for these targets are auto-generated and just redefine the symbols with suffix to be the old symbol without suffix.

The approach should be that new etrans_dp and etrans_sp targets should be created that link with trans_dp and trans_sp rather than patch them.

wdeconinck:
Can the addition of ectrans4py be put on a backlog?

src/programs/CMakeLists.txt
AlexandreMary: > I think something like `OpenMP::OpenMP_Fortran` is needed here, as omp_lib is used in the driver

I wonder if that wouldn't be the reason why we had to disable the PROGRAMS via an option when compiling in a Docker image for the build of the ectrans4py Python wheel ?

CMakeLists.txt
AlexandreMary: See answer to comment by Olivier above

AlexandreMary:
> Configuration with -DENABLE_ETRANS=ON seems broken. Possibly something went wrong with the rebase?
> 
> ```
> CMake Error at src/etrans/CMakeLists.txt:22 (target_include_directories):
>   target_include_directories may only set INTERFACE properties on INTERFACE
>   targets
> ```
> 
> I noticed that this PR patches a posteriori targets like `trans_dp` and `trans_sp` using `target_sources`. These targets are recently INTERFACE targets that wrap `ectrans_dp` and `ectrans_sp` which have precision suffixes to all modules and subroutines. The INTERFACE libraries `trans_dp` and `trans_sp` are there for backwards compatibility where the interface #include files for these targets are auto-generated and just redefine the symbols with suffix to be the old symbol without suffix.
> 
> The approach should be that new etrans_dp and etrans_sp targets should be created that link with trans_dp and trans_sp rather than patch them.

I think we didn't want to create separate libraries for etrans, and at the same time that etrans be as little intrusive to trans as possible. If you have something better to propose... (I have very little cmake experience).

AlexandreMary:
> Can the addition of ectrans4py be put on a backlog?

What do you mean ?

wdeconinck:
- My proposal for etrans would be to have separate libraries, also making it clear that is is extending or adding functionality.
  @ddegrauwe if you want I can have a go at this, or discuss offline to assist you.

- What I meant with ectrans4py on backlog is that I'm not sure that this should be already added at this stage to ectrans.

  1) because it adds a lot of orthogonal things in one PR, so definitely I'd suggest a second PR just for this.
  2) because without having looked at it, I think ectrans is in such "flux" that it will be hard to maintain and test ectrans4py.

  If there's enough reason and urgency to add it at this point I can understand and support that.

ddegrauwe:
I'll put the lam transforms in a separate library, following the single/double precision mechanism from the global transforms. Is it really necessary to have a "common" part (as in common to sp/dp)? What's the problem with just having sp/dp variants for all files?

I'll also add copyright statements (ECMWF+MeteoFrance; I'm not sure if ACCORD is a legal entity that can hold copyrights).

I'll also implement a test for the LAM transforms.

wdeconinck:
> Is it really necessary to have a "common" part (as in common to sp/dp)? What's the problem with just having sp/dp variants for all files?

The common part is not only common to sp/dp (cpu), but also to gpu_sp and gpu_dp.
The ultimate goal we're having is to have multiple backends (cpu/gpu+sp/dp) that can be chosen at runtime rather than at compile time.
Therefore we have this common library with common data structures like the domain decomposition and anything else that is precision independent, which can be shared and accessed from all backends.
Algortihms in the common library, like the domain decomposition have been made double-precision only. They should not give different result in single precision.

AlexandreMary:
Regarding ectrans4py:
it is quite important to MF/ACCORD, as it is used by our python package `epygram` that enable us to read and transform (to gridpoint) spectral fields from Arpege & LAM model files.
`epygram` is used quite broadly at MF and in ACCORD.
The python binding to the spectral transforms `ectranspy` has been around (under a different name) in `ifsaux` for a while (developed mainly by @SebastienRietteMTO), and used to be compiled by gmkpack. It has been "transfered" recently by the work of an ACCORD colleague, @walidchikhi, into `ectrans4py` in ectrans, to make it a standard python package, compilable with cmake, and published on PyPI.

These are simple wrappings to ectrans functions, and the python binding is done using `ctypes` under the hood. That said, I understand that upcoming changes in the interface routines of ectrans may require adaptations in these, and therefore tests are necessary.

Another option could be to make it a standalone project, out of ectrans though relying on it, but I think it would be a pity.

wdeconinck:
OK @AlexandreMary fair enough, I won't object. I'd urge to have some tests for this soon. If nobody has time for this now that can come later but it would be great to have soon. That would also serve as use a minimal documentation.

AlexandreMary:
For the tests, cf. PR #199

PR #175, opened 2024-11-22
Add build-hpc workflow

reuterbal:
This adds a Github Actions workflow that dispatches to our HPC's GPU partition to enable testing for the GPU codepaths.

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -1,0 +1,109 @@
+name: build-hpc
+
+# Controls when the action will run
+on:
+
+  # Trigger the workflow on all pushes, except on tag creation
+  push:
+    branches:
+    - '**'
+    tags-ignore:
+    - '**'
+
+  # Trigger the workflow on all pull requests
+  pull_request: ~
+
+  # Allow workflow to be dispatched on demand
+  workflow_dispatch: ~
+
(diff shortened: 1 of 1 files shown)


.github/workflows/build-hpc.yml
```diff
@@ -0,0 +1,109 @@
+name: build-hpc
+
+# Controls when the action will run
+on:
+
+  # Trigger the workflow on all pushes, except on tag creation
+  push:
+    branches:
+    - '**'
```
mlange05, line 9: I think you might want to restrict general branch push runs to "main" and "develop". Pull requests (including draft PRs) and manual triggers (below) should suffice for a priori testing if that's needed.

reuterbal: Thanks, good point. I agree that PR and main/develop branches should suffice by default - on demand trigger is still available for other branches

samhatfield:
Nice work. Is the idea that AC-based testing will replace the existing system? Or just supplement it?

reuterbal:
This is meant as a supplement that is able to utilize GPUs and the software versions available on our HPC. The testing of the full compiler matrix should remain on standard runners, which are also available to anyone when working on (public) forks.

samhatfield (approved):

PR #176, opened 2024-11-25
Add GPU tests

samhatfield:
This PR tidies up the testing structure and adds tests for the GPU version.

I had to undo PR #158 as it causes a crash on our system: `invalid device buffer`.

I also had to disable the `--nfld 0` test as this fails on GPU.

This PR builds on #175.

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -60,7 +60,7 @@ jobs:
               - ninja
               - ecbuild
               - prgenv/nvidia
-              - hpcx-openmpi
+              - hpcx-openmpi/2.14.0-cuda
               - fftw
             cmake_options:
               - -DENABLE_MPI=ON
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,13 +166,16 @@ ectrans_find_lapack()
 ecbuild_add_option( FEATURE TESTS
                     DEFAULT ON
                     DESCRIPTION "Enable unit testing"
-                    REQUIRED_PACKAGES "MPI COMPONENTS Fortran"
-                    CONDITION HAVE_CPU )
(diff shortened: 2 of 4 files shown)


reuterbal (commented):
Looks great, many thanks!

samhatfield:
I'm not able to reproduce the two tests failing ac-gpu manually.

reuterbal:
I think I might know where the test failures come from: We're building with GPU-aware MPI but are running with the default hpcx-openmpi module that doesn't have GPU-aware MPI baked in. I suppose we may have to specify module versions after all.

In addition, you can also try changing the `sbatch` headers to `--mem=0`. That might have been what broke the pinning maybe, if it can't get enough pages.

samhatfield:
> I think I might know where the test failures come from: We're building with GPU-aware MPI but are running with the default hpcx-openmpi module that doesn't have GPU-aware MPI baked in. I suppose we may have to specify module versions after all.

That's certainly possible, but other tests with `NPROC > 1` succeeded, e.g. `ectrans-benchmark-gpu-sp_T47_O48_mpi2_omp1_nfld10_nlev20_nproma16`. Only the following two are failing:
```
ectrans-benchmark-gpu-dp_T47_O48_mpi2_omp1_nfld10
ectrans-benchmark-gpu-sp_T47_O48_mpi2_omp1_nfld10
```

> In addition, you can also try changing the `sbatch` headers to `--mem=0`. That might have been what broke the pinning maybe, if it can't get enough pages.

I'll give that a go!

samhatfield:
Given that the last commit (9699c64) gives the same error, I think the issue is with linking with `-cuda` rather than using pinned memory.

samhatfield:
Good to go I think. @wdeconinck any strong feelings regarding the CMake changes?

PR #177, opened 2024-11-26
Fixup ext_acc.F90

samhatfield:
Minor PR that builds on #175. I'm just testing out the automated testing infrastructure for GPU running


samhatfield:
Well that was a bit more complicated than I expected. I was able to reduce code duplication in ext_acc.F90 but I had to move it from gpu_common to the precision-specific GPU backend.

samhatfield:
Will close for now as this isn't urgent and wasn't as useful as I expected.

PR #182, opened 2024-12-09
[DRAFT] Test HPC CI

reuterbal:


samhatfield:
I think this is ready for merging now. For now we have limited the tests that are run on LUMI-G, basically excluding everything but the install and `SETUP_TRANS0` tests. We have outstanding issues with running on AMD GPUs which are outside of the remit of this PR. Once those are resolved, we can start testing properly on LUMI-G.

I think there are probably a few things which could be tidied up with this PR. If anyone can think of a neater way of passing toolchain-like info (`OpenMP_C_LIB_NAMES` stuff etc.) then please suggest alternatives.

In fact, I just realised that we can probably at least do the CPU tests on LUMI-G. Let me try that.

reuterbal (commented):
This is fantastic, many thanks for bringing this over the line @samhatfield !

I agree that it's useful to bring it in like this and do the fixing of the actual tests separately.

Also, I think specifying the toolchain-like options like this is fine. The fact that this is required in the first place is painful by itself and should be an exception, but the generic templating mechanism to pass around cmake options as part of the matrix is definitely useful!

Unfortunately, I can't approve the PR as I'm the original author 😆

samhatfield:
I relaxed the test filter so only "*gpu*" tests are filtered. The remaining all work. So now we are at least testing something meaningful!

By the way, the work directories for each CI run are still left behind on LUMI-G. That's fine but at some point we need to figure out how to clean them up automatically.

wdeconinck (approved):
This is all great! Hopefully we can get the GPU configuration working soon as well.

PR #183, opened 2024-12-09
Fixup _CRAYFTN-specific copy ins

samhatfield:
This PR fixes a Cray-specific bug that somehow slipped through my testing on MI250x.

From commit message:

> It looks like you need to put derived types and their members in the same DATA region together, with the Cray compiler. Without this I saw some strange numerical artefacts suggesting a race condition.

I really don't get this one. There's nothing wrong with what @lukasm91 wrote syntactically, but somehow the offloaded code has an apparent race condition. Without this fix, the error norms from the benchmark program are garbage.

I really hope we have more stable behaviour with OpenMP offload from the Cray compiler, once that's mature.


wdeconinck:
If the CRAYFTN code also works fine with NVIDIA, is it then worth it to keep both code paths with preprocessor guards?

lukasm91:
> If the CRAYFTN code also works fine with NVIDIA, is it then worth it to keep both code paths with preprocessor guards?

It is doable for sure, but we have to be careful and make sure that each COPYIN has a corresponding DELETE (in dealloc_resol_mod) - remember that COPYIN(D%some_scalar) increases a counter, so it also needs to have a corresponding DELETE for each scalar. In the scalar member case, what we do is redundant, but it will work. We would probably still want to add a comment *why* we add it, to make sure whoever is going to look at the code understands that this is a workaround rather than "good practice".

wdeconinck:
I'm fine either way :) Perhaps it's good to keep the good practice as default and do exception for CRAYFTN as is proposed.

samhatfield:
All good for merging then @wdeconinck @lukasm91?

wdeconinck (approved):

PR #188, opened 2024-12-15
Missing GSTATS in trgtol

PaulMullowney:
The highlighted lines were missing in the profile:

 157 FTINV_CTL      - L to G transposition         100        22.557        22.560      3.26      0.00
 **158 FTDIR_CTL      - G to L transposition         100         7.908         8.947      1.29      0.15**
 400 GSTATS         - GSTATS itself               3119         0.000         0.000      0.00      0.00


samhatfield:
Good spot. This brings the TRGTOL call in the GPU branch in line with the CPU branch, where `TRGTOL` in in [GSTATS region 158](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ftdir_ctl_mod.F90#L160). It's also consistent with the call to `TRLTOG` which is in [GSTATS region 157](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/internal/inv_trans_ctl_mod.F90#L236) in the GPU version.

samhatfield (approved):

PR #190, opened 2024-12-19
Fix precision of PGW argument in gpu/TRANS_INQ

samhatfield:
The `PGW` argument of `TRANS_INQ` should be `JPRB`. This is the case in the [CPU version](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/external/trans_inq.F90#L181) which matches the [interface](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/include/ectrans/trans_inq.h#L167).

However commit [71a00c1707978e9ae17e7593fdd6f97fc43d6e95](https://github.com/ecmwf-ifs/ectrans/commit/71a00c1707978e9ae17e7593fdd6f97fc43d6e95#diff-d9aa9280ad2e99900301651c43d83bdbaccfba44443249778c34c0ecfa14535eR179) changed it to `JPRD`, probably by accident. This caused a crash in the IFS when running with SP on GPUs because in at least one case the receiving variable for `PGW` was `JPRB`. For some reason the compilation passed despite the type mismatch.

In any case this argument should be `JPRB` also in GPU and that's what this PR does.


samhatfield:
Should we make a hotfix to main and release 1.5.1 too?

samhatfield:
Just testing this commit on top of 1.5.0 in the IFS. Assuming it works I'll instead open a hotfix PR against main.

PR #191, opened 2024-12-19
Add CMAKE_TEST_LAUNCHER capability and gpu labels

wdeconinck:
Before this PR, the serial tests that require a GPU resource would fail when using
```
salloc -q dg -n 4 --gpus-per-task=1 ctest
```
The issue is those tests need to be launched with `srun -n 1` within ctest to gain access to the GPU resource.

Adding the CMAKE_TEST_LAUNCHER (available since cmake 3.29) allows to run serial tests to be run with a launcher, in this case `srun -n 1`.
This feature used to be already available in versions of cmake < 3.29 under the more obscure name CMAKE_CROSSCOMPILING_EMULATOR. This PR makes the use of CMAKE_TEST_LAUNCHER forward compatible.

I also added labels "gpu" to GPU tests.

Following workflow works now on our HPC:

Before configuration:
```
export CMAKE_TEST_LAUNCHER="srun;-n;1"
```
To launch only gpu tests:
```
salloc -q dg -n 4 --gpus-per-task=1 ctest -L gpu
```

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -76,6 +76,9 @@ jobs:
               module load {{module}}
             {% endfor %}
 
+            export CMAKE_TEST_LAUNCHER="srun;-n;1"
+            export DR_HOOK_ASSERT_MPI_INITIALIZED=0
+
             {% for name, options in dependencies.items() %}
                 mkdir -p {{name}}
                 pushd {{name}}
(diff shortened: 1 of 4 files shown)


samhatfield (approved):
Nice work, very neat and tidy. I tested with the following CMakes on ECMWF/AC:
- cmake/3.28.3
- cmake/3.25.2

reuterbal (approved):

marsdeno (approved):
Looks good to me

PR #193, opened 2024-12-19
Improve error handling for GPU code path unimplemented features

wdeconinck:
All following changes are required to more gracefully integrate ectrans GPU code paths in [atlas](www.github.com/ecmwf/atlas) via transi.

- Workaround to avoid some unimplemented gpu-code path features:
    - SETUP_TRANS(LDSPSETUPONLY=.TRUE.)
    - SETUP_TRANS(LDGRIDONLY=.TRUE.)
    
    The workarounds consists in ignoring the "ONLY" setup and setting everything up with a low truncation or grid resolution. Not ideal but a sufficient temporary solution.
    This avoids special handling in downstream code like Atlas that would otherwise perform the same tricks.

- Calling of ABORT_TRANS for other unimplemented gpu-code path features:
    - VORDIV_TO_UV
    - SETUP_TRANS(LDLL=.TRUE.)
    
    Now we don't get obscure errors or segfaults in lower-level code paths
    
- Unsupported known gpu code paths called from within transi are now protected explicitly with error return codes rather than hard failures that cannot be recovered from. These error return codes should be removed as features are implemented of course:
      - trans_vordiv_to_UV
      - trans_dirtrans_adj
      - trans_invtrans_adj
      - trans_setup with llatlon=true
     
  This now means that atlas unit-tests can choose to ignore unimplemented features via exception handling


src/trans/gpu/external/setup_trans.F90
```diff
@@ -283,6 +283,24 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
+! <<<<<<<<<<< EXTRA TO WORKAROUND NOT YET IMPLEMENTED FEATURE
+  IF (LLSPSETUPONLY) THEN
+    WRITE(NOUT,'(A)') "DEVELOPER WARNING: LDSPSETUPONLY IS NOT YET IMPLEMENTED CORRECTLY WITH GPU BACKEND. IGNORING IT FOR NOW"
+    LLSPSETUPONLY = .FALSE.
+    R%NDGL = NPROC
+    ! Make even and positive
+    IF (MOD(R%NDGL,2) /= 0) THEN
+      R%NDGL = NPROC+1
+    ENDIF
+    R%NDGL = MAX(2,R%NDGL)
```
samhatfield, line 298: What's the purpose of this block?

src/trans/gpu/external/setup_trans.F90
```diff
@@ -363,10 +383,14 @@ SUBROUTINE SETUP_TRANS(KSMAX,KDGL,KDLON,KLOEN,LDSPLIT,PSTRET,&
...
 IF(PRESENT(LDGRIDONLY)) THEN
   D%LGRIDONLY=LDGRIDONLY
-ENDIF
-
-IF(PRESENT(LDSPSETUPONLY)) THEN
-  LLSPSETUPONLY=LDSPSETUPONLY
+! <<<<<<<<<<< EXTRA TO WORKAROUND NOT YET IMPLEMENTED FEATURE
+  IF (D%LGRIDONLY) THEN
+      R%NSMAX=1
+      R%NTMAX = R%NSMAX
```
samhatfield, line 389: Likewise, what are these two lines for?

samhatfield (commented):
Approved pending questions above.

lukasm91 (approved):
I am fine with these changes. For the future, in order to fix those (implement LDSETUPONLZ and GRIDONLY), it might help if we could define what they actually mean. At least I always struggled with what it actually means, which variables have to be set, and which don't.

src/trans/gpu/external/setup_trans.F90
wdeconinck: The purpose is to workaround an unidentified grid. For the LDSPSETUPONLY there is no grid. So we're just creating a super coarse grid here, and then perform the normal setup as if we have a grid.
In a first iteration when initially trying this I created a mock grid in Atlas, but that leaks the fact that a workaround is needed. It's better to add the workaround in ectrans and fix it within later.

src/trans/gpu/external/setup_trans.F90
wdeconinck: Also a workaround. The LGRIDONLY approach is used e.g. if you only want to do grid point norms or compute the domain decomposition.
As this does not work currently, I added a very coarse spectral truncation (1) and continue as usual.

samhatfield (approved):

PR #201, opened 2025-01-23
Add approved manual triggering of build-hpc action

samhatfield:


reuterbal (approved):
Very nice, many thanks for this!

Just double-checking: for the regular build.yml workflow we don't need the label trigger, right? (Since that doesn't use any secrets?)

PR #204, opened 2025-01-28
MétéoFrance contributions

samhatfield:
This PR contains the "miscellaneous" contributions from PR #199.


wdeconinck:
Looks OK! 
I am just questioning if the new variable should not be referring to FFTW at all, so replacing `LALL_FFTW` with `LALL_FFT`?

samhatfield:
Yes you're probably right. But then you should change also `LD_ALL_FFTW` to match, and that might break things at MF. @AlexandreMary do you think we can change the `LD_ALL_FFTW` argument of `SETUP_TRANS` to `LD_ALL_FFT`?

samhatfield:
Good point. I will add that argument also to the GPU version and make it print a warning or something.

samhatfield:
Done.

Regarding the changes to `GATH_SPEC`, I note that after this PR, `cpu/GATH_SPEC` and `gpu/GATH_SPEC` are identical (modulo some whitespace). That must be because the changes proposed by MF in this PR were already merged into the CPU version some time ago. When I created this PR from cherry picking from MF's PR, I think those changes were applied a second time, but this time to the GPU source tree, because when we split into "cpu/gpu", we `git mv`'d the existing source tree to `/gpu`.

Long story short, I don't think this PR should be modifying anything under `/gpu`. That's an error caused by the cherry pick, so I'm going to remove those changes.

samhatfield:
Hi @wdeconinck and @AlexandreMary - as explained above, the suggested changes to `GATH_SPEC` (and `GATH_SPEC_CONTROL`) already seem to be in the cpu version. Hence the changes required to sync with MF are actually even smaller than originally thought. The only thing we need to add now is `LD_ALL_FFTW`. Could you take a look again and then we can merge this?

marsdeno (approved):
Looks good for me.

PR #205, opened 2025-01-28
Python binding `ectrans4py`

AlexandreMary:
This branch, based on PRs https://github.com/ecmwf-ifs/ectrans/pull/203 and https://github.com/ecmwf-ifs/ectrans/pull/204, introduces the Python binding `ectrans4py`.

This is a Python interface to some `ectrans` functions, mainly the necessary routines to be able to:

* transform a spectral field into gridpoint
* transform a gridpoint field into spectral
* compute gridpoint derivatives of a field in spectral space

both for global and LAM fields.

It uses [`ctypesForFortran`](https://github.com/UMR-CNRM/ctypesForFortran) Python package for the Fortran<>Python interface.

Note: version number is duplicated in `src/ectrans4py/__init__.py` for now, until better solution.
Replaces PR https://github.com/ecmwf-ifs/ectrans/pull/199


samhatfield:
Yes exactly! :) Then you can just force push to your repo and the PR will update itself.

AlexandreMary:
OK I wait for 203 to be merged in develop then.

samhatfield:
Hi @AlexandreMary - you can try rebasing this branch against develop now.

AlexandreMary:
License of [`ctypesforFortran`](https://github.com/UMR-CNRM/ctypesForFortran) dependency has been switched to Apache-2.0

src/ectrans4py/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+if(HAVE_ETRANS)
```
samhatfield, line 1: Because `HAVE_ETRANS` is a condition of the `ECTRANS4PY` feature, I think this if statement is not necessary.

CMakeLists.txt
```diff
@@ -185,6 +185,13 @@ ecbuild_add_option( FEATURE ETRANS
                     DEFAULT OFF
                     DESCRIPTION "Include Limited-Area-Model Transforms" )
 
+	   
+ecbuild_add_option( FEATURE ECTRANS4PY
+                    DEFAULT OFF
+		    CONDITION HAVE_ETRANS
```
samhatfield, line 191: If I understand right, `HAVE_DOUBLE_PRECISION` is effectively a condition for `ECTRANS4PY`. Why not make it "official" and put it as a condition here?

src/ectrans4py/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
...
+ecbuild_list_add_pattern(
+    LIST ectrans4py_src
+    GLOB ${CMAKE_CURRENT_SOURCE_DIR}/*.F90
+    QUIET
+)
+
+if(HAVE_DOUBLE_PRECISION)
+    # Add sources
+    target_sources(trans_dp PRIVATE ${ectrans4py_src})
+endif()
```
samhatfield, line 11:
```suggestion
# Add sources
target_sources(trans_dp PRIVATE ${ectrans4py_src})
```

`HAVE_ECTRANS4PY` is a condition for this file being included, and that condition is only true if `HAVE_DOUBLE_PRECISION` is also true, hence you can remove this if safely.

samhatfield:
Just one more suggestion 👆🏻

wdeconinck:
Hello @AlexandreMary I have tried this and have some failures in the test.
My instructions on our atos to reproduce:

```bash
module load openmpi
module load intel-mkl          # contains fftw and blas/lapack
module load python3/new

mkdir ectrans4py
git clone git@github.com:ecmwf/ecbuild
git clone git@github.com:ecmwf-ifs/fiat
git clone git@github.com:AlexandreMary/ectrans -b ectrans4py_isolated_and_rebased

# build fiat
cmake -B fiat/build -S fiat
cmake --build  fiat/build
export fiat_ROOT=$PWD/fiat/build

# Build ectrans4py
python3 -m venv venv
. venv/bin/activate
pip install build
python -m build --wheel ectrans -o $PWD
pip install ectrans4py*

# Test ectrans4py
cd ectrans/tests/test_ectrans4py
pip install pytest
python -m pytest
```
The last command then fails with:
```
======================================================================================= test session starts =======================================================================================
platform linux -- Python 3.11.10, pytest-8.3.5, pluggy-1.5.0
rootdir: ***/ectrans
configfile: pyproject.toml
collected 0 items / 1 error

============================================================================================= ERRORS ==============================================================================================
____________________________________________________________________ ERROR collecting tests/test_ectrans4py/test_ectrans4py.py ____________________________________________________________________
test_ectrans4py.py:4: in <module>
    import ectrans4py
../../../venv/lib/python3.11/site-packages/ectrans4py/__init__.py:39: in <module>
    raise FileNotFoundError(msg)
E   FileNotFoundError: 'libtrans_dp.so' was not found in any of potential locations: ['/usr/local/apps/intel/19.0.5/compilers_and_libraries/linux/mkl/lib/intel64_lin', '/opt/mpi/openmpi/4.1.1.1/lib', '/opt/mpi/openmpi/4.1.1.1/fortran-gnu/lib', '***/venv/lib/python3.11/site-packages/ectrans4py/lib', '***/venv/lib/python3.11/site-packages/ectrans4py/lib64']. You can specify a different location using env var LD_LIBRARY_PATH
===================================================================================== short test summary info =====================================================================================
ERROR test_ectrans4py.py - FileNotFoundError: 'libtrans_dp.so' was not found in any of potential locations: ['/usr/local/apps/intel/19.0.5/compilers_and_libraries/linux/mkl/lib/intel64_lin', '/opt/mpi/openmpi/4.1.1.1/...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================================================================== 1 error in 0.62s =========================================================================================
```
The library `libtrans_dp.so` no longer exists. The new library `libectrans_dp.so` should be used, but that contains symbols with `_dp` suffix. 
I tried renaming in the `ectrans4py/__init__.py` the library `libtrans_dp.so` to `libectrans_dp.so` but the test is calling symbols without `_dp` suffix.

Also on macOS which we heavily use for development and testing,  the library suffix `.so` should be renamed to `.dylib`.

So this branch seems not to be compatible with the latest release at the moment.

AlexandreMary:
> The library `libtrans_dp.so` no longer exists. The new library `libectrans_dp.so` should be used

When did that happen ? And which symbols ?
Isn't that the kind of interface breaks that we were talking about on tuesday, that should be flagged more explicitly ?

src/ectrans4py/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# (using CMAKE_CURRENT_SOURCE_DIR is necessary because sources are in a different directory than the target library (trans_${prec}))
+ecbuild_list_add_pattern(
+    LIST ectrans4py_src
+    GLOB ${CMAKE_CURRENT_SOURCE_DIR}/*.F90
+    QUIET
+)
+
+if(HAVE_DOUBLE_PRECISION)
+    # Add sources
+    target_sources(trans_dp PRIVATE ${ectrans4py_src})
```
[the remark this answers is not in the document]
AlexandreMary: why ? Seems simpler like that to me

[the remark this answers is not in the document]
wdeconinck: Hi @AlexandreMary this is for the same reasoning that was applied to etrans being a separate library rather than patched into existing library: modularity and portability. First it is not very intuitive to patch a library afterwards in a different directory scope.
Then there are also the libraries ectrans_sp, ectrans_dp, ectrans_gpu_dp, ectrans_gpu_sp. These should have the same API bar from a suffix as `_sp`, `_dp`, `_gpu_sp`, `_gpu_dp`. The idea is that we will be able to link to all these libraries at the same time in the future, and make runtime decisions which backend to dispatch.

Having a new tiny bindings library linking only to `ectrans_dp` is also only needed when compiling ectrans4py and makes it clear that it is not available as core functionality, which would then also be required in all the other backends.

wdeconinck:
> > The library `libtrans_dp.so` no longer exists. The new library `libectrans_dp.so` should be used
> 
> When did that happen ? And which symbols ? Isn't that the kind of interface breaks that we were talking about on tuesday, that should be flagged more explicitly ?

I agree that going forward and as discussed last Tuesday this should be coordinated better for sure! Sincere apologies from all of us, and we are all learning. I am excited to deeper collaborate on changes from all sides before sudden facing of huge de-facto changes that are incompatible. For this PR I don't think this change is a show-stopper, rather another inconvenience, that would have needed to be communicated better! 

Anyway, this happened one year ago in https://github.com/ecmwf-ifs/ectrans/pull/61 and merged with commit https://github.com/ecmwf-ifs/ectrans/commit/18c8de0a91645e4cb6c049f64a00504fa6966796
and further after the GPU changes were merged as well with PR https://github.com/ecmwf-ifs/ectrans/pull/141. The latter also has a diagram of the interdependency of the cmake library targets.

Technically it was not an API change but an ABI change due the creation of the NEW `ectrans_dp` library containing the symbols with the `_dp` suffix and repurpose the cmake target `trans_dp` to be a header-only library with interfaces that preproc-redefine symbols without `_<backend>`. At least in ifs-source this was happening seamlessly without code change. I could understand that the MF build system might not have made this as seamless (or maybe it was if this went under the radar until now)?

The symbols renamed are only the precision-dependent ones; it was a big job to split this, and try to make as much code as possible precision independent (causing the JPRB->JPRD change of the earth radius in setup_trans0). This suffixes are added with a sedrename script creating new files in the build-directory, leaving the source files intact as if there were no renames. There is also a new precision-INDEPENDENT library `ectrans_common` which is even in common with both CPU/GPU backends containing these files: https://github.com/ecmwf-ifs/ectrans/tree/main/src/trans/common , mostly related to domain decomposition, meaning that setup_trans0 is now precision-independent, and can be tested in isolation from any used backend.

samhatfield:
Draft branch for resolving the above: https://github.com/samhatfield/ectrans/tree/ectrans4py_develop_sync

I'll open a PR against ectrans4py_isolated_and_rebased once it works. Currently all tests in test_ectrans4py fail with errors like:
```
AttributeError: /lus/h2resw01/hpcperm/nash/ectrans4py/venv/lib/python3.11/site-packages/ectrans4py/lib64/libectrans_dp.so: undefined symbol: trans_inq4py_
```

samhatfield:
Nevermind that error: I didn't reinstall the ectrans4py through pip correctly. Now it fails with `undefined symbol: fft992_`. [fft992 no longer exists in ecTrans](https://github.com/ecmwf-ifs/ectrans/pull/104). Yes I know... I thought this was cleared with all the relevant parties but somehow ectrans4py fell through the cracks. It's not an issue though, I'm sure we can adapt it for FFTW.

For now I will delete sp2gp_fft1d4py.F90 to get the other tests working. Now I see:
```
/lus/h2resw01/hpcperm/nash/ectrans4py/venv/lib/python3.11/site-packages/ectrans4py/lib64/libectrans4py_dp.so: undefined symbol: dir_trans_
```
which is probably another issue caused by the _sp/_dp confusion. Will look deeper.

samhatfield:
That issue is resolved. Now I'll have a look at replacing the fft992 call with an equivalent from FFTW.

AlexandreMary:
OK thanks.
That `sp2gp_fft1d4py` may seem weird, but it is basically doing transforms for "single column" model files (actually, a 2x2 homogeneous grid).
I'm not totally sure it is still in use, will check.

samhatfield:
Thanks! I'll wait for your signal then. If you think it can be deleted, I'll delete it, and then open a PR against your branch. Then we can implement a Github action to run the ectrans4py test suite.

AlexandreMary:
> BTW the code referenced `ELLIPS64` which doesn't exist - I'm guessing this is just a version of `ELLIPS` which accepts 64-bit integers. I just casted these integers to 32 bit and used `ELLIPS`. I also had to cast integers to 32-bit when calling `EXEC_FFTW`.

Yes, scories of the past...
Thanks for your implication.

AlexandreMary:
> I tried to convert it to FFTW [here](https://github.com/samhatfield/ectrans/commit/6911df4be862486cfec2468b3b7feb4af0704f4b) but there's no test for `sp2gp_fft1d4py` so I'm not sure 100% if I got it right. Anyway, the other 6 tests pass with this change, following @wdeconinck's instructions above.

Indeed, I didn't include a test for this because I don't have a usecase... which is why I doubted it was still used...
Well I guess if it is and your porting was not right, we'll know when someone will complain !

samhatfield:
Okeydokey - I have opened a PR to update your branch here: https://github.com/AlexandreMary/ectrans/pull/1.

samhatfield:
Note: branch samhatfield:ectrans4py_develop_sync above works on macOS following @wdeconinck's instructions, but only with following diff:

```diff
diff --git a/tests/test_ectrans4py/test_ectrans4py.py b/tests/test_ectrans4py/test_ectrans4py.py
index f96c3f7d..72b1af97 100644
--- a/tests/test_ectrans4py/test_ectrans4py.py
+++ b/tests/test_ectrans4py/test_ectrans4py.py
@@ -3,7 +3,7 @@ import numpy
 from . import data
 import ectrans4py

-ectrans4py.init_env()
+ectrans4py.init_env(unlimited_stack=False)

 KNUMMAXRESOL = 10
 EPSILON = 1e-13
```

samhatfield:
I have added a workflow for testing ectrans4py [here](https://github.com/AlexandreMary/ectrans/pull/2), following @wdeconinck's instructions.

samhatfield:
> This is great @samhatfield ! Now one final check I'd add to the build-ectrans4py CI is to make sure the version of ectrans4py encoded in **init**.py matches the one in the `VERSION` text file. This I feel is something we will overlook often in the future, and we should probably figure out a better way.

Good idea! I noticed that my version is wrong for some reason:
```python
>>> import ectrans4py
>>> ectrans4py.__version__
'1.2.51'
```

samhatfield:
Nevermind, for some reason I had a cached wheel with the wrong version. Not sure how that happened.

samhatfield:
Done: https://github.com/AlexandreMary/ectrans/pull/3/files

AlexandreMary:
> This is great @samhatfield ! Now one final check I'd add to the build-ectrans4py CI is to make sure the version of ectrans4py encoded in **init**.py matches the one in the `VERSION` text file. This I feel is something we will overlook often in the future, and we should probably figure out a better way.

I agree. Maybe you could ask eccodes people if/how they do that ? I guess they must have the same need/issue ?
And thanks @samhatfield

wdeconinck:
> I agree. Maybe you could ask eccodes people if/how they do that ? I guess they must have the same need/issue ?

I just checked and there, the version is retrieved by a library call:
https://github.com/ecmwf/eccodes-python/blob/develop/gribapi/gribapi.py#L2336

If we are linking with `transi_dp` (the C-interface) we could also provide access to such functions:
https://github.com/ecmwf-ifs/ectrans/blob/main/src/transi/version.h

samhatfield:
I've just opened another PR which removes the duplicated version by having ectrans4py call the `ECTRANS_VERSION_STR`:

https://github.com/AlexandreMary/ectrans/pull/4/files

I did have to delete the version from pyproject.toml though, as the build didn't work otherwise. I think the build step tries to get the version from ectrans4py before the Fortran is compiled, which obviously won't work. I did some reading online and it seems like putting the version in pyproject.toml is not mandatory? But I'm not a Python expert.

wdeconinck (approved):
As far as I'm concerned this is now completely ready we can be good as can be future-proof with some testing and CI.

Thanks all for pushing this over the line!

samhatfield:
I'm not concerned about the fail as the failing tests are the GPU ones, and those naturally won't work on LUMI-G with `ectrans4py_isolated_and_rebased` because that was rebased before we merged changes required to run on AMD GPUs. However I'm puzzled why those tests are running in the first place because [they're supposed to be excluded](https://github.com/ecmwf-ifs/ectrans/actions/runs/13946846724/workflow?pr=205#L82). It's almost like the `develop` workflow file is being used rather than `ectrans4py_isolated_and_rebased`. Given that we have the same issue with https://github.com/ecmwf-ifs/ectrans/pull/226, I'm just going to merge this PR and we can figure this out later in https://github.com/ecmwf-ifs/ectrans/pull/226.

PR #210, opened 2025-02-06
Remove _OVERLOAD layer

samhatfield:
This PR removes one layer of routines in the Legendre transforms and permits us to call the cuBLAS/hipBLAS routines directly from LEINV/LEDIR via their Fortran bindings. The `*_OVERLOAD` layer didn't do anything useful and was slowing down debugging of the OpenMP offload development.

I won't mark this as a draft so that the CI suite is triggered, but please wait for my confirmation before merging.


lukasm91 (approved):
Looks good to me - nice simplification :)

wdeconinck (approved):
good to merge

PR #213, opened 2025-02-13
Workaround for HIPFFT_PARSE_ERROR when building with ACCGPU on Cray platform

thomasgibson:
This PR introduces two commits: one fixes an issue when building with OpenACC --- ECTrans is freeing the derived type `FG` twice, causing crashing during the tear-down phase.

The next commit is a workaround for the nefarious 'HIPFFT_PARSE_ERROR' observed on LUMI-G and other Cray systems with AMD GPUs. This should hopefully allow you to enable GPU testing with AMD GPUs while this HIPFFT_PARSE_ERROR can be further investigated.


samhatfield (approved):

samhatfield:
We're having some issues with our CI runners. Once these are resolved and the build-hpc CI jobs pass, I will merge this.

PR #214, opened 2025-02-18
Touchup adjoint test

samhatfield:
The existing adjoint test program seems to check the following identity:

```math
\langle \text{DIR\_TRANS}\left(\delta \text{INV\_TRANS}\left(X \right)\right), Y \rangle = \langle X, \text{INV\_TRANSAD}\left( \delta \text{DIR\_TRANSAD}\left( Y \right)\right) \rangle
```

Where $\delta$ is an arbitrary perturbation. $X$ and $Y$ are both spectral space variables filled with random numbers between -1 and +1.
The program has a number of issues:
- The test has no assertion of the results, hence it doesn't do anything useful. It just prints the values without checking.
- It lacks documentation so it isn't clear what it's doing. I figured out the above by studying the code.
- The perturbation $\delta$ does nothing. The perturbation is only applied to the first `NFLEVG` fields of `PGP`. But because the `INV_TRANS` call that fills `PGP` includes vorticity and divergence (which are both zeroed in this test), the first `NFLEVG` fields correspond to these zero-valued vorticity values, so the multiplicative perturbation does nothing. This was probably an oversight of whoever wrote the code, but in any case I don't think a perturbation is needed, i.e. you can set $\delta=1$ in the equation above and it's still a proper test.
- The program includes GSTATS calls, but we don't care about performance here, only correctness, so they aren't needed. A separate program should be created for benchmarking the adjoint code.
- The test tests the inverse and direct transform together. This means that if the test fails, we don't know which one is causing it. Perhaps this doesn't matter so much.

This PR aims to resolve this and make the adjoint test actually useful.

I'll keep this as a draft for now while I think of any other ways the test could be improved.


samhatfield:
Ready for review.

PR #219, opened 2025-02-24
Optimizations for OpenMP path

PaulMullowney:
Fixed a few things which substantially improve performance. For Single device MI300A runs, 79, 159, 319 truncations yield as good  (if not better) performance using OpenMP compared to OpenACC.


src/trans/gpu/algor/hicblas_gemm.hip.cpp
```diff
@@ -196,12 +196,8 @@ template <typename Real> struct hipblas_gemm_grouped {
                   const Real *A, int lda, const Real *B, int ldb, Real beta,
                   Real *C, int ldc) const {
     // TODO: sort out this nonsense
```
samhatfield, line 198:
```suggestion
```

samhatfield (approved):

PR #222, opened 2025-03-03
Fix argument checks

### LINKED ISSUE #220: Fix interface checks
### What happened?

In inv_trans and dir_trans, at least for the GPU, the interface checks are inconsistent.
https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/external/inv_trans.F90

e.g. Line 464
```
  IF(IUBOUND(1) < NPROMA) THEN
    WRITE(NOUT,*)'INV_TRANS:FIRST DIM. OF PGP TOO SMALL/LARGE ',IUBOUND(1),NPROMA
    CALL ABORT_TRANS('INV_TRANS:FIRST DIMENSION OF PGP TOO SMALL/LARGE ')
  ENDIF
```
The error message (correctly) states that the first dimension should be NPROMA, and nothing else. The comparison should be `/=`, not `<`. I think we should try not to make the interface more generic than needed.

Or line 604:
```
    IF(IUBOUND(1) < NPROMA) THEN
      WRITE(NOUT,*)'INV_TRANS:FIRST DIM. OF PGP3B TOO SMALL/LARGE ',IUBOUND(1),NPROMA
      CALL ABORT_TRANS('INV_TRANS:FIRST DIMENSION OF PGP3B TOO SMALL/LARGE ')
    ENDIF
  ELSE
```
This is actually not even the intent. What should be here is `IUBOUND(4) /= NGPBLKS`.

I think we should make the interface checks more strict (as much as possible, as little as needed). 

Does not apply

### Version

develop

-

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

samhatfield:
Fixes #220.


src/trans/cpu/external/inv_trans.F90
```diff
@@ -576,22 +576,22 @@ SUBROUTINE INV_TRANS(PSPVOR,PSPDIV,PSPSCALAR,PSPSC3A,PSPSC3B,PSPSC2,&
...
-      CALL ABORT_TRANS('INV_TRANS:THIRD DIMENSION OF PGP3B INCONSISTENT ')
+      CALL ABORT_TRANS('INV_TRANS:THIRD DIMENSION OF PGP3B /= IF_SC3B_G3 ')
     ENDIF
-    IF(IUBOUND(4) < NGPBLKS) THEN
-      WRITE(NOUT,*)'INV_TRANS:FOURTH DIM. OF PGP3B TOO SMALL ',IUBOUND(4),NGPBLKS
-      CALL ABORT_TRANS('INV_TRANS:FOURTH DIMENSION OF PGP3B TOO SMALL ')
+    IF(IUBOUND(1) /= NPROMA) THEN
+      WRITE(NOUT,*)'INV_TRANS:FIRST DIMENSION OF PGP3B /= NPROMA',IUBOUND(1),NPROMA
+      CALL ABORT_TRANS('INV_TRANS:FIRST DIMENSION OF PGP3B /= NPROMA')
     ENDIF
```
lukasm91, line 595:
```suggestion
    IF(IUBOUND(4) /= NPROMA) THEN
      WRITE(NOUT,*)'INV_TRANS:FOURTH DIMENSION OF PGP3B /= NGPBLKS',IUBOUND(1),NPROMA
      CALL ABORT_TRANS('INV_TRANS:FOURTH DIMENSION OF PGP3B /= NGPBLKS')
    ENDIF
```

src/trans/gpu/external/inv_trans.F90
```diff
@@ -588,22 +588,22 @@ SUBROUTINE INV_TRANS(PSPVOR,PSPDIV,PSPSCALAR,PSPSC3A,PSPSC3B,PSPSC2,&
...
-      CALL ABORT_TRANS('INV_TRANS:THIRD DIMENSION OF PGP3B INCONSISTENT ')
+      CALL ABORT_TRANS('INV_TRANS:THIRD DIMENSION OF PGP3B /= IF_SC3B_G3 ')
     ENDIF
-    IF(IUBOUND(1) < NPROMA) THEN
-      WRITE(NOUT,*)'INV_TRANS:FIRST DIM. OF PGP3B TOO SMALL/LARGE ',IUBOUND(1),NPROMA
-      CALL ABORT_TRANS('INV_TRANS:FIRST DIMENSION OF PGP3B TOO SMALL/LARGE ')
+    IF(IUBOUND(1) /= NPROMA) THEN
+      WRITE(NOUT,*)'INV_TRANS:FIRST DIMENSION OF PGP3B /= NPROMA',IUBOUND(1),NPROMA
+      CALL ABORT_TRANS('INV_TRANS:FIRST DIMENSION OF PGP3B /= NPROMA')
     ENDIF
```
lukasm91, line 607:
```suggestion
    IF(IUBOUND(4) /= NGPBLKS) THEN
      WRITE(NOUT,*)'INV_TRANS:FOURTH DIMENSION OF PGP3B /= NGPBLKS',IUBOUND(1),NPROMA
      CALL ABORT_TRANS('INV_TRANS:FOURTH DIMENSION OF PGP3B /= NGPBLKS')
    ENDIF
```

lukasm91 (changes requested):
Only went through it very quickly for now.

src/trans/cpu/external/inv_trans.F90
samhatfield: Well spotted.

src/trans/cpu/external/gath_grid.F90
```diff
@@ -103,25 +103,25 @@ SUBROUTINE GATH_GRID(PGPG,KPROMA,KFGATHG,KTO,KRESOL,PGP)
...
 IF(IFRECV > 0) THEN
   IF(.NOT.PRESENT(PGPG)) THEN
     CALL ABORT_TRANS('GATH_GRID:PGPG MISSING')
   ENDIF
-  IF(UBOUND(PGPG,1) < D%NGPTOTG) THEN
-    CALL ABORT_TRANS('GATH_GRID:FIRST DIMENSION OF PGPG TOO SMALL')
+  IF(UBOUND(PGPG,1) /= D%NGPTOTG) THEN
+    CALL ABORT_TRANS('GATH_GRID:FIRST DIMENSION OF PGPG /= D%NGPTOTG')
   ENDIF
  IF(UBOUND(PGPG,2) < IFRECV) THEN
```
lukasm91, line 126: I guess this is `<` on purpose because we don't know how many fields are being gathered? If that's the case, I am fine fine with this (since this is the last dimension), but we might want to give a better error message for this scenario, i.e. tell the user please increase the size of PGPG to XXX (now YYYY), or at lest write this into NOUT.

src/trans/gpu/external/gath_grid.F90
```diff
@@ -102,25 +102,25 @@ SUBROUTINE GATH_GRID(PGPG,KPROMA,KFGATHG,KTO,KRESOL,PGP)
...
 IF(IFRECV > 0) THEN
   IF(.NOT.PRESENT(PGPG)) THEN
     CALL ABORT_TRANS('GATH_GRID:PGPG MISSING')
   ENDIF
-  IF(UBOUND(PGPG,1) < D%NGPTOTG) THEN
-    CALL ABORT_TRANS('GATH_GRID:FIRST DIMENSION OF PGPG TOO SMALL')
+  IF(UBOUND(PGPG,1) /= D%NGPTOTG) THEN
+    CALL ABORT_TRANS('GATH_GRID:FIRST DIMENSION OF PGPG /= D%NGPTOTG')
   ENDIF
  IF(UBOUND(PGPG,2) < IFRECV) THEN
```
lukasm91, line 125: same as above

lukasm91 (approved):
minor comments, but if those are addressed, I approve.

src/trans/cpu/external/dir_trans.F90
```diff
@@ -218,7 +218,7 @@ SUBROUTINE DIR_TRANS(PSPVOR,PSPDIV,PSPSCALAR,PSPSC3A,PSPSC3B,PSPSC2,&
 
 IF(PRESENT(KVSETSC2)) THEN
   IF(.NOT.PRESENT(PSPSC2)) THEN
-    CALL ABORT_TRANS('DIR_TRANS:KVSETSC2 BUT NOT PSPSC2')
+    CALL ABORT_TRANS('DIR_TRANS:KVSETSC2 PRESENT BUT NOT PSPSC2')
```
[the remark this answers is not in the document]
samhatfield: Yup, good suggestion.

src/trans/cpu/external/gath_grid.F90
samhatfield: But here we're talking about `NGPTOTG` -> the global number of grid points, not the number of fields that we're gathering. I agree that the previous check was probably innocuous, but is there any case where someone would intentionally pass a `PGPG` array that has more elements than `NGPTOTG` in the first dimension?

lukasm91: the comment is on line 126 :) IFRECV, essentially I was asking if line 126, you want to keep the < or if we want to change it to /=. With NGPTOTG I am absolutely with you, noone should ever pass an array with not NGPTOTG, that will just not be consistent, and the library user can handle this.

samhatfield: Oh I see. I think it should be `/=`. As with the other cases, I can't think of a reason why you'd pass an array with an oversized field dimension.

wdeconinck (approved):

PR #223, opened 2025-03-04
Remove redundant SP only versions of DIST_GRID and GATH_GRID

### LINKED ISSUE #126: The fate of _32 routines

samhatfield:
Resolves #126.

Please wait for my signal before merging this -> this depends on a PR to ifs-source to remove calls to _32 subroutines which is under review.


samhatfield:
I'd like to get this PR merged. As far as I'm aware, the only place these routines were used was in the IFS prototype (and defunct for almost a decade) EnKF. I recently filed a pull request to remove these cases and now there is no instances of `GATH_GRID_32` nor `DIST_GRID_32` in ifs-source. CY49R1 still has them but CY50R1 will not. The former (and previous) cycles don't use an external ecTrans repository anyway, right?

@RyadElKhatibMF can you confirm that MétéoFrance does not make any use of `GATH_GRID_32` nor `DIST_GRID_32`?

@wdeconinck would you favour completely deleting these routines like in this PR, or leaving in the toplevel subroutines with an abort statement?

RyadElKhatibMF:
I confirm I can't see any use of GATH_GRID_32 nor DIST_GRID_32

samhatfield:
Okay, I have deprecated those routines by replacing all contents with an abort statement.

There are no instances of these routines in ifs-source from CY49R1 onwards. Earlier cycles still contain calls to these subroutines, but only in the EnKF which is an experimental data assimilation system not under further development for almost a decade. So, once CY48R1 (the last one to call these routines) has reached end of life (is that a thing??) we can simply delete these routines entirely.

samhatfield:
It would be good to get this PR out of the way @wdeconinck 🙏

wdeconinck (approved):
Nice @samhatfield, good to go.

PR #224, opened 2025-03-04
Update Github runners to Ubuntu 24.04

samhatfield:


samhatfield:
Note that the `ectrans_test_adjoint` test is broken for intel-llvm (ifx) due to a compiler bug. The following program behaves wrong when compiling with ifx and `-fiopenmp`:
```fortran
PROGRAM REPRODUCER

IMPLICIT NONE

INTEGER, PARAMETER :: N = 5
INTEGER :: ARRAY(N)

ARRAY(:) = 1

PRINT *, "Values in PROGRAM"
PRINT *, ARRAY(1:N)
CALL TEST1

CALL TEST2(ARRAY)

CONTAINS

  SUBROUTINE TEST1
    PRINT *, "Values in TEST1"
    PRINT *, ARRAY(1:N)
  END SUBROUTINE TEST1

  SUBROUTINE TEST2(KVSET)
    INTEGER, INTENT(IN) :: KVSET(:)

    PRINT *, "Values in TEST2"
    PRINT *, KVSET
  END SUBROUTINE TEST2

END PROGRAM REPRODUCER
```

I've reported this to Jacob Poulsen at Intel.

samhatfield:
Good to merge.

.github/tools/install-mpi.sh
```diff
@@ -83,11 +83,6 @@ case "$os" in
           echo "Not taking any action."
           exit 0
         fi
-        if [ -n "${I_MPI_ROOT}" ]; then
-          echo "MPI is already installed at I_MPI_ROOT=${I_MPI_ROOT}."
-          echo "Not taking any action."
-          exit 0
-        fi
```
wdeconinck, line 90: You should not need to remove this.

.github/workflows/build.yml
```diff
@@ -166,13 +158,18 @@ jobs:
...
 
     - name: Install FFTW
       shell: bash -eux {0}
       run: |
         ${ECTRANS_TOOLS}/install-fftw.sh --version 3.3.10 --with-single --prefix ${DEPS_DIR}/fftw
         echo "FFTW_ROOT=${DEPS_DIR}/fftw" >> $GITHUB_ENV
 
+    - name: Install OpenBLAS
+      shell: bash -eux {0}
+      run: ${ECTRANS_TOOLS}/install-openblas.sh
```
wdeconinck, line 171: Which platform requires this? I thought all runners already have this.

.github/tools/install-mpi.sh
samhatfield: Okay, I can't remember why I removed this. I'll put it back and see what happens.

wdeconinck: OK makes sense then. But then why did you add line 161 too, if we are to ignore I_MPI_ROOT?

wdeconinck:
I would add the ifx compiler bug in a separate PR for visibility

.github/workflows/build.yml
```diff
@@ -31,58 +31,50 @@ jobs:
...
-            compiler_cxx: icpc
-            compiler_fc: ifort
+          - name : linux intel-llvm
+            os: ubuntu-24.04
+            compiler: intel-llvm
+            compiler_cc: icx
+            compiler_cxx: icpx
+            compiler_fc: ifx
+             # TODO: this tests is broken with intel-llvm (segfault)
+            ctest_options: -E ectrans_test_install
```
wdeconinck, line 77: ectrans_test_install fails??? That is very concerning...

samhatfield:
See https://github.com/ecmwf-ifs/ectrans/pull/227/files for ifx bug fix.

wdeconinck:
> See https://github.com/ecmwf-ifs/ectrans/pull/227/files for ifx bug fix.

OK that is merged now. This can be rebased.

.github/workflows/build.yml
samhatfield: Yeah, I'll reenable it so you can see the error.

samhatfield: See https://github.com/ecmwf-ifs/ectrans/actions/runs/13815938473/job/38649010125?pr=224#step:12:4336. Tricky to debug non-interactively.

samhatfield: I had a go on Atos with the same version of Intel (2023.2.0) but the test passed. I think we'll need to debug this through the action itself.

wdeconinck:
Intel-classic needs ubuntu-22

.github/workflows/build.yml
```diff
@@ -78,11 +69,18 @@ jobs:
             caching: false
 
           - name : linux intel-classic
-            os: ubuntu-20.04
+            os: ubuntu-24.04
```
samhatfield, line 72:
```suggestion
            os: ubuntu-22.04
```

samhatfield:
> Intel-classic needs ubuntu-22

Trying that now.

wdeconinck:
> intel-classic suite fixed, good suggestion @wdeconinck.

I also encountered this for atlas a few weeks ago, when upgrading runners.

samhatfield:
I tried upgrading the intel-llvm oneAPI version to 2025.0 to see if that resolves the failing install test, but no. Also the functionality I implemented to allow installing different oneAPI versions somehow broke the intel-classic case. No idea why.

wdeconinck:
Yes @samhatfield let's not add intel-llvm right now, and add it after 1.6.0

wdeconinck (approved):

Issue #232, opened 2025-03-14
Format errors in ectrans-benchmark.F90


DJDavies2:
### What happened?

ectrans-benchmark.F90 is failing for some compilers due to problems with the format.

### What are the steps to reproduce the bug?

Building with e.g. NAG shows the problems

### Version

develop

### Platform (OS and architecture)

Linux

### Relevant log output

```shell

```

wdeconinck:
Fixed with #233 , Thanks @DJDavies2 !

PR #233, opened 2025-03-14
Fix some formats in ectrans-benchmark.F90

DJDavies2:
There are some commas missing in i/o formats in ectrans-benchmark.F90, I have fixed them here.  Addresses #232.


wdeconinck (approved):
Non-controversial, thanks @DJDavies2 !

PR #234, opened 2025-03-14
AMD Flang support

PaulMullowney:
This branch is a work-in-progress.

The change to buffered_allocator is due to how it is used throughout the code. Typically it is used as
ASSIGN_PTR(DST, GET_ALLOCATION(ALLOCATOR, RESERVATION_X), ...)

Since, GET_ALLOCATION is a function, it's intent must be passed with INTENT(IN), not INTENT(INOUT). Moreover, one can not modify the values at SRC so I eliminated that code.

If GET_ALLOCATION, were modified to be a subroutine, then the code in ASSIGN_PTR that tries to set values could be kept. However, that would require a lot of source code changes in the compute modules, which I don't think is desired.


lukasm91:
I know this is WIP, but just a small side note: Good point about SRC, that it has to be INTENT(IN)`. I think you can just make SRC INTENT(IN) in ASSIGN_PTR without breaking anything. The SET_VALUE logic is only setting the pointers target, not the pointer itself. 

The clean solution would be to just modify DST and not SRC and set it to `IEEE_VALUE(DST(1), IEEE_QUIET_NAN)`, rather than set individual bytes, that would just make the intent of the code clearer.

PaulMullowney:
> I know this is WIP, but just a small side note: Good point about SRC, that it has to be INTENT(IN)`. I think you can just make SRC INTENT(IN) in ASSIGN_PTR without breaking anything. The SET_VALUE logic is only setting the pointers target, not the pointer itself.
> 
> The clean solution would be to just modify DST and not SRC and set it to `IEEE_VALUE(DST(1), IEEE_QUIET_NAN)`, rather than set individual bytes, that would just make the intent of the code clearer.

Good suggestion! I thought I had tried this but apparently not. The changes to buffered_allocator_mod.F90 are much simpler now.

PaulMullowney:
> @PaulMullowney want me to approve this for CI so we can test it on ECMWF/ac and LUMI-G?

yes

samhatfield:
Everything looks good now. Is this ready for merging @PaulMullowney? Also maybe a silly question, but do you think we can add AMD Flang to the list of compilers in the CI suite? Can it be apt-getted/installed, or is it not open?

wdeconinck (approved):

samhatfield:
So you don't think it'll be possible to add AMD Flang to the list of compilers we test on Github? It would be useful to get early signs for any breaking changes.

PaulMullowney:
Actually, drop 6.0.0 just landed. With a few more changes to ECTrans, it "should" build out of the box. I'll submit the PR next week and we can try adding it to the CI

PR #235, opened 2025-03-14
Synchronise with pafv_CY49R1.0_fixers.IFS-3572

samhatfield:
This PR contains changes required to synchronise with Filip Vana's ifs-source branch pafv_CY49R1.0_fixers.IFS-3572.


wdeconinck:
The GPU code-path will have missing symbols.

wdeconinck:
Could this new development also be accompanied with a unit-test?

samhatfield:
> Could this new development also be accompanied with a unit-test?

It could, but we don't have unit tests for other external routines (apart from `SETUP_TRANS0`) either. Any reason why this one needs one?

wdeconinck:
> > The GPU code-path will have missing symbols.
> 
> Why is that?

Never mind, I see the GPU place-holders. I missed them last review. :)

samhatfield:
macOS tests are failing due to:
```
  <TL(X1),Y2>  =          0.962041244086947E-05
  <X1,AD(Y2)> =          0.962041244087702E-05
  Relative error = .784660498947183E-12
   ABORT_TRANS CALLED
   TEST_GPNORM_TRANS_ADJOINT: test failed
  *******************************
  TEST_GPNORM_TRANS_ADJOINT: test failed
  Relative error greater than 2000 * machine epsilon
   0.78E-12 >  0.44E-12
  *******************************
   ABORT!    1 TEST_GPNORM_TRANS_ADJOINT: test failed
```
So we will need to increase the tolerance again, possibly to 4000 * machine epsilon.

cxzk:
@wdeconinck and/or @marsdeno - please could you take a look at this with a view to approving and merging into a release - it's needed in order to progress with the IFS-3572 merge in develop.

src/trans/cpu/external/gpnorm_transad.F90
```diff
@@ -0,0 +1,91 @@
+! (C) Copyright 2008- ECMWF.
```
marsdeno, line 1: Do we care about accuracy of copyrights here?
If we do, I guess this should be 2024?

src/trans/cpu/external/gpnorm_transtl.F90
```diff
@@ -0,0 +1,91 @@
+! (C) Copyright 2008- ECMWF.
```
marsdeno, line 1: Same question as TRANSAD

src/trans/include/ectrans/gpnorm_transtl.h
```diff
@@ -0,0 +1,63 @@
...
+!                   (these do not have to be just levels)
+!     KPROMA      - required blocking factor (input)
+!     PAVE        - average (output)
+!     KRESOL      -  resolution tag (optional)
+!                    default assumes first defined resolution
+!
+
+!     Author.
+!     -------
+!        George Mozdzynski *ECMWF*
```
marsdeno, line 42: More details ... should author and date not be consistent with that in the routine itself?

marsdeno (commented):
Just a few "cosmetic" questions!

src/trans/include/ectrans/gpnorm_transtl.h
samhatfield: Have updated gpnorm_transtl.h to match gpnorm_transtl.F90.

marsdeno (approved):
Looks good to me, thanks!

wdeconinck (approved):
Thanks looks good!

PR #237, opened 2025-03-18
Fix poa1 bug

### LINKED ISSUE #236: Uninitialised values when grid point space and spectral space resolutions don't match
### What happened?

The output array of the direct Legendre transform, `POA1`, is undefined for certain wavenumbers when certain grid point/spectral resolution combinations are used. Specifically, when the grid point resolution is lower than the optimal (for which the number of latitudes is the 2 * (spectral truncation + 1)), `POA1` is not given any values.

Undefined stack arrays manifest in different ways for different compilers and for different numbers of threads. @cxzk identified a case as part of IFS testing in which, when N32 grid point field is transformed to T159 spectral field, the output from `DIR_TRANS` contains garbage values. This only seemed to occur for `OMP_NUM_THREADS > 1`. This was with the ifort compiler, version 2023.2.0.

I wasn't able to reproduce @cxzk's case in a simple test program, as in my case the undefined arrays usually had zero values. Still, this is undefined behaviour and should be rectified.

For single-threaded tests, I noticed that `POA1` seems to keep the values from the previous wavenumber in the loop, probably because the corresponding local stack array of `LTDIR`, `ZOA1`, has the `SAVE` attribute implicitly? For e.g. two-threaded cases, I think the second thread is responsible entirely for zonal wavenumbers which are uninitialised, hence `ZOA1` never takes sensible values, and therefore garbage values are liable to occur.

The correct solution is to zero `POA1` for wavenumbers for which no meaningful computation can be done. In `LEDIR` there should be an `ELSE` [here](https://github.com/ecmwf-ifs/ectrans/blob/main/src/trans/cpu/internal/ledir_mod.F90#L254), like so:
```fortran
ELSE
  POA1(:,:) = 0.0_JPRB
ENDIF
```
This is scientifically meaningful: when the grid point resolution cannot represented high frequency modes, it's correct to say that there is no (i.e. zero) energy in those modes.

Here's a test program which transforms from an O32 grid to a T159 spectral representation:

```fortran
PROGRAM TEST_PROGRAM

USE PARKIND1, ONLY: JPIM, JPRB

IMPLICIT NONE

! Spectral truncation
INTEGER(JPIM), PARAMETER :: TRUNC = 159

INTEGER(JPIM), PARAMETER :: NDGL = 64
INTEGER(JPIM) :: VERBOSITY = 0

! Arrays for storing our field in spectral space and grid point space
REAL(KIND=JPRB), ALLOCATABLE :: SPECTRAL_FIELD(:,:)
REAL(KIND=JPRB), ALLOCATABLE :: GRID_POINT_FIELD(:,:,:)

! Dimensions of our arrays in spectral space and grid point space
INTEGER(KIND=JPIM) :: NSPEC2
INTEGER(KIND=JPIM) :: NGPTOT

INTEGER(KIND=JPIM) :: I, NLOEN(NDGL)

#include "setup_trans0.h"
#include "setup_trans.h"
#include "trans_inq.h"
#include "dir_trans.h"
#include "trans_end.h"

CALL DR_HOOK_INIT()

! Initialise ecTrans (resolution-agnostic aspects)
CALL SETUP_TRANS0(LDMPOFF=.TRUE., KPRINTLEV=VERBOSITY)

! Define octahedral grid
DO I = 1, NDGL / 2
    NLOEN(I) = 20 + 4 * (I - 1)
    NLOEN(NDGL - I + 1) = NLOEN(I)
END DO

! Initialise ecTrans (resolution-specific aspects)
CALL SETUP_TRANS(KSMAX=TRUNC, KDGL=NDGL)

! Inquire about the dimensions in spectral space and grid point space
CALL TRANS_INQ(KSPEC2=NSPEC2, KGPTOT=NGPTOT)

! Allocate our work arrays
ALLOCATE(SPECTRAL_FIELD(1,NSPEC2))
ALLOCATE(GRID_POINT_FIELD(NGPTOT,1,1))

GRID_POINT_FIELD(:,:,:) = 1.0_JPRB

! Perform a direct transform
CALL DIR_TRANS(PGP=GRID_POINT_FIELD, PSPSCALAR=SPECTRAL_FIELD)

CALL TRANS_END

END PROGRAM TEST_PROGRAM
```

Add this program to the `foreach(  program ... )` loop in src/programs/CMakeLists.txt.

Place a print statement inside the `IF (KIFC > 0 .AND. KDGLU > 0 ) THEN` of `LEDIR`:
```fortran
WRITE(NOUT,*) "KM", KM, "COMPUTED"
```
Wavenumbers 0 to 63 will print. Wavenumbers 64 to 159 will not. For these wavenumbers **`POA1` is undefined**.

Follow the code: `POA1` corresponds to array `POA` of `UPDSPB`. There certain values (all undefined) are copied into `PSPEC` - the output spectral array.

### Version

1.5.1

intel/2023.2.0

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

ECMWF

samhatfield:
Fixes #236.

Commit `(tmp) Add test program for bugfix` just used to test new behaviour with different compilers through the test suite. It should be deleted before merging this.

### DIFF
--- a/src/trans/cpu/internal/ledir_mod.F90
+++ b/src/trans/cpu/internal/ledir_mod.F90
@@ -252,7 +252,7 @@ IF (KIFC > 0 .AND. KDGLU > 0 ) THEN
   ENDDO
 ELSE
   ! This zonal wavenumber KM has no computation to be done (G%NDGLU(KM) = 0)
-  ! This is usuallyy because the wavenumber cannot be represented on the given grid, so we should
+  ! This is usually because the wavenumber cannot be represented on the given grid, so we should
   ! zero POA1
   POA1(:,:) = 0.0_JPRB
 ENDIF
--- a/src/trans/cpu/internal/ledir_mod.F90
+++ b/src/trans/cpu/internal/ledir_mod.F90
@@ -250,7 +250,11 @@ IF (KIFC > 0 .AND. KDGLU > 0 ) THEN
       POA1(IS+(J-1)*2,JK) = ZCS(J,IFLD)
     ENDDO
   ENDDO
-  
+ELSE
+  ! This zonal wavenumber KM has no computation to be done (G%NDGLU(KM) = 0)
+  ! This is usuallyy because the wavenumber cannot be represented on the given grid, so we should
+  ! zero POA1
+  POA1(:,:) = 0.0_JPRB
 ENDIF
 
 !     ------------------------------------------------------------------


src/trans/cpu/internal/ledir_mod.F90
```diff
@@ -250,7 +250,10 @@ SUBROUTINE LEDIR(KM,KMLOC,KFC,KIFC,KSL,KDGLU,KLED2,PAIA,PSIA,POA1,PW)
       POA1(IS+(J-1)*2,JK) = ZCS(J,IFLD)
     ENDDO
   ENDDO
-  
+ELSE
+  ! This zonal wavenumber KM has no computation to be done, probably because the wavenumber cannot
```
marsdeno, line 254: It is specifically because G%NDGLU(KM) = 0 that we are not entering the IF(), but guess this isn't really saying much more than the inline comment you've left.

marsdeno (approved):

cxzk (approved):
This makes sense to me, and I can confirm that it fixes the original problem I encountered in an IFS EDA experiment, as well as my smaller test case.

src/trans/cpu/internal/ledir_mod.F90
```diff
@@ -250,7 +250,11 @@ SUBROUTINE LEDIR(KM,KMLOC,KFC,KIFC,KSL,KDGLU,KLED2,PAIA,PSIA,POA1,PW)
       POA1(IS+(J-1)*2,JK) = ZCS(J,IFLD)
     ENDDO
   ENDDO
-  
+ELSE
+  ! This zonal wavenumber KM has no computation to be done (G%NDGLU(KM) = 0)
+  ! This is usuallyy because the wavenumber cannot be represented on the given grid, so we should
```
lukasm91, line 255:
```suggestion
  ! This is usually because the wavenumber cannot be represented on the given grid, so we should
```

wdeconinck:
Is such fix also necessary for the GPU code path?

src/trans/cpu/internal/ledir_mod.F90
samhatfield: Nice spot :D

samhatfield:
> Is such fix also necessary for the GPU code path?

It's a good question. The GPU version is structured very differently here (`POA1` is used for all zonal wavenumbers at once), so I can't say right away. It would take me some time to figure this out. I'm sure @lukasm91 has thought about this?

lukasm91:
> > Is such fix also necessary for the GPU code path?
> 
> It's a good question. The GPU version is structured very differently here (`POA1` is used for all zonal wavenumbers at once), so I can't say right away. It would take me some time to figure this out. I'm sure @lukasm91 has thought about this?

Good question. Actually I am not sure what is happening on the GPU for this case. I will try it out next week.

lukasm91:
Actually, I just tested it: For CPU, I can easily reproduce the error by adding
ZOA1 = IEEE_VALUE(ZOA1(1,1),IEEE_QUIET_NAN)
in `ltdir_mod.F90` before calling LEDIR, and then modify the print in the test program to also print the sum of spectral field, gives
```
   -1.249681        1.000000        0.000000                 NaN
```

Adding the fix gives
```
   -1.249681        1.000000        0.000000       0.8514420
```

For the GPU version, the corresponding (and more rigorous) test is to go to buffered_allocator_mod.F90 and do `SET_VALUE_EFF=.TRUE.`. This option is purely for verification, but setting it will initialize all temporary storage with NaN. With that I get
```
   -1.249681        1.000000        0.000000       0.8514420
```
So indeed, it looks good for the GPU.

wdeconinck (approved):

wdeconinck:
Actually could we change the name of `test_program` to something meaningful?
Unit tests should preferably be named as `ectrans_test_<what_is_tested>`

samhatfield:
> Actually could we change the name of `test_program` to something meaningful? Unit tests should preferably be named as `ectrans_test_<what_is_tested>` Unless the test program should be deleted :)

The test program was just temporary and I’m gonna drop that commit before merging this.

PR #242, opened 2025-03-21
[DON'T MERGE] Test ECMWF contributions to feature/invtrans-adjoint-ecmwf

samhatfield:

PR #243, opened 2025-03-24
Fix bug in intent and initialisation of spectral arrays

samhatfield:
Let's track the `PSP` arrays through the inverse transform adjoint:
- [`INV_TRANSAD`: `INTENT(OUT) :: PSP...`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/external/inv_transad.F90#L145)
    - [`INV_TRANS_CTLAD`: `INTENT(OUT) :: PSP...`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/inv_trans_ctlad_mod.F90#L109)
        - [`LTINV_CTLAD`: `INTENT(OUT) :: PSP...`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ltinv_ctlad_mod.F90#L67)
          - [`LTINV_AD`: `INTENT(INOUT) :: PSP...`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ltinvad_mod.F90#L95)

The intent changes from `OUT` to `INOUT`!!!

On top of that the [packing logic in `PRFI1BAD` is `+=`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/prfi1bad_mod.F90#L91) even though the spectral arrays are **uninitialised**. This was triggering compiler-specific failing of the `INV_TRANS/AD` test in [this PR](https://github.com/ecmwf-ifs/ectrans/pull/242).

This PR resolves the inconsistency in intents and prevents uninitialised array bugs from appearing in `PRFI1BAD` by changing the packing logic to just `=`.


src/trans/cpu/internal/prfi1bad_mod.F90
```diff
@@ -88,8 +88,8 @@ SUBROUTINE PRFI1BAD(KM,PIA,PSPEC,KFIELDS,KFLDPTR)
     IFLD = KFLDPTR(JFLD)
     DO J=1,ILCM
       INM = IOFF+(ILCM-J)*2
-      PSPEC(IFLD,INM  ) = PSPEC(IFLD,INM  ) + PIA(J+2,IR)
```
marsdeno, line 91: I would want to be very sure that we're not relying on the `+=` anywhere before changing the code like this. If we are relying on values already in `PSPEC` in any call to `PRFI1BAD`, the right fix for this issue might be to zero the 5 arrays (spvor, spdiv, spscalar, spsc2, spsc3b) at the beginning of `LTINV_CTLAD` ?

src/trans/cpu/internal/ltinvad_mod.F90
```diff
@@ -92,12 +92,12 @@ SUBROUTINE LTINVAD(KM,KMLOC,KF_OUT_LT,KF_UV,KF_SCALARS,KF_SCDERS,KLEI2,KDIM1,&
 INTEGER(KIND=JPIM), INTENT(IN) :: KLEI2
 INTEGER(KIND=JPIM), INTENT(IN) :: KDIM1
 
-REAL(KIND=JPRB)  ,OPTIONAL, INTENT(INOUT) :: PSPVOR(:,:)
-REAL(KIND=JPRB)  ,OPTIONAL, INTENT(INOUT) :: PSPDIV(:,:)
-REAL(KIND=JPRB)  ,OPTIONAL, INTENT(INOUT) :: PSPSCALAR(:,:)
-REAL(KIND=JPRB)   ,OPTIONAL,INTENT(INOUT) :: PSPSC2(:,:)
-REAL(KIND=JPRB)   ,OPTIONAL,INTENT(INOUT) :: PSPSC3A(:,:,:)
-REAL(KIND=JPRB)   ,OPTIONAL,INTENT(INOUT) :: PSPSC3B(:,:,:)
+REAL(KIND=JPRB)   ,OPTIONAL,INTENT(OUT) :: PSPVOR(:,:)
```
marsdeno, line 95: LTINVAD is being called inside a loop in ltinv_ctlad, so intent(out) on these arguments is not conceptually correct.

samhatfield:
There is no need for this PR in any case as I now initialise the spectral arguments with zeros in the test that triggered this PR in the first place: https://github.com/ecmwf-ifs/ectrans/pull/250/commits/88bca4018edf6f9d79551a15631b2bb52f0c25c7. So I suggest we close this PR and instead consider changing the intents to `INOUT` in another PR.

marsdeno:
Can confirm based on IFS experiment that += semantics in PRFI1BAD are actively used.

PR #245, opened 2025-03-27
Fix typo in PRESENT checks

### LINKED ISSUE #244: leinv_strides optional parameter check/set mismatch
### What happened?

In the subroutine `leinv_strides` the `PRESENT` checks on the optional parameters `IOUT_STRIDES0` and `IOUT0_STRIDES0` don't match the values being set. See:

https://github.com/ecmwf-ifs/ectrans/blob/163ae8e398fc2e8bd3b5a2e22f21c13741afdf43/src/trans/gpu/internal/leinv_mod.F90#L39-L40

and

https://github.com/ecmwf-ifs/ectrans/blob/163ae8e398fc2e8bd3b5a2e22f21c13741afdf43/src/trans/gpu/internal/leinv_mod.F90#L47-L48

N/A

### Version

develop

N/A

### Relevant log output

```shell
N/A
```

### Accompanying data

N/A

### Organisation

JCSDA

samhatfield:
Fixes #244.

@lukasm91 - I'm guessing this was just a typo but can you sanity check it?


lukasm91:
Good catch! But the fix should be the other way. Not the assignment is wrong, but the present check.

IOUT0_STRIDES0 = ALIGN(KF_LEG,A)

and

IOUT_STRIDES0 = ALIGN(2*KF_LEG,A)

(because KM=0 is real)

samhatfield:
> Good catch! But the fix should be the other way. Not the assignment is wrong, but the present check.
> 
> IOUT0_STRIDES0 = ALIGN(KF_LEG,A)
> 
> and
> 
> IOUT_STRIDES0 = ALIGN(2*KF_LEG,A)
> 
> (because KM=0 is real)

Oh yeah! That probably explains the failed tests.

marsdeno (approved):
Nice catch!

PR #246, opened 2025-03-31
Resize PIA 2nd dimension to R%NSMAX+4

samhatfield:
In the CPU version, `ZIA`, which stores the input to the Legendre transform for all fields  (local to [`LTINV`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ltinv_mod.F90)) has dimensions `(R%NLEI1,KLEI2) = (R%NSMAX+4,NFIELDS)`. The first dimension seems odd, but basically we have `R%NSMAX+1` values going from `0` to `R%NSMAX`. We then add one more row, which I'm not sure the reason for yet. Then finally we add two more rows because `ZIA` will be used to also store gradients which are computed using a central differencing +1/-1 approach, so we need two more at the beginning and end so that every "normal" point has a gradient calculated. So `R%NSMAX+4` in total.

In the GPU version we are currently allocating the equivalent array `PIA` of [`LTINV`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/internal/ltinv_mod.F90) with dimensions `(2*IF_READIN,R%NTMAX+3,D%NUMP)` (unlike the CPU version which is defined inside a loop over zonal wavenumber, `PIA` stores data for all zonal wavenumbers, hence the third dimension).

`R%NTMAX = R%NSMAX` (which is another issue...) so `PIA` only has `R%NSMAX+3` elements in the second dimension. However, it still needs another one like in the CPU version so that the gradient is calculated correctly. In fact the loop for calculating north-south derivatives of scalar fields accesses element `R%NSMAX+4` so there is an out-of-bounds array access currently. See [the loop in `SPNSDE`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/internal/spnsde_mod.F90#L116) (look at the `KM=0`, `JN=0` iteration and note that `PF` is a pointer to a slice of `ZIA` over the first dimension).

According to @lukasm91, this out-of-bounds probably wasn't dangerous as in all cases we can see, the out-of-bounds access is multipled by zero (e.g. `ZEPSNM(1,0)` and `F%RLAPIN`. Still, it's good to have the dimensions consistent with the loop limits.

This PR enlarges `PIA` so it matches `ZIA`.

Note that `R%NTMAX = R%NSMAX` for all intents and purposes so I replaced some `NTMAX` with `NSMAX` to be consistent with the CPU version.

Thanks @l90lpa for starting this discussion in https://github.com/ecmwf-ifs/ectrans/pull/242.


lukasm91 (approved):
Yes I think this makes sense.  We can later think about if there is a way to make the indexing in PIA more natural (or at least document it)

src/trans/gpu/internal/prfi1b_mod.F90
```diff
@@ -109,7 +109,7 @@ SUBROUTINE PRFI1B(PIA,PSPEC,KFIELDS,KDIM,KFLDPTR)
 #endif
 #endif
     DO KMLOC=1,D_NUMP
-      DO JN=0,R_NSMAX+3
+      DO JN=0,R_NSMAX+4
```
l90lpa, line 112: Hi @samhatfield, I might be misunderstanding your change, so apologies if so. Changing the second dimension of `PIA` to be `R%NSMAX+4` makes sense to me. However, I think this subroutine in its original form was already writing to `R%NSMAX+4` and so didn't need changing? In the original code `JN` is in the interval `[0, R_NSMAX+3]`, but whenever `JN` is used for indexing it is used like `JN+1` which lies in the intervals `[1, R_NSMAX+4]`. Where as with the new loop bound and if-condition `JN+1` will lie in the intervals `[1, R_NSMAX+5]`?

lukasm91: I think you are right, and yes I also did it this way in my attempt to revert pia (which I believe is more natural than what we do here, because indexing is very complex): https://github.com/ecmwf-ifs/ectrans/compare/main...lukasm91:ectrans:fix_pia

dhaumont:
> According to @lukasm91, this out-of-bounds probably wasn't dangerous as in all cases we can see, the out-of-bounds access is multipled by zero (e.g. ZEPSNM(1,0) and F%RLAPIN. Still, it's good to have the dimensions consistent with the loop limits.

I'm not as optimistic as @lukasm91, as in PRFI1B, we write 0.0_JPRB to the out-of-bound location https://github.com/ecmwf-ifs/ectrans/blob/29cb2bf14d06d02ca4aabb16eaebebdd54527009/src/trans/gpu/internal/prfi1b_mod.F90#L125

This can be reproduced with ectrans-bencmark and the arguments `-f 1 -l 1`.
I believe this should be fixed asap.

By the way, do you know why this isn't catched in debug mode, when bound-checking is activated ?
UPDATE: bound checking is deactivated for GPU
```
nvfortran-Warning-CUDA Fortran or OpenACC GPU targets disables -Mbounds
```

src/trans/gpu/internal/prfi1b_mod.F90
samhatfield: Yes, looks like you're correct.

dhaumont: If you run with `s -f 1 -l 1`, here is what I observed on the GPU and on the CPU:

- GPU

   - SHAPE(PIA):            2           82           80
   - LBOUND(PIA):            1            1            1
   - UBOUND(PIA):            2           82           80
   - D_NUMP:           80
   - R_NSMAX:           79
   - The loops in PRFI1B try to access PIA in the range [1:83]: the access 83 is out-of range

- CPU  
    - SHAPE(PIA):          83           2
    - LBOUND(PIA)           1           1
    - UBOUND(PIA)          83           2
     - The loops in PRFI1B only access PIA in the range [1:82]. On top of that, PIA is dimensioned one element larger than it is on the GPU 

So if we want to keep consistent, we need to fix:
- the size of PIA
-  reduce the range of the loop in PRFI1B (and not increase it as was suggested initially)

samhatfield: > The loops in PRFI1B only access PIA in the range [1:82].

The loops only access elements [3:82]. Below that loop elements 1, 2, and 83 are zeroed separately. So I don't think `PRFI1B` GPU needs modifying at all.

dhaumont: > > The loops in PRFI1B only access PIA in the range [1:82].
> 
> The loops only access elements [3:82]. Below that loop elements 1, 2, and 83 are zeroed separately. So I don't think `PRFI1B` GPU needs modifying at all.

I agree with you. But why do we zeroed the 83th element, which is not done on the GPU ?

samhatfield: The loop in GPU/PRFI1B runs from JN = 0, R_NSMAX + 3 = e.g. [0, 82] for T79. In the loop we zero elements 1, 2 and 83, right?

dhaumont: sorry, my mistake: please read instead:

I agree with you. But why do we zero the 83th element, which is not done on the CPU ?

samhatfield: That's what I meant above: CPU/PRFI1B contains this logic:
```
DO JFLD=1,2*KFIELDS
  PIA(1,JFLD) = 0.0_JPRB
  PIA(2,JFLD) = 0.0_JPRB
  PIA(ILCM+3:,JFLD) = 0.0_JPRB
ENDDO
```
`ILCM=80` for T79, so ` PIA(ILCM+3:,JFLD)` is the 83rd element.

dhaumont: I think I got it finally, you are right. My logging code was incorrect on the CPU. Thanks for your patience :-)

samhatfield: Your challenges are welcome... I had to stop and think about it about 5 times...

wdeconinck (approved):
From the description:
> In the CPU version, ZIA, which stores the input to the Legendre transform for all fields (local to [LTINV](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ltinv_mod.F90)) has dimensions (R%NLEI1,KLEI2) = (R%NSMAX+4,NFIELDS). The first dimension seems odd, but basically we have R%NSMAX+1 values going from 0 to R%NSMAX. We then add one more row, which I'm not sure the reason for yet. Then finally we add two more rows because ZIA will be used to also store gradients which are computed using a central differencing +1/-1 approach, so we need two more at the beginning and end so that every "normal" point has a gradient calculated. So R%NSMAX+4 in total.

It would be good if this were somewhere written in the code in comments, or in some kind of ectrans documentation.
When reading the code again in a year (or less) it will otherwise be a mystery then.

Approving since that is just cosmetic.

dhaumont:
> Approving since that is just cosmetic.

What do you mean ?

wdeconinck:
> > Approving since that is just cosmetic.
> 
> 
> 
> What do you mean ?
> 
> 

My request to add comment is only cosmetic. I trust Sam will not break compilation 😁

PR #248, opened 2025-04-02
Add amdflang to CI

samhatfield:


samhatfield:
Status:
- MPI has been disabled for amdflang. Somehow the OpenMPI built by amdflang can't be found by CMake. We will work on this in a future PR.
- Otherwise all tests are passing with amdflang except for `ectrans_test_install ` which segfaults when running `transi_sptogp`. This is very similar to what we saw with IntelLLVM actually, which is very suspicious. However I cannot yet debug this crash interactively as I can't build a working amdflang on ECMWF's systems (I could probably build amdflang on my Mac but I have a feeling it's OS-specific). Will liaise with @PaulMullowney about how to debug this further.

samhatfield:
Note #249 must be merged into this in order for the amdflang tests to pass.

wdeconinck (approved):
This is great! Thanks also for the install-amd-flang script. Very useful.

marsdeno:
Excellent!

PaulMullowney:
We should hook the GPU build in, next. It will runtime fail, but at least we'll be covering compilation.

samhatfield:
> Note #249 must be merged into this in order for the amdflang tests to pass.

Correction: #249 must be merged into this in order for the amdflang tests to _run_. In this branch the `TESTS` feature is disabled for amdflang because we are building without MPI. After #249 is merged to develop (and I rebase this branch) we will be able to enable `TESTS` without MPI and we can then check whether the test suite passes for amdflang.

wdeconinck:
@samhatfield I rebased on develop.
Question. Why do we need to add the -fPIC and -fopenmp flags directly?
Does CMake not detect these flags automatically?
There exist alternatives like `-DOpenMP_Fortran_FLAGS=-fopenmp` to give CMake hints about which OpenMP Flag to use.

samhatfield:
> @samhatfield I rebased on develop. Question. Why do we need to add the -fPIC and -fopenmp flags directly? Does CMake not detect these flags automatically? There exist alternatives like `-DOpenMP_Fortran_FLAGS=-fopenmp` to give CMake hints about which OpenMP Flag to use.

Good question. It might be that I added those flags before I added the requirement for a CMake version that is amdflang-aware. I'll try removing them now and see what happens.

samhatfield:
- As for the OpenMP flag issue, I have tried using instead `OpenMP_{lang}_FLAGS=-fopenmp` but this isn't working. I would like to use a cleaner CMake-idiomatic solution but it's difficult to debug this as I can't reproduce it locally.

samhatfield:
> > * As for the OpenMP flag issue, I have tried using instead `OpenMP_{lang}_FLAGS=-fopenmp` but this isn't working. I would like to use a cleaner CMake-idiomatic solution but it's difficult to debug this as I can't reproduce it locally.
> 
> Can you come up with a list of issues and we'll discuss at Friday's meeting?

I can. It's just not really amdflang-new's fault necessarily, just CMake.

samhatfield:
Any ideas @PaulMullowney ?

PaulMullowney:
> Any ideas @PaulMullowney ?
Are these CPU and/or GPU builds?

PaulMullowney:
> CPU builds, on the GH virtual machines

For a local install, libomp.so is in something like:
/PATH/TO/rocm-afar-8248-drop-7.0.5/lib/llvm/lib/libomp.so

You could try adding -L/PATH/TO/rocm-afar-8248-drop-7.0.5/lib/llvm/lib/

I know it's not a solution but we can at least try it as a workaround.

samhatfield:
> > CPU builds, on the GH virtual machines
> 
> For a local install, libomp.so is in something like: /PATH/TO/rocm-afar-8248-drop-7.0.5/lib/llvm/lib/libomp.so
> 
> You could try adding -L/PATH/TO/rocm-afar-8248-drop-7.0.5/lib/llvm/lib/
> 
> I know it's not a solution but we can at least try it as a workaround.

I *think* the issue is that lib/llvm/lib is not on the `LD_LIBRARY_PATH` at the point when `ectrans-benchmark` is invoked. What you're suggesting would be the solution to a link-time failure, right? Whereas what we have is a runtime failure.

Still you gave me an idea. I tried adding an export of `LD_LIBRARY_PATH` in the install-amdflang step, but then it occurred to me that env vars are not by default shared between different steps of an action. So we would have to add that export to the `Build & Test` stage, but only for amdflang. Not sure how to do that. I'm also not sure why other compilers don't have this issue.

samhatfield:
amdflang is giving a segfault for the direct and inverse transform adjoint tests. I've spent a day trying to figure out what's going wrong, but sadly none the wiser. @PaulMullowney thinks it might be a stacksize issue, but as far as I can tell there's no way to modify the ulimits in our reusable action. I've tried simplifying the dirtrans test by removing the `SCALPRODGP` function and just directly calculating the values in the main program, reusing already allocated arrays as much as possible. But somehow, the segfault still creeps back in. I'll pick this up again later.

samhatfield:
After offline discussion with @reuterbal and @PaulMullowney, we strongly suspect that amdflang is more stack hungry than other compilers, and that the segfaults are due to running out of stack space. Unfortunately we don't believe it's possible to increase the stack size given the restricted structure of the reusable action we're using to build and test ecTrans.

Assuming this theory is correct, we have three options:
- Disable the failing tests for amdflang.
- Reduce the resolution of the failing tests to reduce the memory consumption. This is what I've done in this PR. I've also simplified the failing tests to reduce the size of allocated memory (mostly reusing work arrays).
- Try and write a wrapper script, akin to [test-install.sh.in](https://github.com/ecmwf-ifs/ectrans/blob/develop/tests/test-install.sh.in) which calls `ulimit -s unlimited`.

wdeconinck:
@samhatfield I tried installing latest amd-flang drop 22.3.0 and I noticed the flag `-fno-stack-arrays` exists.
It could be worth trying?
Also your install script does not work with just this latest drop 22.3.0 ; that should be adapted.

samhatfield:
I'll try upgrading to amdflang 22.2.0 first, then adding -fno-stack-arrays.

samhatfield:
Now compiling with -fno-stack-arrays. Still doesn't work :(

msleigh (approved):
Approving from copyright point of view only

msleigh (approved):

PR #249, opened 2025-04-02
MPI-free testing

samhatfield:
If possible, MPI should not be a requirement for the TESTS feature. At the moment we can't build ecTrans with amdflang and MPI (at least not with OpenMPI). For now we will disable MPI in the CI when building with amdflang. We still want to run the remaining serial test suite when building with amdflang though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -174,10 +174,6 @@ ecbuild_add_option( FEATURE ECTRANS4PY
 
 ectrans_find_lapack()
 
-ecbuild_add_option( FEATURE TESTS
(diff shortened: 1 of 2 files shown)


wdeconinck:
Indeed, all the logic in the tests is there to not rely on MPI, and I remember having tested this in the past and it was working then.
Just removing the check is indeed enough.
Further, the entire TESTS option can be removed as it is provided by default with ecbuild. I have rebased updated the PR with the option removed entirely.

wdeconinck (approved):

PR #255, opened 2025-04-16
Fix failing ectrans4py tests

samhatfield:
[#222](https://github.com/ecmwf-ifs/ectrans/pull/222)  broke the ectrans4py tests so this makes them work again.

I also noticed that the ectrans4py correctness checker doesn't actually do anything useful. It just compares one of the arrays to itself so obviously always passes. After this PR we will actually be checking correctness.


tests/test_ectrans4py/test_ectrans4py.py
```diff
@@ -13,7 +13,7 @@
     raise NotImplementedError("ectrans4py does not support Windows")
 
 KNUMMAXRESOL = 10
-EPSILON = 1e-13
+EPSILON = 1e-10
```
marsdeno, line 16: This is needed once we actually compare output against input rather than input against itself?

samhatfield: Exactly. 1e-13 is too strict, at least when testing on ECMWF/AC with GCC, MKL etc. I don't know if this indicates faulty behaviour or if it's to be expected.

samhatfield:
@AlexandreMary I can't add you as an official reviewer but please do take a look as well.

marsdeno (approved):
Looks good

AlexandreMary:
Oops, my bad for the comparison of the same arrays...
Looks good to me, thanks

PR #256, opened 2025-04-16
Static linking fix and etrans interface linking fix

awnawab:
A small PR with a couple of CMake linking fixes.


marsdeno (approved):
Looks good to me, nicecatch

wdeconinck (approved):
Thanks @awnawab , good to go!

PR #257, opened 2025-04-24
Correctly zero all elements of rspscalary

samhatfield:


samhatfield:
I don't know what's going on with the transi test. To be continued.

samhatfield:
Okay this is super weird. First issue: `ectrans_test_transi_invtrans_adjoint_gpu` is **not** reproducible on LUMI-G, even though we seed the RNG to `123`! If you move the `srand` call to _after_ `trans_setup(&trans)`, it becomes reproducible. Conclusion: something inside ecTrans is resetting the RNG only in the GPU library. CPU version doesn't have this problem. For now, we can call `srand` after `trans_setup` and continue to solve the problem.

After making this change, the test still fails, but at least one of the two values matches the CPU version:
```
# GPU version
[adjval1=116397298.114164][adjval2=116397299.433201]

# CPU version
[adjval1=116397299.433201][adjval2=116397299.433201]
```
The LHS comes from the adjoint version so there's still an issue with the GPU adjoint code. This issue only occurs on LUMI-G with OpenMP offload. The OpenACC version passes the test with the correct values matching the CPU version.

samhatfield:
Note that the two failing checks above (macos) are unrelated to this PR. These adjoint tests are proving to be quite temperamental. It seems that the only way to ensure that even correct adjoint code passes is to use quite a high tolerance, e.g. 100,000 times the machine epsilon. We need to have a proper discussion about how we want to deal with these adjoint tests.

tests/transi/transi_test.h
```diff
@@ -35,6 +35,7 @@
 #define ASSERT( assertion ) do {\
   if( !(assertion) ) {\
     printf("ERROR: Assertion `%s' failed @%s:%d\n",#assertion,__FILE__,__LINE__);\
+    TRANS_CHECK( trans_delete(&trans) );\
```
wdeconinck, line 38: Is this required? we're exiting anyway?

samhatfield: It's more of a neatness thing so that GPU arrays are properly deallocated. Without this you get a cryptic error message (which I can't remember off the top of my head) on exit when running on AMD devices with the Cray stack. Nvidia GPUs don't show the same behaviour so it may be that the NVHPC-compiled version has some kind of device garbage collection.

wdeconinck (approved):

PR #260, opened 2025-05-01
Change LUMI-G CI sbatch parameters

samhatfield:
I noticed that tests on LUMI-G were timing out as they can sometimes take longer than 20 minutes. This PR increases the CI job time limit to 40 minutes. It also submits jobs to `standard-g` instead of `dev-g` (the GPU development queue) as the latter only permits one job at a time.


marsdeno (approved):
All good

PR #264, opened 2025-05-09
Add LAM capability to transi via etrans

wdeconinck:
This consolidates work performed by @ddegrauwe late 2017 to bring LAM functionality to transi.
The work was performed as part of the ESCAPE-1 project, in a branch "feature/atlas_etrans" in the standalone `transi` repository made available to ESCAPE partners and users. At this time ectrans was not yet open-source and publicly available.

Now, with etrans part of ectrans, it was the perfect time to make this available to ectrans.

This work will open the way for Atlas to use spectral transforms for limited area grids via ectrans.

@ddegrauwe even though you did this work, I'd like you to review because we are now much later and things may have changed.

PR #265, opened 2025-05-14
Fix DIR_TRANS issue creating and using uninitialised values for PSPVOR and PSPDIV

wdeconinck:
When running the newly created "transi_test_lam" from PR #264 I ran into NaN when printing out vorticity and divergence.
After the best part of a day valgrinding, I tracked it down to the routine `EUVTVD`, and fixed it, see the PR changes.

In this PR I have set the uninitialised values to zero.
@ddegrauwe, could you or advise someone to please have a closer look if this is OK, or if there's an actual indexing bug? I'm not familiar with this and possible extension zones.

A second harmless change regarding checking for `PRESENT(KFLDPTRUV)` was also avoiding a valgrind warning "a jump was made depending on an uninitialised value"


ddegrauwe:
Hi @wdeconinck, I wasn't able to reproduce the NaNs (with Intel 2021.4). Which compiler(-settings) are you using?

Anyway, two remarks on this: 
* in transi_test_lam.c, the truncation should probably be set as
`trans_set_trunc_lam(&trans, 8, 9);`
instead of 
`trans_set_trunc_lam(&trans, 9, 8);`
* some entries of the PVOR/PDIV arrays indeed are not set in EUVTVD. These entries correspond to waves that fall outside the elliptic spectral truncation that's used in etrans, and they shouldn't make it to the final result (PSPVOR/PSPDIV arrays in EUPDSP)

ddegrauwe:
I don't see a lot of use cases to have higher waves than the Nyquist wavenumber, but still I think a safety like this (putting zeros) is better than aborting.

ddegrauwe (approved):

wdeconinck:
I rebased this on develop, changing the order of stacked PRs, so that this is clear this is not a "transi" development.

PR #226, opened 2025-03-12
Re-phase an optimisation lost after cycle 48

AlexandreMary:
Cherry-pick of PR ACCORD-NWP#8 from @RyadElKhatibMF on ACCORD-NWP fork.

Numerical impact: bit-reproducible
Computational impact: Positive or neutral, depending on the configuration.

PR #267, opened 2025-05-16
HOTFIX: Revert optimisation of DIST_GRID

samhatfield:
We detected that PR https://github.com/ecmwf-ifs/ectrans/pull/226 broke bit reproducibility in our IFS development CI suite. In fact there seems to be a bug in that the output of `DIST_GRID` in some circumstances contains spurious values in certain levels indicating uninitialised data. It's not clear yet whether the bug is in `DIST_GRID` or in the IFS. Until then unfortunately we need to undo this change so that we can have a stable release of ecTrans to include in the upcoming IFS CY50R1 release.

Our hope is to reintroduce this optimisation once the bug is fixed. When I next have some free time I will take a look (mid-June as I'm away for a few weeks soon).

@AlexandreMary I hope this is okay. To be clear, we have no reason to suspect @RyadElKhatibMF's code is wrong - it could well be that the previous `DIST_GRID` was too tolerant and that we are not preparing arrays properly on the IFS side. This PR will be merged to main which will permit a 1.6.1 release without the optimisation, but you can still use 1.6.0 which does contain the optimisation.

1.7.0 will hopefully also contain the optimisation, with a fix or workaround for the bug.


samhatfield:
Hotfix so no PR required, now I remember.

PR #270, opened 2025-05-21
Fix CI teardown

samhatfield:


.github/workflows/build-hpc.yml
```diff
@@ -164,11 +179,7 @@ jobs:
               export {{var}}
             {% endfor %}
 
+            blah
```
wdeconinck, line 182: beuh

samhatfield:
This PR adds an error trapping function so that work directories are cleaned up even when a CI job fails.

wdeconinck (approved):
Nice! That was indeed missing.

PR #271, opened 2025-05-21
GHA build-hpc: On LUMI, build with 1 task, and do not use ninja

wdeconinck:
Just for LUMI, this builds with only 1 thread, and doesn't use ninja.
- Using 1 thread because there seems to try to avoid an OpenACC compilation error:
   ```
    prfi1b_mod.F90-pp_1.acc.o: No such file or directory
    ftn-7932 ftn: ERROR in command line
    Error generating accelerator code: offload bundling failure
    ```
- Not using ninja because using ninja leads to many warnings of the Cray compiler:
    `The lister is using the source file "xxx.F90", rather than the specified source file "xxx.F90-pp.f90"`

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -57,7 +57,7 @@ jobs:
               - prgenv/nvidia
               - hpcx-openmpi/2.14.0-cuda
               - fftw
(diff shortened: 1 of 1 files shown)


samhatfield (approved):
All good!

PR #266, opened 2025-05-15
Add LAM adjoint

benjaminmenetrier:
This PR adds the code required for LAM adjoints in `transi`, with a bugfix in `src/etrans/cpu/internal/eftinv_ctlad_mod.F90` (inspired from `src/etrans/cpu/internal/eftinv_ctl_mod.F90`).

PR #272, opened 2025-05-28
LAM etrans tests fix

benjaminmenetrier:
I don't know any C, so I don't know if this PR is relevant to fix https://github.com/ecmwf-ifs/ectrans/pull/266#issuecomment-2892299758, but it made the problem disappear on my laptop...


wdeconinck:
This is anyway a good change! I'll merge it and let you know if it also fixes this for me.

PR #273, opened 2025-06-03
Fix gpnorm in LAM (for 50t1)

HaroldMTO:
Weights for LAM GP norms are wrong: it should be 1/r%ndgl, instead of 1/g%nloen.
they are now fixed with a reproducibility trick possible for a while
the trick is to use the env. var. EGPNORM_OLD: if set (non void), get the old wrong calculation


wdeconinck:
Oh I see, #274 supersedes this one! Thanks! I'll close this one now.

PR #274, opened 2025-06-03
Fix GP norms in LAM (for 50T1)

HaroldMTO:
Weights for LAM GP norms are wrong: should be 1/r%ndgl
they are now fixed with a reproducibility trick possible for a while
the trick is the env. var. EGPNORM_OLD: if set (non empty), get the old wrong calculation


HaroldMTO:
Comment: I don't know about this CI process. Seems to fail, is it normal or not? It's quite surprising...
I also would like to add a review from Daan Degrauwe. How can I do it? It seems I can't add reviewers...
Thanks for help.

src/etrans/cpu/external/egpnorm_trans.F90
```diff
@@ -78,27 +78,26 @@ SUBROUTINE EGPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL)
...
 
-DO JGL=1,R%NDGL
-  ZW(1:)=1._JPRB/G%NLOEN(JGL)
-ENDDO
-CALL GPNORM_TRANS_CTL(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,ZW(1:R%NDGL))
+CALL GET_ENVIRONMENT_VARIABLE("EGPNORM_OLD",CLENV)
+IF (LEN_TRIM(CLENV) > 0) THEN
+  ZW(:) = 1._JPRB/G%NLOEN
+ELSE
+  ZW(:) = 1._JPRB/R%NDGL
```
wdeconinck, line 94:
```suggestion
  ZW(:) = 1._JPRD/R%NDGL
```
For the `EGPNORM_OLD` behaviour we may want to stay bit-identical for now.

src/etrans/cpu/external/egpnorm_trans.F90
```diff
@@ -78,27 +78,26 @@ SUBROUTINE EGPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL)
 
 !ifndef INTERFACE
 
-! Local variables
-INTEGER(KIND=JPIM) :: JGL
-REAL(KIND=JPRD) :: ZW(R%NDGL)
-
+CHARACTER(LEN=80) :: CLENV
+REAL(KIND=JPRB) :: ZW(R%NDGL)
```
wdeconinck, line 82:
```suggestion
REAL(KIND=JPRD) :: ZW(R%NDGL)
```
This should be reverted to JPRD for API compatibility with `GPNORM_TRANS_CTL`

wdeconinck (approved):
This is great now for me now, thanks @HaroldMTO !
Pending @ddegrauwe .

ddegrauwe (approved):
Ok for me.

I'm just wondering about the potential penalty of reading an environment variable during every call. 
Maybe a switch with a SAVE attribute makes it possible to only read the environment variable once? Unless the penalty really is negligible of course.

wdeconinck:
> I'm just wondering about the potential penalty of reading an environment variable during every call. Maybe a switch with a SAVE attribute makes it possible to only read the environment variable once? Unless the penalty really is negligible of course.

I am pretty sure that this cost, even though a system call, is negligible compared to the following computations, including MPI operations etc.
Hopefully this can also be phased out after some time.

AlexandreMary:
Just one comment. We'll need to remember, when we switch to a version of ectrans that contains this (CY51 ?), that this metric that we use for validation will have changed, and so we'll have to use that env var to validate the switch of ectrans version.
Past that, we could remove the env var.

PR #276, opened 2025-06-17
Fix setup_trans0 and LAM initialization

benjaminmenetrier:
Hi @wdeconinck. Just a few modifications:
1) The content of `setup_trans0` should not be run more than once per execution, but it might be hard to control this when `transi` is used in different parts of a code. So I've added a `RETURN` if the key `MSETUP0` (already present) is not 0.
2) In the LAM initialization, the grid cell sizes `dx` and `dy` were not passed, so the derivative factors `pexwn` and `peywn` were kept to 1 (initialization value), leading to incorrect vorticity and divergence fields in the spectral transform.


wdeconinck:
Thank you @benjaminmenetrier

I was wondering, what are the dimensions of dx and dy, and why not use the same dimensions as pexwn and peywn ?

PR #277, opened 2025-06-19
Improved portability

SebastienRietteMTO:
The main goal is to suppress the constraint on the ctypesForFortran version to use, to allow versions 1.x and 2.x to be used


samhatfield:
Thanks @SebastienRietteMTO ! We'll take a look.

BTW: are you one of the authors of ectrans4py originally?

samhatfield:
Also please could you rebase your commit eb7587e436848bca27d83cb57ab3223dd669b8f2 on top of the develop branch and then push --force to [SebastienRietteMTO:SR_ectrans4py_improved_portability](https://github.com/SebastienRietteMTO/ectrans/tree/SR_ectrans4py_improved_portability)? It looks like you created your branch starting from main, but one should use the develop branch for new changes.

SebastienRietteMTO:
> Thanks @SebastienRietteMTO ! We'll take a look.
> 
> BTW: are you one of the authors of ectrans4py originally?

Yes, but before it was called ectrans4py (the first version was developed in 2014)

AlexandreMary:
@samhatfield yes, the commits are signed by me or Walid, because we did the externalisation from IAL, but Sébastien is the real initial author ;-)
And the author of the ctypesForFortran package for that matter, if you have any question about it.

samhatfield:
Previously our CI build used ctypesForFortran/1.2.0. Now it uses 2.2.0. This uses the [`six`](https://github.com/UMR-CNRM/ctypesForFortran/blob/2.2.0/src/ctypesForFortran/__init__.py#L32) module. However we don't have `six` installed, so the tests now fail. I'm not an expert but shouldn't `six` be added to the dependencies of the ctypesForFortran project [here](https://github.com/UMR-CNRM/ctypesForFortran/blob/2.2.0/pyproject.toml#L34)?

SebastienRietteMTO:
I'm building v2.2.1 without six,  I'll push it this afternoon.

SebastienRietteMTO:
I just pushed version 2.2.1 of ctypesForFortran. Could you run the test again? Thank you.

samhatfield:
> I just pushed version 2.2.1 of ctypesForFortran. Could you run the test again? Thank you.

That works now! Okay, we are ready to merge.

samhatfield:
> Would it be possible to have a numbered version with this in it ? Thanks

Hi @AlexandreMary - you mean like a version 1.7.0? Would you like that urgently?

AlexandreMary:
Well it's not urgent per se (I understand you may have other priorities for a next numbered version), but we have a new release of epygram pending, that relies on this ctypesForFortran-2.1, and hence requires this to be uploaded on PyPI.

BTW, what is the policy for tagging/labelling new releases of ectrans ? Is it based on time (say, not too long between 2 versions, or roughly every N months) or based on contents (expected or integrated) ? Tagging @marsdeno here too as we could discuss that in one of our upcoming regular meetings.

PR #286, opened 2025-07-03
Fix tests

### LINKED ISSUE #285: Unit test does not take arguments into account
### What happened?

1) The arguments provided to the benchmark unit tests are not taken into account, due to a mismatch between the arguments definition (base_args) and its usage (baseargs) in the CMakeLists.txt of the unit test (https://github.com/ecmwf-ifs/ectrans/blob/develop/tests/CMakeLists.txt)

Definition: base_args
https://github.com/ecmwf-ifs/ectrans/blob/2a06728c1eb0c97b350d308cfd586750cd380808/tests/CMakeLists.txt#L276

Usage: baseargs
https://github.com/ecmwf-ifs/ectrans/blob/2a06728c1eb0c97b350d308cfd586750cd380808/tests/CMakeLists.txt#L285

2) After fixing bug 1), the error generated by providing a wrong arguments in base_args is not catched : the unit test report "PASSED" but the tests fails to be launched.

3) base_args contains a list of arguments, which will be passed as a single argument to the command and will make the benchmark fail. 

1) add new arguments to base_args definition https://github.com/ecmwf-ifs/ectrans/blob/2a06728c1eb0c97b350d308cfd586750cd380808/tests/CMakeLists.txt and recompile. The new arguments are ignored

2) provide a wrong argument to "base_args", recompile and launch the unit test
```
 set( base_args "--niter 2 --meminfo --norms -v --wrong_argument" )
```

The unit test will display "PASSED"

3)  After renaming baseargs into base_args, run the test, and open /test/Temporary/LastTest.log.tmp:

Unrecognised argument: --niter 2 --meminfo --norms -v

### Version

1.6.1

Ubuntu GCC

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

samhatfield:
This PR contains three fixes:
1. `${baseargs}` in [tests/CMakeLists.txt](https://github.com/ecmwf-ifs/ectrans/blob/develop/tests/CMakeLists.txt) should be `${base_args}`. Without this the base args were not properly passed to the test definition. Fortunately only unimportant arguments were defined in this variable.
2. After fixing this, the `${base_args}` variable still wasn't expanded properly as CMake was inserting it with the quotes intact. I have removed the quotes from the definition of `${base_args}` and now it works.
3. Because we didn't throw an error code when exiting from the benchmark program when an invalid argument is provided, those cases didn't cause a test with an invalid argument to fail. That's now fixed by exiting with `error stop`.

Many thanks to @dhaumont for finding these!

Closes #285.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -1161,7 +1161,7 @@ subroutine parsing_failed(message)
     call print_help(unit=nerr)
   endif
   if (luse_mpi) call mpl_end(ldmeminfo=.false.)
-  stop
+  error stop
 
 end subroutine
 
--- a/src/programs/ectrans-lam-benchmark.F90
+++ b/src/programs/ectrans-lam-benchmark.F90
@@ -997,7 +997,7 @@ subroutine parsing_failed(message)
     call print_help(unit=nerr)
   endif
   if (luse_mpi) call mpl_end(ldmeminfo=.false.)
-  stop
+  error stop
 
 end subroutine
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -273,7 +273,7 @@ foreach( benchmark ${benchmarks} )
 
       # Base arguments -> 2 iterations, memory consumption/pinning information, spectral norms, and
       # verbose output
-      set( base_args "--niter 2 --meminfo --norms -v" )
+      set( base_args --niter 2 --meminfo --norms -v )
 
       set (base_title "${benchmark}_T${t}_${grid}_mpi${mpi}_omp${omp}")
 
@@ -380,7 +380,7 @@ if( HAVE_ETRANS )
 
       # Base arguments -> nlat x nlon, 2 iterations, memory consumption/pinning information,
       # spectral norms, and verbose output
-      set( base_args "--nlon ${nlon} --nlat ${nlat} --niter 2 --meminfo --norms -v" )
+      set( base_args --nlon ${nlon} --nlat ${nlat} --niter 2 --meminfo --norms -v )
 
       foreach( mpi ${ntasks} )
         foreach( omp ${nthreads} )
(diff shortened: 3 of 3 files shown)


wdeconinck (approved):
Thanks @samhatfield  looks good!

wdeconinck:
Nice catch @dhaumont !

dhaumont:
The bug 3 is still present in the lam benchmark.

dhaumont (changes requested):
please fix the arguments for all the unit tests, and make sure the bugs are also fixed for the LAM benchmark.

samhatfield:
It should be working now @dhaumont.

dhaumont:
@samhatfield Yes, I tested and everything is working now. Thanks for the fix and swift reaction!
LGTM!

dhaumont (approved):

wdeconinck:
uh-oh, failing test now with linux clang-18 debug build. This seems alarming.
```
At line 506 of file /home/runner/work/ectrans/ectrans/src/programs/ectrans-lam-benchmark.F90
  Fortran runtime error: Index '1' of dimension 3 of array 'zspsc3a' outside of expected range (1:0)
  
  Error termination. Backtrace:
  #0  0x7f3c77023e59 in ???
  #1  0x7f3c77024a71 in ???
  #2  0x7f3c77025082 in ???
  #3  0x55866cde19a1 in ectrans_lam_benchmark
  	at /home/runner/work/ectrans/ectrans/src/programs/ectrans-lam-benchmark.F90:506
  #4  0x55866cdf7228 in main
  	at /home/runner/work/ectrans/ectrans/src/programs/ectrans-lam-benchmark.F90:44
```

dhaumont:
Before these fixes, no test was executed. So there was no way to detect this failure and it was certainly already present before.
Can't we merge merge this, and create a new issue with all the failing unit tests? As you will see in this other #287, some other tests are failing in lam (maybe related to this issue)

PR #289, opened 2025-07-09
Add safeties in lam benchmark for case nfld=0

ddegrauwe:
LAM tests were failing with clang in case nfld=0 because of passing empty arrays to norms computation.
Norms computations have been put under `if ( nfld>0 )` and `if ( lvordiv )` to avoid this situation.

### DIFF
--- a/src/programs/ectrans-lam-benchmark.F90
+++ b/src/programs/ectrans-lam-benchmark.F90
@@ -472,12 +472,6 @@ endif
 
 ndimgmv = jend_scder_EW
 
-!allocate(zgmv(nproma,nflevg,ndimgmv,ngpblks))
-!allocate(zgmvs(nproma,ndimgmvs,ngpblks))
-!zgpuv => zgmv(:,:,1:jend_vder_EW,:)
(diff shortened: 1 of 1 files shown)


wdeconinck:
Thanks @ddegrauwe for fixing! Even though you fixed the tabs, there still seems to be some whitespace inconsistencies if you look at the diff.

src/programs/ectrans-lam-benchmark.F90
```diff
@@ -1338,15 +1393,15 @@ subroutine dump_gridpoint_field(jstep, myproc, nlat, nproma, ngpblks, fld, fldch
 #endif
     open(noutdump, file=filename, form="unformatted", access="stream")
     write(noutdump) kgptotg/nlat,nlat ! dimensions
-	write(noutdump) fldg ! data
+  write(noutdump) fldg ! data
```
samhatfield, line 1396:
```suggestion
    write(noutdump) fldg ! data
```

samhatfield:
Well spotted @ddegrauwe. Just a couple more suggestions for code style.

ddegrauwe:
Thanks a lot for the suggestions @wdeconinck and @samhatfield .

PR #292, opened 2025-07-15
Add ectrans4py function for retrieving "Legendre assets" from ecTrans

samhatfield:
For AIFS work, we are writing a spectral transform in native PyTorch. Rather than generating our own Legendre polynomials (and Gaussian quadrature weights) we decided to use ecTrans' existing functionality. This PR adds a function `get_legendre_assets` to the ectrans4py interface to permit one to obtain the Legendre polynomials and Gaussian weights from ecTrans in Python. It also returns the "highest zonal wavenumber for each latitude" array (`NMEN`) as this is required to perform a spectral transform that matches what ecTrans does.

This PR also contains two related bugfixes:
- In trans_inq4py.F90, `INMENG` is set to `KNMENG` first, even though `KNMENG` is an `INTENT(OUT)` uninitialised variable. `INMENG` is set by `TRANS_INQ` so this should be deleted.
- Previously it wasn't possible to get `PRPNM` (Legendre polynomials) through `TRANS_INQ` because of a check on `.NOT. S%LKEEPRPNM`. The latter variable only has meaning when the fast Legendre transform is enabled, otherwise polynomials are always kept. I have added a second Boolean so that this check only happens when the FLT is enabled.


src/ectrans4py/get_legendre_assets.F90
```diff
@@ -0,0 +1,70 @@
...
+!    KNMENG: cut-off zonal wavenumber
+!    PGW: Gaussian weights
+!    PRPNM: Legendre polynomials
+!
+! ** AUTHOR
+!    2 July 2025, S. Hatfield
+!
+! I. Dummy arguments declaration
+IMPLICIT NONE
+INTEGER(KIND=8), INTENT(OUT) :: KRETURNCODE
```
marsdeno, line 23: Not suggesting the following for this PR, as the style here is done to match ectrans4py, but should we not aim to use kind specifiers that are more robust than `integer(kind=8)` ?
Something like `kind=int64` from the iso_fortran_env module might be an acceptable alternative?

marsdeno (approved):
Looks fine to me.

PR #293, opened 2025-07-18
Restore compatibility with IFS

reuterbal:
Two conceptual modifications in this PR:

1. The changes in b0475fd8d13609f9f603a888205df7d2d2caa5fd broke use in IFS, thus reverting this change. 
2. The CMake for etrans used wrong target names in several places, resulting in missing headers when using etrans interface includes in IFS.


wdeconinck (approved):
These changes look good to me! Thanks for catching all this @reuterbal !

PR #297, opened 2025-07-29
Remove unused variables from GPU code path

samhatfield:
There are tons of unused variables, both declared and imported, in the GPU code path. Here I try to remove as many as possible.


wdeconinck (approved):
CI seems to pass on all gpu runners. Great job!

samhatfield:
> CI seems to pass on all gpu runners. Great job!

Thanks. You only do these things when everyone is away on holiday 😅

PR #299, opened 2025-08-05
[DRAFT] Allow GPU code path to be executed on CPU

samhatfield:
This PR implements a "host backend" for the GPU code path. Essentially I have made some minor tweaks to the Fortran and implemented host-side C++ backends for the Fourier and Legendre transforms, using the existing LAPACK and FFTW functionality brought in by the CPU version.

Along the way I have tried to separate out "host OpenMP" and "accelerator OpenMP" functionalities into separate CMake features, as I didn't want to affect the CPU build when switching off OpenMP for the GPU (actually also running on CPU) build. When this PR is working we should properly scrutinise how all the various features and builds interact.

Right now the host GPU backend is not working numerically, hence why this is still a draft. I'm almost there though... see this screenshot of the transformed spherical harmonic used in the benchmark case:

Screenshot 2025-08-05 at 11 19 12

It seems to be correct, except for some latitudes which have obvious error patterns. Really weird. Not sure what's going on there, but I will keep testing.


samhatfield:
I forgot to mention why I'm doing this! The GPU code path has a number of data layout changes and general optimisations that we expect will improve also CPU performance. In the long term, we may therefore want to use this code path for operations instead of the "legacy" CPU version. But we need to do a proper side-by-side comparison first.

Also, this will be useful for debugging the behaviour of the GPU version when it's actually running on GPU.

samhatfield:
The latest commit seems to fix the above error.

PR #302, opened 2025-08-14
Fix for cublas link failures

awnawab:
When building on systems where libcublas lives in a different folder to libcudart, ectrans suffers from link failures. In such a configuration, FindCudaToolkit.cmake correctly finds cublas and sets the target `CUDA::cublas` to the full path of this library. It is this full path to ectrans that appears on the link line. However when building the executable, the linker tries to simply link to `-lcublas` (even though this isn't on the link line, hence my suspicion that it's a transient dependency) and therefore expects the cublas parent directory to also be on the link line  i.e. `-L<path-to-cublas-dir>`. FindCudaToolkit.cmake however only adds the cudart parent directory to the link line, not that of cublas.

This PR addresses the above and also adds the parent directory of `-lcublas` to the link line. Ultimately this should probably be a patch to FindCudaToolkit.cmake but it should be harmless to also include it here.

### DIFF
--- a/cmake/ectrans_find_cuda.cmake
+++ b/cmake/ectrans_find_cuda.cmake
@@ -21,5 +21,18 @@ macro( ectrans_find_cuda )
       ecbuild_info( "cuda arch               : [${CMAKE_CUDA_ARCHITECTURES}]" )
       ecbuild_info( "cublas                  : [${CUDA_cublas_LIBRARY}]" )
       ecbuild_info( "cufft                   : [${CUDA_cufft_LIBRARY}]" )
+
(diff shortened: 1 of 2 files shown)


awnawab:
> Sounds good. Is [this commit](https://github.com/ecmwf-ifs/ectrans/pull/302/commits/79a3a0e6d8157c766e51858861472ef0dc770e01) unrelated?

No in the case of static linking, the link to cublas will be delayed until the final executable, so we need to propagate the extra link directory there too.

marsdeno (approved):
Looks good to me

PR #304, opened 2025-08-20
Reduce test set size

samhatfield:
Currently there are 530 tests executed for the GPU CI actions which is starting to reach the reasonable limit in terms of the job wall time on the HPC systems these actions run on. Here is the breakdown for the 530 tests:
- 1: ectrans_test_install
- 3: ectrans_test_setup_trans0
- 6: ectrans_test_adjoint
- 12: ectrans_test_{dir/inv}trans_adjoint
- 6: ectrans_test_gpnorm_trans_adjoint
- 288: benchmark tests (CPU)
- 72: benchmark tests (GPU)
- 126: LAM benchmark tests (CPU)
- 16: transi tests

The big one is the CPU benchmark tests. For each precision in "dp sp", for each ntasks in "0 1 2", for each nthreads in "1 4 8", and for each callmode in "1 2", we execute the following:
- benchmark with nfld 0
- benchmark with nfld 10
- benchmark with nfld 10 and nlev 20
- benchmark with nfld10, nlev 20 and scders
- benchmark with nfld 10, nlev 20 and vordiv
- benchmark with nfld 10, nlev 20, vordiv, and uvders
- benchmark with nfld 10, nlev 20, nproma 16
- benchmark with nfld 10, nlev 20, flt

I don't think we need to do all of these. Instead I suggest we do just:
- benchmark with nfld 1, nlev 137, scders, vordiv, uvders
- benchmark with nfld 1, nlev 137, scders, vordiv, uvders and nproma 16
- benchmark with nfld 1, nlev 137, scders, vordiv, uvders and flt

This brings the total number of tests down to 302. These consolidated tests mostly have the same coverage as the old ones, except that there's no nfld 0 test, but do we actually care about that?

We could probably reduce the 126 LAM tests as well (especially as we'll be adding GPU LAM tests at some point).

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -281,7 +281,7 @@ foreach( benchmark ${benchmarks} )
         # Check it works with 10 3D scalar fields, 20 levels, and vordiv in grid point space
         ecbuild_add_test( TARGET ${base_title}_nfld10_nlev20_vordiv
             COMMAND ${benchmark}
-            ARGS --truncation ${t} --grid ${grid} --nfld 10 --nlev 20 --vordiv --check 200  --callmode ${callmode} --dump-checksums ${base_title}_nfld10_nlev20_vordiv ${base_args}
+            ARGS --truncation ${t} --grid ${grid} --nfld 10 --nlev 20 --vordiv --check 100  --callmode ${callmode} --dump-checksums ${base_title}_nfld10_nlev20_vordiv ${base_args}
             MPI ${mpi}
             OMP ${omp}
         )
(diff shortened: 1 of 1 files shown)


wdeconinck:
I am reluctant to remove *all* of these tests.
What we can also do to remove certain tests on e.g. LUMI, is filter them out with some regex or labels.

Even if the IFS does not make use of all these configurations, they are valid code paths.

For CI we also don't need 137 levels. We want to test correctness, and code paths, not speed.
We should instead have a few select benchmark configuration that we can either manually run, or as an extra step after the CI on the HPC or a new workflow on the HPC. These should reflect our operational runs.

samhatfield:
Okay, second go, and now we are down to 375 tests when GPU is enabled. We're now doing the following tests in the MPI/OMP loop:
- nfld0
- nfld10_nlev20
- nfld10_nlev20_derivatives (--scders and --uvders)
- nfld10_nlev20_vordiv
- nfld10_nlev20_nproma16

I've remove the case with 4 threads so we're only testing with 1 and 8 threads now.

samhatfield:
Further reduced to 333 tests by removing the 4-thread option for LAM tests.

PR #307, opened 2025-08-27
Remove PROMATR capability

samhatfield:
`NPROMATR` (option `KPROMATR` of `SETUP_TRANS0`) allows one to split up the transform of a large number of fields into "blocks" (also called "packets" in the code). One "global" transform is split into a loop over the number of blocks which is determined by the total number of fields divided by `NPROMATR` (see [here](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/dir_trans_ctl_mod.F90#L128)).

This feature is presumably to save memory consumed by some of the work arrays of the transforms, which of course only need to be big enough to store the number of fields in a block at one time.

This capability comes with a need for two further subroutines `FIELD_SPLIT` and `SHUFFLE`. The former calculates the number of different types of fields for each block and indices for slicing `KVSET` arrays etc. I have no idea what the latter does after looking at it on and off for a year to be honest. Something to do with load balancing.

The `NPROMATR` capability only supports "call mode 1" (pass in `PGP`, receive `PSPVOR`, `PSPDIV`, `PSPSCALAR` and vice versa) which means it's not used by any of the important calls to ecTrans in the IFS (we mostly use call mode 2). Therefore from the ECMWF side, we should be able to deprecate this feature without harming any operational system.

On the positive side we can remove 1300 lines of code. Less code to maintain and think about.

There are only two reasons why I think we might have to keep this feature:
- If it's used in ARPEGE.
- If it's used by any of the applications of etrans.

So this PR is intended as a discussion to how or if we can deprecate this feature to help reduce the size, complexity, and maintenance requirements of ecTrans.

@ioanhadade's idea.

@wdeconinck @marsdeno @dhaumont @ddegrauwe @RyadElKhatibMF (maybe @AlexandreMary too)


RyadElKhatibMF:
NPROMATR (as well as its equivalent NEPROMATR for the LAM transforms) is used in operations at Météo-France.
It serves for changes of geometries made by Fullpos : small jobs in term of CPU time, but they need to read and transform many large fields. NPROMATR is very helpful to save memory for these jobs, allowing to allocate a single node.
Therefore we should keep it.

samhatfield:
> NPROMATR (as well as its equivalent NEPROMATR for the LAM transforms) is used in operations at Météo-France.
> It serves for changes of geometries made by Fullpos : small jobs in term of CPU time, but they need to read and transform many large fields. NPROMATR is very helpful to save memory for these jobs, allowing to allocate a single node.
> Therefore we should keep it.

Many thanks @RyadElKhatibMF for the prompt response. That's all we wanted to know. Sometimes it's not clear when a feature is effectively deprecated or not. In this case it's good to know this code path is still important, so we will keep it as you say.

wdeconinck:
OK, but just choose one test configuration as not to multiply the number of tests.

dhaumont:
It should be possible to refactor the PROMATR code using the new field API interface, and make it much smaller.
To be investigated further.

PR #308, opened 2025-08-27
Satisfy various warnings by the Cray compiler

samhatfield:


wdeconinck:
Nice, still quite a number of warnings to consider, looking at the log.

samhatfield:
> Nice, still quite a number of warnings to consider, looking at the log.

Yeah, I've basically ignored "Fortran allocatable variable "..." is being used before being allocated." as they're just ... wrong. E.g.:

```
ftn-7206 ftn: WARNING CONSTRUCT_BUTTERFLY, File = .../src/trans/cpu/generated/ectrans_dp/algor/butterfly_alg_mod.F90, Line = 79 
Fortran allocatable variable "zbcomb" is being used before being allocated.
```
The [line](https://github.com/ecmwf-ifs/ectrans/blob/satisfy_cce_warnings/src/trans/cpu/algor/butterfly_alg_mod.F90#L79) looks fine to me. In fact I don't think any of the errors of that type are genuine, so must be an issue with CCE.

There are a couple genuine ones though, true:
1. [dealloc_resol_mod.F90](https://github.com/ecmwf-ifs/ectrans/blob/satisfy_cce_warnings/src/trans/gpu/internal/dealloc_resol_mod.F90#L121): `Variable "r_" is used before it is defined`
2. fsc_mod.F90: `The result of function name "HFSC" in the function subprogram is not defined.`

For 1., maybe we should be nullifying those pointers (`=> null()`). @lukasm91 I think you wrote this eventually. Is there a reason we assign the resol-specific pointers to uninitialised objects rather than just nullifying them?

For 2., `FSC` doesn't require any work arrays so this function doesn't do anything currently. I guess we can just remove it.

samhatfield:
Nice work @wdeconinck - it's looking a lot cleaner now.

BTW is this warning a concern?

```
/opt/cray/pe/cce/17.0.1/binutils/x86_64/x86_64-pc-linux-gnu/bin/ld: warning: size of symbol `$data_init$yomhook_' changed from 4 in {file}.F90.o to 8 in /scratch/***/github-actions/ectrans/17436523330/1/ecmwf-ifs/fiat/installation/lib64/libfiat.so
```

wdeconinck:
This linker warning has been showing before on our own Cray, and I never really understood how this happens. I also don't see anything strange about yomhook. It has not caused issues before.

PR #309, opened 2025-08-28
Add test cases for NPROMATR feature

samhatfield:


samhatfield:
As I test, I measured the memory consumption of a benchmark at TCO319, 137 levels, callmode 1 (which is mandatory of course) and with U/V and scalar derivatives enabled, varying NPROMATR:

memory

Memory is reported under the `MEMORY USED(MB)` column when the `--meminfo` flag is passed.

As expected memory consumption goes down with NPROMATR. Smaller field blocks means smaller work arrays.

Surprisingly the wall time also goes down, with NPROMATR=8 running about 2.5x faster than NPROMATR=412 (which means the option is disabled)!

walltime

This is all for a single task running on my Macbook Pro. Other platforms may vary, especially when running with multiple tasks. Still, it hints that NPROMATR could actually be an important feature even for reducing walltime.

To make this genuinely useful we'd have to allow blocking also with callmode 2.

PR #310, opened 2025-09-01
Add etrans dummy library when ETRANS is disabled

samhatfield:
Currently, it isn't possible to build an IFS bundle with the ETRANS feature disabled. The IFS contains calls to etrans which are only guarded by runtime `IF`s (rather than preprocessor `#if`s).

This makes it harder to experiment with the global transforms of ecTrans while testing within the IFS. The ETRANS feature must be kept enabled in order for the IFS build to succeed, which means all of the etrans source files are compiled. Some subroutines are shared between global and local transforms. This means that when a modification is made to suit the global transforms, a similar modification must also be made to the local transforms in order for the code to compile.

Eventually any significant modification to the code shared by the global and local transforms must work in both cases. But this should not hinder early development which focuses only on the global transforms.

This PR adds a etrans "dummy" library. When the ETRANS feature is disabled, the etrans library is still built. However it only contains the external subroutines, and these all abort immediately when called. This means one can disable the ETRANS feature and still build the IFS against ecTrans.


src/etrans/cpu/CMakeLists.txt
```diff
@@ -47,11 +47,71 @@ endfunction(generate_backend_sources)
...
+  set(${_PAR_OUTPUT} ${outfiles} PARENT_SCOPE)
+endfunction(generate_etrans_dummy_sources)
+
 foreach( prec dp sp )
   if( HAVE_${prec} )
 
     generate_backend_includes(BACKEND ${prec} TARGET ectrans_etrans_${prec}_includes DESTINATION ${BUILD_INTERFACE_INCLUDE_DIR} INCLUDE_DIRECTORY ${PROJECT_SOURCE_DIR}/src/etrans/include )
-    generate_backend_sources( BACKEND ${prec} OUTPUT ectrans_etrans_${prec}_src  DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/generated/ectrans_etrans_${prec})
+
+    if( NOT HAVE_ETRANS )
```
reuterbal, line 101: I guess longer term we might want to consider making the dummy generation optional, too. But fine for this PR.

wdeconinck: I think we can do this indeed but then provided that ifs-source can link to ectrans without any symbols needed.

PR #311, opened 2025-09-04
Switch from omp_get_wtime to timef in LAM benchmark

samhatfield:
The main benchmark program uses [`timef`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/programs/ectrans-benchmark.F90#L238) to calculate wall time. This appears to be a non-standard vendor extension which is accepted by most compilers.

The LAM program uses [`omp_get_wtime`](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/programs/ectrans-lam-benchmark.F90#L251), which again is not ideal because it introduces an obligation to enable `OMP` in order to compile and run the LAM benchmark.

This PR introduces a simple wall time function based on the standard `SYSTEM_CLOCK` for use in both benchmark programs.

Once this PR is clear I will introduce another removing the `OMP` feature [constraint](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/programs/CMakeLists.txt#L77) on the LAM benchmark program, with an appropriate generator expression. Currently **it's not possible to enable `ETRANS` while disabling `OMP`** (assuming `OpenMP_Fortran_FLAGS` is unset).

Once that's done, I'll be able to disable the `OMP` feature on the amdflang-based action pending resolution of a runtime error due to inability to find the OpenMP library (see [#248](https://github.com/ecmwf-ifs/ectrans/pull/248)).

### DIFF
--- a/src/programs/ectrans-lam-benchmark.F90
+++ b/src/programs/ectrans-lam-benchmark.F90
@@ -43,7 +43,6 @@ program ectrans_lam_benchmark
 
 use parkind1, only: jpim, jpib, jprb, jprd
 use oml_mod ,only : oml_max_threads
-use omp_lib, only: omp_get_wtime
 use mpl_module
 use yomgstats, only: jpmaxstat
 use yomhook, only : dr_hook_init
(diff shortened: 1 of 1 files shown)


samhatfield:
I see, that's news to me. Then we should stick with `timef` and make sure the LAM benchmark uses it too. And also make a note that this comes from FIAT (and isn't to be confused with e.g. https://www.ibm.com/docs/en/xl-fortran-linux/16.1.0?topic=procedures-timef). Will update the PR.

wdeconinck (approved):
That looks good Sam! Thanks!
Let's change the PR title and then you can squash-merge when you're ready.

ddegrauwe:
> @ddegrauwe I think you originally chose `omp_get_wtime`. I thought you would want to know about this change.

Fine by me to change to TIMEF. The reason we introduced omp_get_wtime in the LAM benchmark was because TIMEF gave substantially different results on LUMI (Cray compiler), making comparison with other compilers difficult.

PR #312, opened 2025-09-05
Improve compare_checksum

samhatfield:
Another PR to fix edge cases. Currently compare_checksums always runs even if MPI is disabled. This test is supposed to compare results between different numbers of MPI tasks, so it's not appropriate to run it if MPI is disabled.

Also compare_checksums has hard-coded the number of tasks and threads to check, so I've modified it to accept these as arguments passed from CMake, based on `ntasks` and `nthreads`.

Also it didn't exit correctly when no test results were found (this is a fail), nor when the wrong arguments are passed (this could happen and go unnoticed if the CMake call is modified in future and care isn't taken to pass the arguments properly.

This PR is needed in order to permit [amdflang to be added to the CI](https://github.com/ecmwf-ifs/ectrans/pull/248) (we are disabling MPI there).

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -229,7 +229,7 @@ foreach( benchmark ${benchmarks} )
     endif()
   endif()
 
-  if( NOT TEST compare_checksums AND HAVE_MPI )
+  if( NOT TEST compare_checksums AND (HAVE_MPI OR HAVE_OMP) )
     configure_file( compare_checksums.py ${CMAKE_CURRENT_BINARY_DIR}/compare_checksums.py @ONLY )
     list( JOIN ntasks "," ntasks_joined )
     list( JOIN nthreads "," nthreads_joined )


wdeconinck:
Thanks Sam that is indeed better. We still have to cater for the case where we don't have MPI, but still use OpenMP.
That's important to detect OMP race conditions regardless of MPI.

samhatfield:
> Thanks Sam that is indeed better. We still have to cater for the case where we don't have MPI, but still use OpenMP. That's important to detect OMP race conditions regardless of MPI.

Ah yeah, good point. So it should be `HAVE_MPI OR HAVE_OMP`?

samhatfield:
Updated!

dhaumont:
> > I don´t understand why you decided to restrict the usage of compare_checksum.py to openmp and mpi.
> > I implemented it in the first place to compare results between the field_api implementation and the regular implementation. It is also very useful to compare the results for the MPI/openmp with the ones from the single thread implementation.
> 
> compare_checksum.py compares a list of cases against a reference right? The reference is the case with one thread and one task. So if OpenMP and MPI are disabled, what is there to compare against the reference?

Thanks for the clarification. I understood from your sentence that you deactivated dump-checksum for single thread execution. After looking more carefully into the code, I see that you only deactivated the reference comparison, which makes a lot of sense of course.

PR #313, opened 2025-09-24
Add API test suites

samhatfield:
Currently we have the following tests in ecTrans:
- ectrans_test_install: test whether a program can be built against ecTrans
- ectrans_test_setup_trans0: a basic test for setup_trans0 with one fixed set of arguments
- ectrans_test_adjoint: a high-level test of the tangent-linear/adjoint correspondence of `INV_TRANS`, `INV_TRANSAD`, `DIR_TRANS`, `DIR_TRANSAD`, all in one
- Separate tests of tangent-linear/adjoint correspondence of `INV_TRANS/AD` and `DIR_TRANS/AD`
- A test of tangent-linear/adjoint correspondence of `GPNORM_TRANSTL/AD`
- Tests of ecTrans global and LAM transforms under many settings, using the benchmark program. These are sort of integration tests
- transi tests

I propose to add a mechanism for unit tests, by which I mean specifically _an isolated test for a single feature of one of the ecTrans subroutines_. I suggest to focus on the external subroutines (at least for now). These tests will complement the tests above, and in some cases possibly replace them (e.g. we won't need `ectrans_test_setup_trans0` anymore, and the benchmark program could perhaps finally be repurposed for what it was built for - benchmarking).

This draft PR gives an idea of how this could be achieved in vanilla CMake. I have followed [this example](https://github.com/dryman/fortran_unit_test). Here is how it works:
- There is a new directory: tests/unit.
- In this directory we have a directory for each external subroutine (currently just setup_trans0 and setup_trans).
- In each of these, we write a plain old Fortran module with a `INTEGER FUNCTION` for each unit test. **These all are named `UNIT_TEST_*`**. Each function carries out a unit test and returns 0 for success or 1 for failure. **They all have `BIND(C)` specified so they can be called from a C function linked into the final executable.**
- In the CMakeLists.txt file for this suite, we define again the tests in the corresponding Fortran file in a list.
- Using `create_test_sourcelist` we create a C program which provides a callable wrapper for each of the Fortran unit tests. From this we create an executable which CTest can call.
- We then add each unit test one by one so CTest is aware of them, using our C program as the test executable.

The above logic is repeated for each enabled precision.

All unit tests will then be visible when running `ctest -R unit_test`.

Here are the tests I've implemented so far:
```
➜  build ctest -N -R unit_test
Test project /Users/samhatfield/Work/ECMWF/ectrans/build
  Test #188: unit_test_setup_trans0_eq_regions_dp
  Test #189: unit_test_setup_trans0_eq_regions_sp
  Test #190: unit_test_setup_trans_without_setup_trans0_dp
  Test #191: unit_test_setup_trans_basic_dp
  Test #192: unit_test_setup_trans_odd_ndgl_dp
  Test #193: unit_test_setup_trans_octahedral_dp
  Test #194: unit_test_setup_trans_stretching_dp
  Test #195: unit_test_setup_trans_all_fftw_dp
  Test #196: unit_test_setup_trans_belusov_dp
  Test #197: unit_test_setup_trans_flt_dp
  Test #198: unit_test_setup_trans_without_setup_trans0_sp
  Test #199: unit_test_setup_trans_basic_sp
  Test #200: unit_test_setup_trans_odd_ndgl_sp
  Test #201: unit_test_setup_trans_octahedral_sp
  Test #202: unit_test_setup_trans_stretching_sp
  Test #203: unit_test_setup_trans_all_fftw_sp
  Test #204: unit_test_setup_trans_belusov_sp
  Test #205: unit_test_setup_trans_flt_sp

Total Tests: 18
```

I have implemented a "will fail" logic to handle cases that are expected to fail (e.g. calling `SETUP_TRANS` without `SETUP_TRANS0` or initialising with an odd number of latitudes - both cases above). However this will require modifying the abort behaviour of [`SDL_SRLABORT`](https://github.com/ecmwf-ifs/fiat/blob/main/src/fiat/system/sdl_mod.F90#L88): we need to somehow disable the raising of `SIGABRT` and use `ERROR STOP` instead of `STOP` so an error code 1 is returned. Otherwise I don't think there's any way to recover from `ABORT_TRANS`.

Also, for now I've disabled MPI in these tests. We obviously need to test at least some of the subroutines with MPI enabled (e.g. `DIST_GRID`) but I'm not sure how to do this yet without wrapping the whole suite in a for loop over 0, 1, 2 MPI tasks and having the number of tests exploding.

Anyway, any thoughts @wdeconinck @marsdeno @ddegrauwe @dhaumont @RyadElKhatibMF?


samhatfield:
Ninja doesn't like the fact that I generate the same Fortran modules twice. Need to find a way to have a precision-qualified module file emitted when each test suite is compiled.

dhaumont:
We also need to think about:
- testing for GPU and CPU
- test for different grid partitionning (link to MPI)
- test  the new field-api interface
- cover the LAM

samhatfield:
I have renamed the "unit" tests to "API" tests.

samhatfield:
PR reduced in scope to only include new testing infrastructure and test suite (actually just a single test) for `SETUP_TRANS0`.

.github/workflows/build.yml
```diff
@@ -189,7 +189,7 @@ jobs:
         dependency_cmake_options: |
           ecmwf-ifs/fiat: "-G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DENABLE_TESTS=OFF -DENABLE_MPI=ON"
         cmake_options:    "-G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.cmake_options }} -DENABLE_MPI=ON -DENABLE_FFTW=ON -DENABLE_ETRANS=ON"
-        ctest_options: "${{ matrix.ctest_options }}"
+        ctest_options: "${{ matrix.ctest_options }} -LE will_fail"
```
wdeconinck: I would not exclude the tests labeled as "will_fail" from running? Don't you want these to run as well? The test property WILL_FAIL should invert the criteria of passing, so that tests that fail but marked with WILL_FAIL will actually pass.

tests/trans/api/setup_trans0/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+# API tests for SETUP_TRANS0
+
+# Define list of SETUP_TRANS0 tests
+set( test_list
+  api_test_setup_trans0_eq_regions
```
wdeconinck:
```suggestion
  ectrans_test_trans_api_setup_trans0_eq_regions
```
The test names need to be more descriptive and easily identifiable as an ectrans test for trans (especially within a bundle).
We should be able to only test trans with`ctest -R ectrans_test_trans` and only api tests then with `ctest -R ectrans_test_trans_api`

I guess this will require a matching change in function name due to the create_test_source_list function.

wdeconinck (changes requested):
Nice initiative @samhatfield,  I think this will be nice going forward when more APIs will get tested.  Then we will need common routines for grid setup, initial condition, memory allocation (pinning), ... , which are all the extra subroutines in the ectrans-benchmark program. I think following this PR we will need to consolidate something like this.

For this PR, I think what's missing in the `generate_api_tests` function, are options to set how many MPI tasks to use, and possibly OMP threads. Also perhaps being selective on a backend CPU/GPU or precision would be nice.

I think the main advantage of going with this `create_test_source_lists` driver is reduced amount of binaries to compile/link. For shared libraries this is probably not a big deal, but it is nice for static builds for sure. Let's see if the driver and linking in C will pose issues when  GPU backend is tested for other APIs.

.github/workflows/build.yml
samhatfield: The reason I excluded them is because it's not possible to recover gracefully from a call to `ABOR1` currently. If I remember right, it's because `SDL_SRLABORT` raises `SIGABRT` which CTest can't handle. I think it needs a "failed" test to exit cleanly with error code 1. We could modify `SDL_ARLABORT` to exist gracefully somehow.

wdeconinck: It appears that this is a ctest design decision. "Abort == crash" (counting as unexpected fail) whereas "Fail == non-zero exit code".

Some testing shows that tests using mpiexec seem to work with the WILL_FAIL, as I guess it is mpiexec that exits rather than aborts.

Perhaps we can leave out the WILL_FAIL for now.

Anyways, if we are truly wanting to test an abort, we should also check the abort message to be certain of the abort reason. There exists the test property "PASS_REGULAR_EXPRESSION" but that has the same problem with the abort.

We should run the abort cases with a wrapper (cmake?) script that checks for output. This becomes all quite messy with this automatic suite generation, but could be done. This is an example how one could do it:
https://github.com/ecmwf-ifs/fiat/blob/e20641eb7d6a9747ce8e3273b66255a6cd31713a/tests/CMakeLists.txt#L31-L43

tests/trans/api/setup_trans0/CMakeLists.txt
samhatfield: Good idea.

samhatfield:
> For this PR, I think what's missing in the `generate_api_tests` function, are options to set how many MPI tasks to use, and possibly OMP threads. Also perhaps being selective on a backend CPU/GPU or precision would be nice.
> 
> I think the main advantage of going with this `create_test_source_lists` driver is reduced amount of binaries to compile/link. For shared libraries this is probably not a big deal, but it is nice for static builds for sure. Let's see if the driver and linking in C will pose issues when GPU backend is tested for other APIs.

Yeah, you raise a good point there at the end. I think I should investigate that before we merge this. There's no point in adopting an API testing framework which is fundamentally incompatible with GPU builds.

wdeconinck:
The framework seems only to require to define a BIND(C) function that returns an "int". That is very generic andcould very well still be compatible.

I don't think it would be any difference to the actual test code for any of the tests if we would later need to create a Fortran template program instead and have a CMake function to handle it.

So I'd just go for this still and then we see later.

.github/workflows/build.yml
samhatfield: Given that we'll be needing to run ctest basically always with `-LE will_fail` for the foreseeable future, do you think we should just delete those tests?

samhatfield:
I've modified the test suite generation code so it iterates over the list of platforms (CPU and GPU) and generates tests for each. You can't test it in this branch because only `SETUP_TRANS0` tests are implemented here, and those tests are not replicated for each platform. However I do have a followup PR in preparation which implements `DIR_TRANS` tests etc. so I can confirm that this logic works there.

So I think the last thing I need to address is

> For this PR, I think what's missing in the generate_api_tests function, are options to set how many MPI tasks to use, and possibly OMP threads. Also perhaps being selective on a backend CPU/GPU or precision would be nice.

Do you mean that I should add `precisions`, `platforms` and `ntasks` (and maybe `nthreads`) as arguments to `generate_api_test_suite`? Currently they're inherited from the parent scope which is a bit nasty.

wdeconinck:
> Do you mean that I should add `precisions`, `platforms` and `ntasks` (and maybe `nthreads`) as arguments to `generate_api_test_suite`? Currently they're inherited from the parent scope which is a bit nasty.

I was thinking indeed that there need to be variables for this, We can also combine the ntasks and nthreads in one variable MPIxOMP, and the platform and precision perhaps BACKEND
```cmake
generate_api_test_suite( ...
                         MPIxOMP 
                             0x1 1x1           # serial --  0 MPI: without mpiexec, 1 MPI: with mpiexec
                             0x8 1x8 2x4 4x2   # total threads = 8
                         BACKEND cpu_sp cpu_dp gpu_sp gpu_dp )
```
Depending on the BACKEND you link to the correct ectrans libraries, on the condition it is being compiled. For SETUP_TRANS0 tests the BACKEND argument can then be empty, and then you link with ectrans_common only.
These arguments can of course be created as list variables but at least it will be a clear API without existing state.
If you prefer to keep BACKEND split into PLATFORM and PRECISION that is fine...

samhatfield:
Okay @wdeconinck I have made a few changes like you suggest - take a look :)

wdeconinck:
That looks very good.
Perhaps it would be nice to leave the checks like
`if (HAVE_OMP)` and `if (HAVE_MPI)` and `if (HAVE_dp)` and `if (HAVE_sp)` and `if (HAVE_CPU)` and `if (HAVE_GPU)` in the testsuite macro, and not add the tests that don't match the criteria there.

So we wouldn't need to assemble lists in particular if we only wanted to test `gpu_sp` with `MPI`, removing the need to guard with conditionals before adding the testsuite.

samhatfield:
Hmm not sure I understand. The reason I construct a list of backends / MPIxOMP configurations to test is so I can declare all the tests in a loop.

wdeconinck:
Instead of e.g.
```
set (ntasks 0)
if (HAVE_MPI)
  list (APPEND 1 2 4)
endif()
set (nthreads 1)
if (HAVE_OMP)
  list(APPEND nthreads 2 8)
endif()
foreach( mpi ${ntasks})
  foreach( omp ${nthreads})
     list( APPEND mpixomps ${mpi}x${omp}
  endforeach()
endforeach()
```
We could just define something more hand-coded / tuned
```
set (mpixomps 0x1 1x2 1x4 2x4 ... )
```
or still with a double loop, but if we don't have OMP or MPI then the testsuite generation macro will filter out the unsupported tests, so that only `0x1` would be generated.
Same idea for precisions and backends.

samhatfield:
Ahh I see. Okay I've made that change. The main tests/CMakeLists.txt could be tided a bit now (e.g. we still use lists of supported precisions for the other tests) but that can be done in another PR.

.github/workflows/build.yml
wdeconinck: Indeed, let us add capability for WILL_FAIL later.

wdeconinck:
You could move the variables mpixomps into the api/setup_trans0/CMakeLists.txt
I don't envision that we will need to test each api with all mpi or omp combination etc.

samhatfield:
Done.

wdeconinck (approved):
I think this is good to go! Thanks @samhatfield for addressing all suggestions.

PR #287, opened 2025-07-07
Dump checksums in ectrans tests

dhaumont:
## Summary
This PR adds checksum verification of fields in the benchmark program and in the tests suite.

## Description of changes

1. In ectrans-benchmark and ectrans-lam-benchmark:
    - a new argument `--dump-checksum [file_path]` has been added. When present, the benchmark will output two checksum files: one after dir_trans and one after inv_trans. The filename are constructed by appending '_dir_trans.checksums' and `_inv_trans_checksums` to the provided `file_path`.
    Warning:  The folder in file_path must be created beforehand.
    - the checksum files are in text format and are generated using the new function `dump_checksums`. They will contain the checksums for each level of each field, and for every iterations (see the example below).
    - the generation is performed on the first MPI rank, in which fields are gathered and aggregated before computing the checksums
   - the data at the end of the last block of the surface fields are set to zero before computing the checksums

```
! Remove trash at end of last block    
  iend = ngptot - nproma * (ngpblks - 1)      
  zgp2 (iend+1:, :, ngpblks) = 0
```

2. the unit tests using the benchmarks have been modified to add  `--dump-checksum`, the arguments provided beeing the name of the test

3. a new unit test `compare_checksums` has been added. This unit test uses internally the python script `src/test/compare_checksums.py` to iterate over all the checksums files and check that the files produced with different MPI and OMP settings are the same. 

```
cd ./build/ectrans/tests
ctest -R compare_checksums
```
It is also possible to run directly  to perform the comparisons by starting `compare_checksums.py` from the command line 

```
python3 ./test/compare_checksums.py ./build/ectrans/tests
```

## Testing  and validation 

The feature has been on tested on CPU on:
1. laptop (GNU compiler):  6 comparisons are failing (./ectrans_lam_test_benchmark_sp_48x40_mpi*_omp*_nfld0dir_trans.checksums)
2. atos (intel compiler): 10 comparisons are failing (ectrans_lam_test_benchmark_dp_48x40_mpi*_omp*_nfld0)
3. leonardo (nvhpc compiler): all the cpu comparions are OK

<del>No tests have been performed on GPU.</del>
UPDATE: The gpu test suits runs successfully on ATOS

## Example of checksum file

```
$ cat ectrans_lam_test_benchmark_sp_48x40_mpi0_omp4_nfld10dir_trans.checksums
 ====================
 iteration           1
 ====================
sp3d (1, 1) = 256EDE02F32A2AA3
[...]

PR #314, opened 2025-09-26
Introduction of field API interface in ectrans

dhaumont:
## Summary
This PR introduces a field API interface to ectrans. It is based on top of dump-checksums (PR https://github.com/ecmwf-ifs/ectrans/pull/287)

## Description of changes
- The new field API interface is implemented in `src/field_api`:
   * `src/field_api/trans/inv_trans_field_api.F90`: inverse transform (calling inv_trans internally)
   * `src/field_api/trans/dir_trans_field_api.F90`: direct transform (calling dir_trans internally)
   * `src/field_api/field_api_ectrans_mod.F90`: helper functions

- Optional compilation triggered by a new USE_FIELD_API configuration option. Adds a new dependency to field API.
- Single and double precision for CPU and GPU
- Accessible in ectrans-benchmark via a new option `--field-api`
   * `src/programs/util/ectrans_field_api_helper.F90` contains the field API implementation
- In the GPU version of ectrans-benchmark , setting `llacc = .true.` will force the intermediate fields on GPU. This requires additional memory copies host <-> device before and after the regular ectrans calls.
- New unit tests have been added to test `--field-api` (CPU and GPU)
- `compare_checksums.py` updated to compare the checksums with `field-api` with the regular version
- nvhpc22.11 on CPU: bit reproducibility between the field API and the regular version has been partially adressed by clamping small values in spectral space after dir_trans in ectrans benchmark (2a7aa8705667a05d60eaac3999b83f31cadf7792). This only applies to double precision (single precision is only bit reproducible during the first time step and slightly different after)

## Testing  and validation 

- ATOS:
  * Intel  2021.4.0 - CPU: bit reproducible
  * nvhpc 22.11 - CPU: dp : bit reproducible, sp: not reproducible
  * nvhpc 22.11 on GPU:  bit reproducible
- Ubuntu laptop
 *Gfortran 13.3.0 - CPU : bit reproducible

## Known limitations

- no LAM version 
- no openMP version ( OpenACC only)
- The GPU version with `llacc = .true. ` requires memory transfers between CPU and GPU, that should be avoided
- for nvhpc 22.11,  the CPU version in sp is not bit reproducible  ( first time step is bit reproducible, tiny differences after first iteration)

## Initial pull request
https://github.com/dhaumont/ectrans/pull/1
Contains some comments from @samhatfield that needs to be integrated

## Prototype in ARPEGE

See for instance this prototype of transinv_mdl_field_api.F90: https://github.com/dhaumont/IAL/blob/50t1_field_api_ectrans/arpifs/transform/transinv_mdl_field_api.F90
This version can be compiled using this bundle:
[50t1_ectrans_field_api_interface](https://github.com/dhaumont/IAL-bundle/tree/50t1_ectrans_field_api_interface/)


src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,480 @@
...
+    if (inum_sc_2d_fields > 0) then
+      call field_new(ywflds%spscalar, data=zspscalar(:,:))
+    endif
+    ioffset = 1
+
+    ! grid-point vector fields
+
+    ! gridpoint vector field vorticitity and divergence
+    if (lvordiv) then
+        call field_new(ywflds%vor, data=zgp(:, ioffset:ioffset+inum_wind_fields-1, :))
```
samhatfield, line 86: I think this might be an error. You're using `inum_wind_fields` to set the "fields" dimension of `zgp`, i.e. the total number of grid point fields. But `inum_wind_fields` is computed from the first dimension of `zspvor`. In spectral space the fields dimension is partitioned over the V set, so these two aren't the same.

wdeconinck (commented):
Hi Denis this is nice work!
I have not yet had the time to review it all but I already have a few comments on the integration in CMake.

Also I would not use the argument "LDACC" in the external outward facing API. It seems to indicate "OpenACC" but nothing should prevent to use "OpenMP target" with the same argument if the library was compiled as such. The ectrans_gpu targets do support OpenMP target already. How about another name, e.g. "LD_ON_DEVICE"?

But first, could you perhaps clarify what the intention of "LDACC" is, in case we have following situations:
1.  - library = ectrans_field_api_**cpu**_dp
     - fields are on host

2.  - library = ectrans_field_api_**gpu**_dp
     - fields are on host

3.  - library = ectrans_field_api_**cpu**_dp
     - fields are on device

4.  - library = ectrans_field_api_**gpu**_dp
     - fields are on device

5.  - library = ectrans_field_api_**cpu**_dp
     - some fields are on device and some fields are on host

6.  - library = ectrans_field_api_**gpu**_dp
     - some fields are on device and some fields are on host

Do we need such option as "LDACC" at all? As we can infer the gpu or cpu backend (where to compute) from the used library and if we need to update fields based on the presence of a up-to-date field on host/device from FIELD_API?

src/programs/ectrans-benchmark.F90
```diff
@@ -187,11 +206,15 @@ program ectrans_benchmark
 real(kind=jprb), allocatable :: global_field(:,:)
 
 !===================================================================================================
-
+#include "fspgl_intf.h"
```
samhatfield, line 209: Do we actually need this?

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,480 @@
...
+  class (field_3rb), pointer :: scalar2      ! grid-point surfacic scalar fields
+  class (field_3rb), pointer :: scalar2_ew   ! grid-point surfacic scalar fields derivatives ew
+  class (field_3rb), pointer :: scalar2_ns   ! grid-point surfacic scalar fields derivatives ns
+
+  class (field_3rb), pointer :: scalar   ! grid-point scalar fields as a single field
+  class (field_3rb), pointer :: scalar_ew! grid-point scalar fields derivatives ew
+  class (field_3rb), pointer :: scalar_ns! grid-point scalar fields derivatives ns
+end type wrapped_fields
+
+type fields_lists
```
samhatfield, line 38: This type seems very similar to `wrapped_fields`. Could you explain why we need both?

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,480 @@
...
+
+  ! Create field lists in ylf from field API objects in ywflds
+
+  type(wrapped_fields), intent(in) :: ywflds       !input fields api objects
+  type(fields_lists), intent(inout), target :: ylf ! output field lists
+  integer(kind=jpim), optional, intent(in) :: kvsetuv(:)     ! 'b-set' for vector fields
+  integer(kind=jpim), optional, intent(in) :: kvsetsc(:)     ! 'b-set' for scalar fields
+  integer(kind=jpim), optional, intent(in) :: kvsetsc2(:)    ! 'b-set' for surfacic fields
+
+  if(associated(ywflds%spvor)) ylf%spvor=[b(ywflds%spvor,'spvor',kvsetuv)]
```
samhatfield, line 225: Haven't seen this syntax before - is `b` an alias for something in FIELD API? Why do we pass `kvsetuv` too?

src/programs/ectrans-benchmark.F90
dhaumont: We need it to be able to use the optional argument fspgl_intf, which is an external routine called during inv_trans.
But as this optional routine is not tested in the benchmark, we don't need to include this file, you are right. 

Shouldn't we add the test of this routine in the benchmark (at least when it is used as a unit test)?

dhaumont:
@wdeconinck If I am not mistakem, only the options 1. and 2., are implemented  in the benchmark (fields are always created on the host, and then we execute either the CPU or the GPU backend).

As such, only these two configurations were tested, and all the other have not been tested yet.

src/programs/util/ectrans_field_api_helper.F90
dhaumont: OK, good idea. I will change this

CMakeLists.txt
```diff
@@ -181,12 +181,9 @@ ectrans_find_lapack()
 ecbuild_add_option( FEATURE FIELD_API
                     DEFAULT OFF
                     DESCRIPTION "Compile field API interface to ectrans"
+                    REQUIRED_PACKAGES "field_api"
```
wdeconinck, line 195: Indeed

src/programs/util/ectrans_field_api_helper.F90
dhaumont: renamed to `make_field`

src/programs/ectrans-benchmark.F90
dhaumont: just removed it

dhaumont:
UPDATE for LDACC:
@wdeconinck After a discussion with Philippe M., the specifications for `LDACC` is to allow the choice at runtime for ectrans to run on CPU or on GPU, whatever we do for the rest of the model. 
For the record, I copy-paste here the emails we exchanged about that

1. Denis: 
_Currently, ectrans_benchmark_gpu is statically linked to the GPU ectrans library and ectrans_benchmark_cpu is linked to the CPU version. What is missing to have a single executable, that would allow us to run ectrans on CPU or on GPU, with the choice of the backend at runtime. This was the intent of the variable LDACC in DIR_TRANS_FIELD_API and INV_TRANS_FIELD_API_

2. Willem:
_We would very much like to be able to choose the backend at runtime. This is not yet implemented but steps have been taken in this direction. In the beginning each of the implementations (CPU, GPU) and precisions (sp,dp) were overwriting the same symbol names. A parkind value JPRB was used to compile the chosen precision, and then the GPU version was just implementing the same symbols as the CPU version. Some time ago we did some work to make the symbols unique automatically via some sedrename preprocessor script, creating backend-specific with libraries and interfaces.
For backwards compatibility we then have an interface that makes the old symbols available via preprocessor defines, but then you cannot really mix-and-match backends.
What is missing is a final interface library independent of precision, and backend, linking to all of the 4 currently existing libraries and then dispatches to the unique symbol. e.g.: call  ectrans_invtrans( …, backend=…)
Probably the ‘KRESOL’  handle has to be setup specific to  {(GPU,CPU),(SP,DP)} pair because of specific backend specific data structures, in which case you cannot choose the backend at the invtrans callsite, but has to be chosen at `setup_trans()`
So it is not quite clear yet how this should look like.  Sam has also a draft PR which makes the GPU-version work on CPU which could also influence here._

3. Sam:
_I was just about to share an example program that calls either CPU or GPU depending on a logical. This program should be linked against trans_sp and trans_gpu_sp. Because of the preprocessing we do to avoid duplicate symbols, there will be no symbol clashes._

```
program backend_test

implicit none

#include "setup_trans0.h"
#include "setup_trans_gpu_sp.h"
#include "setup_trans_sp.h"

logical :: use_gpu = .true.

call setup_trans0(ldmpoff=.true.)

if (use_gpu) then
    call setup_trans_gpu_sp(ksmax=79, kdgl=160)
else
    call setup_trans_sp(ksmax=79, kdgl=160)
endif

end program backend_test
```
4. Denis: 
_Could inv_trans_field_api and dir_trans_field_api play the role of the missing common interface?_

5. Willem:
_I think that an interface that uses field_api derived types is still an alternative (and very attractive) API that could be implemented on top of described common API (which is using just array-pointers and some metadata), as  we would also like to make use of the common interface that can dispatch to different backend via the C-interface (transi).
But as mentioned it is not yet completely clear at this point how this new API should look like. Simultaneously we’re trying to see how we can remove other restrictions currently present in ectrans, such as removing the restriction of a fixed MPI communicator and domain decomposition configuration for all transforms within the same scope. E.g one question is: “Do we need a setup_trans0 at all in that case???” Those are some questions that we need to explore before nailing down a new API like that_

src/field_api/CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+          MODULE_DIRECTORY  ${CMAKE_BINARY_DIR}/module/ectrans
+          INSTALL_DIRECTORY module/ectrans
+        )
+
+      if (${platform} MATCHES "gpu")
+          if( CMAKE_Fortran_COMPILER_ID MATCHES "NVHPC" )
+             set_source_files_properties(  ${ectrans_field_api_${platform_tag}${prec}_src} PROPERTIES COMPILE_OPTIONS "-acc" )
+             #     set_source_files_properties(  ${ectrans_field_api_${platform_tag}${prec}_src} PROPERTIES COMPILE_OPTIONS "-O1" )
+          endif()
+      endif()
```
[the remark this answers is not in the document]
dhaumont: @wdeconinck 
Not sure what to do here. I just removed the instructions from L109 to 114, which had nothing to do with linkin,. they were adding the compilation options `-acc -o1` to the field_api interface file.
I did not add the `target_link_libraries`, and it seems that everything compiles fine without.
Is it mandatory ?

src/field_api/CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+ 
+      generate_backend_sources(BACKEND ${platform_tag}${prec}
+                               OUTPUT ectrans_field_api_${platform_tag}${prec}_src
+                               DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/generated/ectrans_field_api_${platform_tag}${prec})
+
+      ecbuild_add_library(
+          TARGET           ectrans_field_api_${platform_tag}${prec}
+          LINKER_LANGUAGE  Fortran
+          SOURCES          ${ectrans_field_api_${platform_tag}${prec}_src}
+          PUBLIC_LIBS      ectrans_${prec} field_api_${prec} fiat parkind_${prec}
```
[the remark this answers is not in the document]
dhaumont: I just changed it, thanks for catching this.
Not sure what was the behaviour before, I was probably linking all the time with the CPU version, don't know how this was working.

src/programs/ectrans-benchmark.F90
samhatfield: Good idea, I should add a test for `FSPGL_PROC` functionality when I eventually add specific testing for `INV_TRANS` (waiting for PR #313 first).

samhatfield:
FYI @dhaumont it looks like you're still rewriting the calls to `dump_checksum`, but just in case you missed it, we just changed the signature in a [recent PR](https://github.com/ecmwf-ifs/ectrans/pull/328/files). Only small changes required.

dhaumont:
> FYI @dhaumont it looks like you're still rewriting the calls to `dump_checksum`, but just in case you missed it, we just changed the signature in a [recent PR](https://github.com/ecmwf-ifs/ectrans/pull/328/files). Only small changes required.

No, my changes come from resolving conflicts with the changes made in #328. Do you think I broke anything ?  In any case I am staring to test everything again, to get this PR out of the draft state.

src/programs/util/ectrans_field_api_helper.F90
dhaumont: I just added some documentation in the source code to clarify the situation: 008ef1d68b0882106f543468d57f04141bd00cda

samhatfield:
> No, my changes come from resolving conflicts with the changes made in #328. Do you think I broke anything ? In any case I am staring to test everything again, to get this PR out of the draft state.

You need to update calls to `dump_checksum`, e.g. instead of 

```fortran
call dump_checksums(filename=checksums_filename, noutdump=noutdump, &
                  & jstep=jstep, myproc=myproc, nproma=nproma, ngptotg=ngptotg, &
                  & ivset=ivset, ivsetsc=ivsetsc, ivsetsc2=ivsetsc2, &
                  & nspec2g=nspec2g, zgp=zgp)
```
it should be
```fortran
call dump_checksums(filename=checksums_filename, noutdump=noutdump, &
                  & jstep=jstep, myproc=myproc, nproma=nproma, ngptotg=ngptotg, zgp=zgp)
```
You don't need to provide spectral-related arrays (`ivset`, `nspec2g` etc.) when calculating checksums of grid point arrays, and vice versa you don't need to provide gridpoint-related arrays (`nproma`, `ngptotg`) when calculating for spectral arrays.

dhaumont:
> You don't need to provide spectral-related arrays (`ivset`, `nspec2g` etc.) when calculating checksums of grid point arrays, and vice versa you don't need to provide gridpoint-related arrays (`nproma`, `ngptotg`) when calculating for spectral arrays.

Thanks Sam, indeed those changes escaped my notice. I just put back your change, and indeed, the code is a bit clearer that way
41fd1f40ad3c4205e50376765aefbfcec39872a4

src/trans/include/ectrans/field_basic_type_ptr.h
wdeconinck, line 1: I am not sure I see the point of this file.
Why can you not use the module directly?
Also for `field_api_ectrans.h`

src/field_api/CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+                           $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src/field_api/include>
+                           )
+
+      ectrans_target_fortran_module_directory(
+          TARGET            ectrans_field_api_${platform_tag}${prec}
+          MODULE_DIRECTORY  ${CMAKE_BINARY_DIR}/module/ectrans
+          INSTALL_DIRECTORY module/ectrans
+        )
+
+      if (${platform} MATCHES "gpu")
```
dhaumont: @wdeconinck  Are these lines to add -acc compilation OK,  or should I proceed in another way?

src/field_api/CMakeLists.txt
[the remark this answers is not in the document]
dhaumont: UPDATE: I had to put them back, otherwise the files in the field_api folder were not compiled with "-acc"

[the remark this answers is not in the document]
wdeconinck: Normally `target_link_libraries( ectrans_field_api_${platform_tag}${prec} PRIVATE OpenACC::OpenACC_Fortran )` should do a number of things given `OpenACC::OpenACC_Fortran` is a cmake target:
- compile the files with the -acc flag and others possibly when defined in the toolchain as "OpenACC_Fortran_FLAGS"
- add link flags or link libraries
- add include directories defined by the target

Because it is marked as PRIVATE it does not propagate the compile flags to downstream, which is good. But then when downstream also needs those flags then that code is responsible to link with OpenACC itself.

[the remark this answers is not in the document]
dhaumont: @wdeconinck Thanks for the explanation. It is fixed now

src/trans/include/ectrans/field_basic_type_ptr.h
dhaumont: I had to introduce them to compile in sp and dp precision.
For instance, the file  `src/trans/include/ectrans/field_basic_type_ptr.h`:

```
! Header to manage single and double precision
USE FIELD_API_BASIC_TYPE_MOD, ONLY: FIELD_BASIC_PTR
```
will be transformed into  `./build/include/ectrans/trans_dp/field_basic_type_ptr.h`:
```
! Automatically generated interface header for backward compatibility of generic symbols !
#if defined(field_basic_type_ptr)
#undef field_basic_type_ptr
#endif
#if defined(FIELD_BASIC_TYPE_PTR)
#undef FIELD_BASIC_TYPE_PTR
#endif
#include "field_basic_type_ptr_sp.h"
#define field_basic_type_ptr FIELD_BASIC_TYPE_PTR_sp
#define FIELD_BASIC_TYPE_PTR FIELD_BASIC_TYPE_PTR_sp

```

dhaumont:
@wdeconinck @samhatfield I finished testing and implementing the requested changes for this PR. For me it is no longer a draft, and can be considered for merging.
Please tell me if something else requires some improvements. Please also close the comments which you are happy with, so that we have a better status. Thanks

tests/CMakeLists.txt
```diff
@@ -325,6 +325,102 @@ foreach( benchmark ${benchmarks} )
...
+            MPI ${mpi}
+            OMP ${omp}
+        )
+        ectrans_set_test_properties( ${base_title}_nfld10_nlev20_flt )
+      endif()
+
+  endforeach()
+    endforeach()
+  endforeach()
+endforeach()
```
samhatfield: This basically duplicates the loop above, from lines 219 to 326, only with Field API enabled. Why not modify the above loop rather than creating a new one?

dhaumont: They are similar but not identical.
My concern was the amount of tests to run.
Given the number of available options (precision, use of field api,  call mode, MPI tasks, etc), it seems impossible to perform all the possible combinations.
That's why I did'nt want to double the number of tests by systematically running them with the field API interface. Having a separate routine allows us to select the subset of tests to run with Field API.

src/trans/include/fspgl_intf.h
samhatfield, line 1: Looks like you're trying to solve a similar problem to PR #337 ? Is this required for Field API functionality or can we review it in a separate PR, to simplify this one?

src/trans/include/ectrans/field_api_ectrans.h
samhatfield, line 1: Is this file needed? I don't think it's used anywhere?

src/trans/include/ectrans/field_basic_type_ptr.h
samhatfield: Same comment as above and as Willem - what's the intended use for these two header files? They aren't used in ectrans-benchmark for example.

samhatfield:
Hey @dhaumont, thanks for your patience with this. Big PR so we want to take our time. As with other PRs is it okay if I open a PR to field_api_interface which removes whitespace changes? This will make it easier to look at the true differences. E.g. this: https://github.com/ecmwf-ifs/ectrans/pull/314/changes#diff-bfd9d146dfab7708c079d896a09f7c52c70fa17572bd2c12fddaf6316ff93427R1804-R1812

src/trans/include/fspgl_intf.h
dhaumont: Indeed, this was our attempt to improve the situation regarding the call to the external routine fspgl_intf. There is nothing specific to Field API functionality, therefore we can keep the implementation from PR #337 which is better than what was proposed here.
How do you suggest we proceed in practice ?

src/trans/include/ectrans/field_api_ectrans.h
dhaumont: It is needed to manage sp and dp compilation. It is included in `field_api_helper.F90`

dhaumont:
> Hey @dhaumont, thanks for your patience with this. Big PR so we want to take our time. 

No problem at all, I really appreciate all the effort you've put into reviewing it and putting it in good shape. Since it is a new interface to the outside world, it's important that it is clean from the start.

> As with other PRs is it okay if I open a PR to field_api_interface which removes whitespace changes? This will make it easier to look at the true differences. E.g. this: https://github.com/ecmwf-ifs/ectrans/pull/314/changes#diff-bfd9d146dfab7708c079d896a09f7c52c70fa17572bd2c12fddaf6316ff93427R1804-R1812

Yes, of course. Sorry if I messed up with whitespaces, I did my best to limit the changes, but feel free to correct the indentation as needed.

src/trans/include/ectrans/field_basic_type_ptr.h
dhaumont: These headers are necessary to manage the compilation of sp and dp, and are included in `./src/programs/util/ectrans_field_api_helper.F90`.

I simply replicated what had been done in the other header files. But perhaps I've missed something?

src/trans/include/fspgl_intf.h
samhatfield: You can probably tell from the discussion in #337 that I lost motivation... What we are doing with `FSPGL_INT` is obviously not ideal, and currently prevents one compiler (NAG) from working with ecTrans. On the other hand, changing the arguments of `INV_TRANS` would break compatibility with arpifs and probably trigger a lot of versioning headaches in the transition period, with some branches of ifs-source working with ecTrans, others not. This is unfortunately what happens when you change the API.

If we are introducing new external routines (dir_trans_field_api / inv_trans_field_api) then we can use the modern approach (`TYPE(FSPGL_INTF)`) with those and leave the others untouched, for now. So I guess we can keep `FSPGL_INTF` in this PR.

Having said that, shouldn't fspgl_intf.h go in src/trans/include/ectrans? It's currently in src/trans/include.

src/programs/ectrans-benchmark.F90
```diff
@@ -1302,9 +1395,10 @@ subroutine get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, n
       case('--nprtrw'); nprtrw = get_int_value('--nprtrw', iarg)
       case('-c', '--check'); ncheck = get_int_value('-c', iarg)
       case('--no-pinning'); lpinning = .False.
+      case('--field-api'); lfield_api = .True.
       case('--callmode')
           icall_mode = get_int_value('--callmode', iarg)
-          if (icall_mode /= 1 .and. icall_mode /= 2) then
+          if (icall_mode < 1 .or. icall_mode > 2) then
```
samhatfield, line 1380: Interesting - any reason you made this change?

src/trans/include/ectrans/field_basic_type_ptr.h
samhatfield: I see what you mean. It's okay, you followed the right approach for introducing a new include file. This is a new situation for us - before, we have not needed to make a derived type externally accessible in order to use one of the ecTrans API subroutines. `DIR/INV_TRANS_FIELD_API` are the first. I need to think about this a bit.

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,401 @@
...
+    integer, intent(in) :: k2d
+
+    integer :: ioffset
+
+    ! spectral vector fields
+    call field_new(ywflds%spvor,  data=zspvor(:,:))
+    call field_new(ywflds%spdiv,  data=zspdiv(:,:))
+
+    ! spectral scalar fields
+    if (k2d > 0) then
```
samhatfield, line 94: I think `k2d` is always greater than 0, because we always have one 2D scalar field, so maybe you could just delete these if statements?

src/programs/util/ectrans_field_api_helper.F90
samhatfield: Okay, I don't fully understand but I trust you. I think one thing that's confusing me is why you need a "list" of e.g. vorticities. Isn't there only one vorticity variable in IFS? In the benchmark program the list has one element. In what situations will it NOT have just one element?

src/programs/ectrans-benchmark.F90
dhaumont: It is a remnant of a temporary merge implementation where used `icall_mode = 3` and `icall_mode = 4` to trigger the field API interface.
Since `icall_mode` was an integer and not a boolean, I decided to keep the more generic test, compatible with other values of icall_mode, however we can revert this change if you wish.

src/programs/util/ectrans_field_api_helper.F90
dhaumont: The new interface will remove the limitation requiring at least one 2D scalar field for the spectral transforms. I would prefer to keep this test, so that all boundary cases can be tested in the future (providing only 3d fields, only 2d fields, no fields, etc)

src/trans/include/fspgl_intf.h
dhaumont: OK, so we keep it and I'll move it to `src/trans/include/ectrans`

src/programs/util/ectrans_field_api_helper.F90
dhaumont: Good question. It was also confusing for me in the begining. 
You are right, there's only one vorticity for now, so a single element is enough.
However, there are good reasons to provide the vorticity as a list:
- to have a uniform interface for all  fields: the user of the interface always provides them as a list
- by definition, vorticity is the curl of velocity. Providing a list allows calculating the curl of any vector fields, which could be useful in the future
-  in IFS and ARPEGE, we have many duplicate calls to inv_trans and dir_trans. Thanks to the field API interface, our hope is to be able to replace them by a single call. In many cases, this is only possible by using lists
- we don't have it now, but think about having relative velocity added, or transforming the vorticity at different time steps in a single ecTrans call.

dhaumont:
@samhatfield @wdeconinck The unit tests are failing because the build doesn't have `ENABLE_FIELD_API=ON`.
This commit will hopefully fix this: 88af96bc20932a3df5b73f024f252f2847f47bad
Can you trigger a new build ? Thanks

dhaumont:
If you activate `ENABLE_FIELD_API=ON`, compilation will succeed with Ninja but will fail with Make.
- Ninja: compilation OK
```
./ial-bundle build  --arch ./arch/ecmwf/hpc2020  --ninja -j32
```
- Make: compilation failed:
```
./ial-bundle build  --arch ./arch/ecmwf/hpc2020  -j32

[ 80%] Generating /lus/h2resw01/scratch/cvap/ectrans_bundle_field_api_interface/build/ectrans/src/field_api/generated/ectrans_field_api_dp/trans/inv_trans_field_api.F90
/bin/sh: /lus/h2resw01/scratch/cvap/ectrans_bundle_field_api_interface/build/ectrans/src/field_api/generated/ectrans_field_api_dp/trans/inv_trans_field_api.F90: No such file or directory
make[2]: *** [ectrans/src/field_api/CMakeFiles/ectrans_field_api_dp.dir/build.make:85: ectrans/src/field_api/generated/ectrans_field_api_dp/trans/inv_trans_field_api.F90] Error 1
make[1]: *** [CMakeFiles/Makefile2:10278: ect
```

The reason is that the folder `generated/ectrans_field_api_dp` is not created with Make when we generate the field API files for different precisions:
https://github.com/dhaumont/ectrans/blob/88af96bc20932a3df5b73f024f252f2847f47bad/src/field_api/CMakeLists.txt#L66

Thanks @chesnelj for reporting this bug.

src/programs/ectrans-benchmark.F90
samhatfield: The two tests are identical in integer arithmetic, so not a problem.

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,401 @@
...
+    endif
+
+end subroutine wrap_benchmark_fields
+
+subroutine create_fields_lists(ywflds,ylf, kvsetuv, kvsetsc,kvsetsc2)
+
+  ! Create field lists in ylf from field API objects in ywflds
+
+  type(wrapped_fields), intent(in) :: ywflds       !input fields api objects
+  type(fields_lists), intent(inout), target :: ylf ! output field lists
```
samhatfield:
```suggestion
  type(fields_lists), intent(inout) :: ylf ! output field lists
```
It's probably harmless, but do we actually need this property?

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,401 @@
...
+  integer(kind=jpim), optional, intent(in) :: kvsetuv(:)     ! 'b-set' for vector fields
+  integer(kind=jpim), optional, intent(in) :: kvsetsc(:)     ! 'b-set' for scalar fields
+  integer(kind=jpim), optional, intent(in) :: kvsetsc2(:)    ! 'b-set' for surfacic fields
+
+  if(associated(ywflds%spvor)) ylf%spvor=[make_field(ywflds%spvor,'spvor',kvsetuv)]
```
samhatfield: Does this operation automatically allocate memory for `spvor`?

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -0,0 +1,401 @@
...
+  nullify(ywflds%scalar3_ns)
+
+  nullify(ywflds%scalar2)
+  nullify(ywflds%scalar2_ew)
+  nullify(ywflds%scalar2_ns)
+
+  nullify(ywflds%scalar)
+  nullify(ywflds%scalar_ew)
+  nullify(ywflds%scalar_ns)
+end subroutine nullify_wrapped_fields
```
samhatfield: I think you can just declare all pointer members of `wrapped_fields` with `=> null()`. This is a common pattern for initialising pointers in Fortran and would save you some lines of code here.

src/programs/CMakeLists.txt
```diff
@@ -24,10 +28,21 @@ foreach( program ectrans-benchmark )
                                   fiat
                                   parkind_${prec}
                                   trans_${prec}
+                                  $<${HAVE_FIELD_API}:ectrans_field_api_${prec}>
+                                  $<$<AND:${HAVE_FIELD_API},${HAVE_ACC}>:OpenACC::OpenACC_Fortran>
```
samhatfield: Is this necessary?

src/programs/util/ectrans_field_api_helper.F90
dhaumont: Yes, it does. 
"make_field" allocates an array and fill it with the pointers of the fields received.
It runs in two passes: the first determines the size of the array, then the array is filled.

src/field_api/ectrans_field_api_mod.F90
```diff
@@ -0,0 +1,478 @@
...
+      LS3RBV(JCOUNT)%NAME=TRIM(NAME)//TRIM(CLOUT)
+      LS3RBV(JCOUNT)%IVSET = IVSET(JLEV)
+      JCOUNT = JCOUNT + 1
+
+    ENDDO
+  ENDDO
+
+END SUBROUTINE LS3RB
+
+FUNCTION LS(LSV,YLFL,LDACC,LDRDONLY) RESULT(IOFF)
```
samhatfield: This function (I think) calculates the total number of 2D fields stored in an input `FIELD_BASIC_PTR` and returns it in `IOFF`. However it also does something else in a (currently) dead code path which I don't understand... `LSV` is populated with values, but its intent isn't declared (`INOUT` or otherwise) so I'm not sure what the intended purpose is. Is this code truly dead or did you leave it behind for a future purpose?

src/field_api/dir_trans_field_api.F90
```diff
@@ -0,0 +1,396 @@
...
+ENDIF
+! 4. Copy back temporary array data into spectral fields
+
+! copy spectral vorticity and divergence
+IF (IUVG>0) THEN
+    IF (LDACC) THEN
+    !$ACC UPDATE SELF(ZPSPVOR,ZPSPDIV)
+    ENDIF
+
+    C = LS(YLSPVVOR, YDFSPVOR, LDACC, .FALSE.)
```
samhatfield: Possible answer to my question below here...

src/field_api/ectrans_field_api_mod.F90
samhatfield, line 1: Could you add intents to all arguments of subroutines and functions in this module?

src/field_api/inv_trans_field_api.F90
```diff
@@ -0,0 +1,602 @@
...
+
+! b-set for inv-trans
+INTEGER(KIND=JPIM),ALLOCATABLE :: IVSETUV(:)
+INTEGER(KIND=JPIM),ALLOCATABLE :: IVSETSC2(:)
+
+INTEGER(KIND=JPIM)          :: ISPUV
+INTEGER(KIND=JPIM)          :: IFLDXG
+INTEGER(KIND=JPIM)          :: IFLDXL
+INTEGER(KIND=JPIM)          :: IFLDXGUV
+INTEGER(KIND=JPIM)          :: IFLDXLUV
```
samhatfield:
```suggestion
```
These aren't used

src/field_api/inv_trans_field_api.F90
```diff
@@ -0,0 +1,602 @@
...
+#include "abor1.intfb.h"
+
+!     ------------------------------------------------------------------
+IF (LHOOK) CALL DR_HOOK('INV_TRANS_FIELD_API',0,ZHOOK_HANDLE)
+
+ISPUV = 0
+IFLDXG= 0
+IFLDXL= 0
+IFLDXGUV= 0
+IFLDXLUV= 0
```
samhatfield: Same as above

src/field_api/inv_trans_field_api.F90
```diff
@@ -0,0 +1,602 @@
...
+LOGICAL                     :: LLUVDER                                ! INDICATING IF E-W DERIVATIVES OF U AND V ARE REQ.
+REAL(KIND=JPHOOK)           :: ZHOOK_HANDLE
+
+#include "inv_trans.h"
+#include "abor1.intfb.h"
+
+!     ------------------------------------------------------------------
+IF (LHOOK) CALL DR_HOOK('INV_TRANS_FIELD_API',0,ZHOOK_HANDLE)
+
+ISPUV = 0
```
samhatfield: `ISPUV` also doesn't seem to be used - can you check?

src/field_api/inv_trans_field_api.F90
```diff
@@ -0,0 +1,602 @@
...
+
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFSPVOR(:), YDFSPDIV(:)        ! SPECTRAL VECTOR FIELDS : VORTICITY AND DIVERGENCE FIELDS (IN)
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFSPSCALAR(:)                  ! SPECTRAL SCALAR FIELDS (IN)
+
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFU(:),YDFV(:)                 ! GRID VECTOR FIELDS     (OUT)
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFVOR(:),YDFDIV(:)             ! GRID VECTOR FIELDS :VORTICITY AND DIVERGENCE     (OUT)
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFSCALAR(:)                    ! GRID SCALAR FIELDS     (OUT)
+
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFU_EW(:),YDFV_EW(:)             ! GRID VECTOR FIELDS DERIVATIVES EW (OUT)
+TYPE(FIELD_BASIC_PTR),INTENT(IN), OPTIONAL  :: YDFSCALAR_NS(:), YDFSCALAR_EW(:)  ! GRID SCALAR FIELDS DERIVATIVES EW AND NS (OUT)
```
samhatfield: Are these intents correct?

src/field_api/ectrans_field_api_mod.F90
```diff
@@ -0,0 +1,478 @@
...
+USE FIELD_BASIC_MODULE, ONLY: FIELD_BASIC
+USE FIELD_MODULE, ONLY:FIELD_1RB, FIELD_2RB, FIELD_3RB, FIELD_4RB, FIELD_2RB_VIEW, FIELD_3RB_VIEW
+USE FIELD_API_BASIC_TYPE_MOD, ONLY: FIELD_BASIC_PTR
+IMPLICIT NONE
+
+TYPE SPEC_VIEW
+  ! Spectral field view
+  REAL(KIND=JPRB),POINTER :: P(:)
+  INTEGER               :: IVSET
+  CHARACTER(LEN=12)     :: NAME
```
samhatfield, line 40: I couldn't find a case where `NAME` is used. Could you check this is still required?

src/field_api/ectrans_field_api_mod.F90
```diff
@@ -0,0 +1,478 @@
...
+END TYPE
+
+TYPE GRID_VIEW
+! Grid point field view
+  REAL(KIND=JPRB),POINTER :: P(:,:)
+  INTEGER               :: IVSET
+  CHARACTER(LEN=12)     :: NAME
```
samhatfield, line 47: Same here.

src/programs/util/ectrans_field_api_helper.F90
samhatfield: My question wasn't so much about `make_field` but rather whether it's legal to assign an array literal to an unallocated array (`spvor` in this case). The pattern is essentially this:
```fortran
real, allocatable :: spvor(:)
spvor = [1.0, 2.0, 3.0]
```
Normally you would have to allocate `spvor` before assigning values. Apparently [it is not necessary since Fortran 2003](https://stackoverflow.com/a/42141701).

wdeconinck:
I took the liberty to commit directly in your branch @dhaumont, I hope you don't mind.
I fixed the CI by disabling the tests that require field_api when the feature FIELD_API is not enabled.

wdeconinck:
It is not possible that a downstream package would be linking to multiple versions of ectrans_field_api (i.e. ectrans_field_api_cpu_sp, ectrans_field_api_cpu_dp, ectrans_field_api_gpu_sp, ectrans_field_api_gpu_dp).
This is because you would have each ectrans_field_api_cpu_dp and ectrans_field_api_cpu_sp etc, each using `FIELD_MODULE` from field_api but this module and its types/symbols are not uniquely defined (once for field_api_sp and once for field_dp).
This is not a problem because its use is intended for IAL/IFS, which would link to only one of these versions.

However this means that we don't need to go through the same procedure as done with trans to define unique symbols, as the uniqueness is not important. For trans this was important because we want to be able to make use of trans through downstream libraries (like Atlas) that may make a runtime choice on precision and cpu/gpu execution.

If you don't mind I will refactor this a bit, reducing the complexity also with the includes which would not be required any longer.

dhaumont:
> I took the liberty to commit directly in your branch @dhaumont, I hope you don't mind. I fixed the CI by disabling the tests that require field_api when the feature FIELD_API is not enabled.

So the field api interface is not compiled nor executed in the CI ????
In my opinion, what is not tested is just dead code.

dhaumont:
> It is not possible that a downstream package would be linking to multiple versions of ectrans_field_api (i.e. ectrans_field_api_cpu_sp, ectrans_field_api_cpu_dp, ectrans_field_api_gpu_sp, ectrans_field_api_gpu_dp). This is because you would have each ectrans_field_api_cpu_dp and ectrans_field_api_cpu_sp etc, each using `FIELD_MODULE` from field_api but this module and its types/symbols are not uniquely defined (once for field_api_sp and once for field_dp). This is not a problem because its use is intended for IAL/IFS, which would link to only one of these versions.
> 
> However this means that we don't need to go through the same procedure as done with trans to define unique symbols, as the uniqueness is not important. For trans this was important because we want to be able to make use of trans through downstream libraries (like Atlas) that may make a runtime choice on precision and cpu/gpu execution.
> 
> If you don't mind I will refactor this a bit, reducing the complexity also with the includes which would not be required any longer.

One of the main requirement of this interface is the ability to dynamically choose betweeen CPU and GPU execution: you want to be able to run the spectral transform on CPU or on GPU with LDACC.
Is this compatible with what you suggest ?

I'm not sure what problem you are trying to solve exactly, is this the fact that you have to include an interface file ?

src/field_api/ectrans_field_api_mod.F90
dhaumont: Names are really useful for debugging, I would prefer to keep them

src/field_api/inv_trans_field_api.F90
dhaumont: Those are a list of pointers to some arrays of data (ie field data). 
In this function, we only modify the data, and don't touch the pointers, which are const  and read only. So the pointers themselves are INTENT(IN).
We could change them to INTENT(INOUT) if you really want to convey the information to the user of  the interface that you will modify the data. But this will have the side effect to disable some compiler optimizations that can be applied with const arguments

src/field_api/inv_trans_field_api.F90
dhaumont: correct. I am removing it.

src/field_api/inv_trans_field_api.F90
dhaumont: good catch. they will go away

src/field_api/inv_trans_field_api.F90
samhatfield: No that's a fair point. I was trying to understand the signature of the subroutine at a high level and my brain expected to see "input variables" and "output variables" as with INV_TRANS. Fortran's intents kind of stop making sense when we're dealing with pointers.

samhatfield:
> So the field api interface is not compiled nor executed in the CI ???? In my opinion, what is not tested is just dead code.

It's tested in the "HPC" suite (ECMWF HPC2020 and LUMI) but not the Github suite. So what we need is a commit like https://github.com/ecmwf-ifs/ectrans/pull/314/changes/88af96bc20932a3df5b73f024f252f2847f47bad but applied to .github/workflows/build.yml.

wdeconinck:
> So the field api interface is not compiled nor executed in the CI ???? In my opinion, what is not tested is just dead code.

The FIELD_API feature is optional, meaning that when FIELD_API is not enabled those particular tests should not be running, we can all agree on that. That is what I fixed.
Next, I definitely agree that the CI should be running also with FIELD_API feature enabled.

I forgot to mention that for the build-hpc workflow I also added the missing field_api dependency in commit [ce5ae51](https://github.com/ecmwf-ifs/ectrans/pull/314/commits/ce5ae51798a22cbe7cf4b5e9213d77ac4524fc22) . It was not sufficient to set "-DENABLE_FIELD_API=ON", so hopefully the CI is going to test this at least on the HPCs now (lumi and ecmwf-atos).
Looking at the logs however it seems to ignore changes to the Github-Actions workflows for pull-requests with "secrets". I guess that is a "feature" of Github-Actions for security reasons. We would need to first change the workflow in the develop branch then to make sure this PR is testing FIELD_API changes.

src/field_api/ectrans_field_api_mod.F90
dhaumont: The function `LS` computes the size of the array `LSV` (pass 1) and fills `LSV` with pointers to field API pointers. The array `LSV` has to be pre-allocated somewhere else.

From dir_trans_field_api and inv_trans_field_api: 

1. We first call `LS_COUNT`, which calls `LS` with an empty dummy list for `LSV`. This will only execute the first pass of 'LS' and returns the length of the array `IOFF`.
2. We initialize the array of pointers `LSV` with the correct size 
3. We call again `LS` with the proper `LSV` array. Then the two passes are executed, the first one count again the size, and the second one will populate the array.

The fact that we have two passes in 1. and in 3. is coming from the first implementation, where the array `LSV` was allocated directly in `LS`. But unfortunately there was a compiler bug with NVIDIA, so I had to preallocate the array, and implement this `LS_COUNT` function to find the size of the array.
In the current state, the first pass is not strictly needed anymore during 3., it was kept to ensure that the size of `LSV` was correct before filling the array of pointers.

This design is a bit complicated, we could try to simplify this further.

src/field_api/dir_trans_field_api.F90
dhaumont: Yes, we call LS to fill the array. 
In this case, `LSV` should be `INTENT(OUT)` in `LS`, since we are filling the array with the field API pointer values

dhaumont:
> > So the field api interface is not compiled nor executed in the CI ???? In my opinion, what is not tested is just dead code.
> 
> The FIELD_API feature is optional, meaning that when FIELD_API is not enabled those particular tests should not be running, we can all agree on that. That is what I fixed. Next, I definitely agree that the CI should be running also with FIELD_API feature enabled.
> 
> I forgot to mention that for the build-hpc workflow I also added the missing field_api dependency in commit [ce5ae51](https://github.com/ecmwf-ifs/ectrans/pull/314/commits/ce5ae51798a22cbe7cf4b5e9213d77ac4524fc22) . It was not sufficient to set "-DENABLE_FIELD_API=ON", so hopefully the CI is going to test this at least on the HPCs now (lumi and ecmwf-atos). Looking at the logs however it seems to ignore changes to the Github-Actions workflows for pull-requests with "secrets". I guess that is a "feature" of Github-Actions for security reasons. We would need to first change the workflow in the develop branch then to make sure this PR is testing FIELD_API changes.

OK, thanks a lot Willem.

wdeconinck:
> One of the main requirement of this interface is the ability to dynamically choose betweeen CPU and GPU execution: you want to be able to run the spectral transform on CPU or on GPU with LDACC. Is this compatible with what you suggest ?
> 
> I'm not sure what problem you are trying to solve exactly, is this the fact that you have to include an interface file ?

I think the simplicification will still allow us to develop later on the ability to choose between CPU and GPU backends, but not between single and double precision, due to linking to two field_api libraries that are incompatible with each other, i.e. the meanings of FIELD_1RB, FIELD_2RB etc.

What will be possible in the future is to have two libraries : `ectrans_field_api_${prec}`  (`${prec}=(sp,dp)`) which will link with both `ectrans_cpu_${prec}` and `ectrans_gpu_${prec}`.
Based on the `LDACC` flag you would then need to call `invtrans_cpu_${prec}` or `invtrans_gpu_${prec}` within `invtrans_field_api(KRESOL=NRESOL, ..., LDACC=...)`

It is however not yet clear to me if `NRESOL` is in this case the same value for `LDACC=.TRUE.` or `LDACC=.FALSE.`
We are not there yet. Perhaps the GPU or CPU backend will be embedded in the `KRESOL`, so that we don't need the LDACC flag:
```
call invtrans_field_api(KRESOL=NRESOL_GPU, ...)
call invtrans_field_api(KRESOL=NRESOL_CPU, ...)
```
and then introspection on the KRESOL parameter will need to be used instead of the LDACC flag.

What is clear to me however is that you also did not intend to create these symbols: `invtrans_field_api_(cpu,gpu)_(sp,dp)`.

wdeconinck:
See https://github.com/dhaumont/ectrans/pull/4 for the proposed changes to remove the unique symbol handling with the sedrenames etc.

src/field_api/ectrans_field_api_mod.F90
dhaumont: Done in 059309a76d31ace44e33d49b6fe63fe802a1b007

src/field_api/inv_trans_field_api.F90
dhaumont: done in ecbf2d24629f52783225bf6545235153d98435e7

src/field_api/inv_trans_field_api.F90
dhaumont: done in ecbf2d24629f52783225bf6545235153d98435e7

src/field_api/inv_trans_field_api.F90
dhaumont: done in 7cc791248d8f628290538bc1c1b5162ea879384e

dhaumont:
@samhatfield @wdeconinck After our discussion about different passes in LS and LG that were not needed anymore, I simplified the code by making the rguments optional. I also removed the passes.

c7af7eaba8618577aef6fb2397220b3994d61d50

IMPORTANT: I just found a crash in the unit test due to this commit. I'm inverstigating

src/field_api/ectrans_field_api_mod.F90
wdeconinck: I agree with @dhaumont , this is useful for potential logging.

dhaumont:
> IMPORTANT: I just found a crash in the unit test due to this commit. I'm inverstigating

The crash has been fixed soon after its discovery: fd0ef29bbd62d3452015f18b0f08d2d67d0380f1

dhaumont:
@samhatfield @wdeconinck 
I just reordered the arguments and remove the ones that could be inferred from array dimensions:
https://github.com/dhaumont/ectrans/pull/5
Could you review it before I merge it here ? Thanks

wdeconinck:
@samhatfield I am trying to fix the nvhpc-26.3 build in [another branch](https://github.com/ecmwf-ifs/ectrans/tree/update-ci-nvhpc-26.3). I already had to fix the install-nvhpc.sh script in Atlas and other packages as well to work for nvhpc 26 onwards. A download URL has changed.

samhatfield:
> @samhatfield I am trying to fix the nvhpc-26.3 build in [another branch](https://github.com/ecmwf-ifs/ectrans/tree/update-ci-nvhpc-26.3). I already had to fix the install-nvhpc.sh script in Atlas and other packages as well to work for nvhpc 26 onwards. A download URL has changed.

Thanks for taking that on. I had a feeling it was a URL issue when I stopped work on Friday.

samhatfield:
So, nvhpc/25.1 has an issue with the `SELECT TYPE` statement. We cannot update all the way to 26.3 for reasons [outlined in this PR](https://github.com/ecmwf-ifs/ectrans/pull/386).

So let's try just updating to 25.11? Maybe that will treat the `SELECT TYPE` statement properly. I picked this one because we have this version locally available on the ECMWF HPC.

samhatfield:
Something is totally missed up with how GH builds an action from build.yml. If I call the nvhpc job anything other than `nvhpc-26.3`, it reports a syntax error on line 105. `nvhpc` doesn't work, nor `nvhpc-25` or `nvhpc-25.11`. Some weird caching bug?

samhatfield:
Some of the checksum comparisons are failing for nvhpc/25.11. All of the failing comparisons have a --field-api run as one of the runs being checked. Progress but we're not there yet. Initial investigations suggest that the problem lies in the direct transform and is improved slightly (more levels match) when the "clamping" is removed.

samhatfield:
Here are some investigations of the remaining nvhpc checksum reproducibility issue:

I debug locally on ECMWF HPC with nvhpc/25.11. I build FIAT without MPI so I just focus on mpi0 tests. I enable FIELD_API in ecTrans.

I run the branch at commit 63b2bcbccaa28be37df2a99502c3a136aa945cc3. 112 comparisons failed out of 288.

I disable the clamping code in the benchmark program. 64 comparisons failed out of 288.

So first observation: I think clamping is not helping us.

Further observations:
- The remaining failures are all call mode 2.
- Not all call mode 2 cases fail, only some.
- It looks like the checksums match after the first call to INV_TRANS. This implies that DIR_TRANS is the problem.

src/programs/CMakeLists.txt
samhatfield: This needs to be reviewed before we merge -> why are we linking OpenACC to the CPU build?

CMakeLists.txt
```diff
@@ -119,6 +119,9 @@ if( (NOT HAVE_CPU) AND (NOT HAVE_GPU) )
 endif()
 
 if( HAVE_GPU )
+  if( HAVE_MPI)
+     ecbuild_find_package( NAME MPI REQUIRED COMPONENTS Fortran )
+  endif()
```
samhatfield: This also needs reviewing - we don't necessarily need MPI for the GPU feature. When GPU-aware MPI is disabled, we use MPL. The `GPU_AWARE_MPI` feature brings in MPI automatically, so this block should be redundant.

samhatfield:
Suspecting an issue with the clamping from my local tests, I tried disabling it in the branch and relaunching the CI tests. Now, everything passes. So clamping was the cause all along so I think we should just delete it.

dhaumont:
> Suspecting an issue with the clamping from my local tests, I tried disabling it in the branch and relaunching the CI tests. Now, everything passes. So clamping was the cause all along so I think we should just delete it.

I'm glad to read that the clamping had been deleted, I disliked this hack from the beginning.
One year ago, I had to implement it to ensure bit reproducibility. I'm wondering what can explain that it's not needed anymore...some changes to ecTrans ? A compiler upgrade ?
But since it's going into the right direction, I believe it's not necessary to investigate this any further.
Thank you Sam for having fixed this !

samhatfield:
> I'm glad to read that the clamping had been deleted, I disliked this hack from the beginning. One year ago, I had to implement it to ensure bit reproducibility. I'm wondering what can explain that it's not needed anymore...some changes to ecTrans ? A compiler upgrade ? But since it's going into the right direction, I believe it's not necessary to investigate this any further. Thank you Sam for having fixed this !

It's hard to say really... I can't think of anything that would cause nvfortran-compiled code to behave differently like this.

dhaumont:
> It's hard to say really... I can't think of anything that would cause nvfortran-compiled code to behave differently like this.

Could it be that we changed the callmode used in the field API unit tests, after you introduced this feature in the benchmark ?

samhatfield:
> Could it be that we changed the callmode used in the field API unit tests, after you introduced this feature in the benchmark ?

Not sure. The clamping rounded all spectral coefficients below 1e-14 to zero, but was only applied when --field-api was enabled. So if you're comparing the checksums against a reference that doesn't do clamping, there's no way they can match, right?

dhaumont:
> > Could it be that we changed the callmode used in the field API unit tests, after you introduced this feature in the benchmark ?
> 
> Not sure. The clamping rounded all spectral coefficients below 1e-14 to zero, but was only applied when --field-api was enabled. So if you're comparing the checksums against a reference that doesn't do clamping, there's no way they can match, right?

My guess at that time was that there was some filtering applied to the spectral fields deep in the FFT algorithm  in some configurations.
With our current  implementation, still based on the legacy INV_TRANS/DIR_TRANS, the only effect of `--field-api` is to change the order of the rows in the PGP/PGFL arrays that are passed to  INV_TRANS/DIR_TRANS. All the rest is completely identical.
This means that with our latest version, we provide the data in the exact same order to INV/DIR_TRANS  with or without the `--field-api` flag.
And that we use now the exact same FFT's algorithm (and filtering ?) in both cases.

samhatfield:
> My guess at that time was that there was some filtering applied to the spectral fields deep in the FFT algorithm in some configurations. With our current implementation, still based on the legacy INV_TRANS/DIR_TRANS, the only effect of `--field-api` is to change the order of the rows in the PGP/PGFL arrays that are passed to INV_TRANS/DIR_TRANS. All the rest is completely identical. This means that with our latest version, we provide the data in the exact same order to INV/DIR_TRANS with or without the `--field-api` flag. And that we use now the exact same FFT's algorithm (and filtering ?) in both cases.

With --field-api enabled, some arrays are probably being passed to INV/DIR_TRANS with a different arrangement in the "fields" dimension compared to the reference (no --field-api). After I disabled clamping, I still saw mismatching checksums in my local tests. Locally on my machine, I was using MKL in its default configuration. This is known to give different results for GEMM calls depending on the order of the leading dimension. When I put MKL in its reproducible mode (`export MKL_CBWR=AUTO,STRICT`), the tests passed. This suggests that indeed --field-api is rearranging fields, and as you say, this is the major difference numerically between with and without --field-api.

Anyway, everything's passing now. I'm ready for approval.

samhatfield (approved):

dhaumont:
@samhatfield @wdeconinck I'm afraid I just found a bug in the way we compute the dimensions of the temporary arrays in `INV_TRANS_FIELD_API`.

The following logic is not correct:
```
  IUVDIM = 2

  ! Output derivatives of vector fields
  IF (PRESENT(YDFU_EW) .AND. PRESENT(YDFV_EW))    THEN
    IUVDIM = 5
 ENDIF

  ! Output divergence of vector fields
  IF (PRESENT(YDFDIV)) THEN
    IUVDIM = 5
  ENDIF

  ! Output vorticity of vector fields
  IF (PRESENT(YDFVOR)) THEN
    IUVDIM = 6
  ENDIF
```

Instead, we should have something along those lines:
```
  IUVDIM = 2

  ! Output derivatives of vector fields
  IF (PRESENT(YDFU_EW) .AND. PRESENT(YDFV_EW))    THEN
    IUVDIM =     IUVDIM + 2
 ENDIF

  ! Output divergence of vector fields
  IF (PRESENT(YDFDIV)) THEN
       IUVDIM =     IUVDIM + 1
  ENDIF

  ! Output vorticity of vector fields
  IF (PRESENT(YDFVOR)) THEN
           IUVDIM =     IUVDIM + 1
  ENDIF
```

This is not detected by the unit tests, which does not test all the configurations of LDVOR, LDIV and UVDERS.

I'm investigating

dhaumont:
I'm trying to call `INV_TRANS_FIELD_API`  from ARPEGE, but it is not working for the moment. If you agree, I think it would be wiser to wait until I could fix it before merging the PR.

samhatfield:
Yeah I see what you mean. `IUVDIM` should be incremented according to which actual derivatives are requested. Keep us posted.

dhaumont:
> Yeah I see what you mean. `IUVDIM` should be incremented according to which actual derivatives are requested. Keep us posted.

I don't understand how it is possible that I introduced this into the code, this was correctly computed from the very first prototype: https://github.com/dhaumont/inv_trans/blob/master/inv_trans_field_api.F90.
Probably a wrong merge at some point.
I would keep you posted, but will be away for the next two weeks.

dhaumont:
I had a quick look at the failures and the builds are failing for various reasons.
Most of the time, it is a failure with the install unit test.
I also saw a crash in ectrans_lam_test_benchmark_dp_48x40_mpi0_omp8_nfld10_nlev20_vordiv  unit test.
Are those issues known and recurrent ?

Regarding the time out durnig compilation, is it due to the additional workload for the compilation of field API ? Is the compilation of executables and unit tests properly deactivated for ecTrans?

In any case, this PR can only impact the *field_api* unit tests. In my opinion, the failures we observe have nothing to do with it, and should not block the PR for beeing merged.

dhaumont:
Should we raise issues for those failures: https://github.com/ecmwf-ifs/ectrans/actions/runs/25960111528/job/76313871133?pr=314 ?

wdeconinck:
I agree with you @dhaumont there is something unrelated to this PR going on.
What seemed to be working in develop is also not working at the moment reliably of a rerun of the same commit (https://github.com/ecmwf-ifs/ectrans/actions/runs/25789542004/job/76478281467 (attempt 2, failed) vs https://github.com/ecmwf-ifs/ectrans/actions/runs/25789542004/job/75751235638 (attempt 1, succeeded)
, related to ETRANS (LAM) tests are segfaulting, only on our intel-ops HPC build.
This and other failures are now tracked in #403 

Furthermore there was something pathological going on with the hpc I believe regarding time-out of tests taking too long, and I needed to increase job time significantly for the hpc-ac-gpu build.
We can revisit test times later.

As far as I'm concerned we can merge this PR now.

wdeconinck (approved):
Thanks @dhaumont @samhatfield and everyone else involved in getting this over the line.

dhaumont:
and thank to you, Willem !!!

dhaumont:
@wdeconinck @samhatfield I see that you both approved the PR. Will you take care of the merge to develop ?

samhatfield:
Hi @wdeconinck @dhaumont, I noticed that fspgl_intf.h isn't in the include/ectrans directory like the other header files. Any reason why? https://github.com/ecmwf-ifs/ectrans/tree/develop/src/trans/include

dhaumont:
> Hi @wdeconinck @dhaumont, I noticed that fspgl_intf.h isn't in the include/ectrans directory like the other header files. Any reason why? https://github.com/ecmwf-ifs/ectrans/tree/develop/src/trans/include

It looks like a mistake, at least I can't remember any reason why it would have to be there.

PR #315, opened 2025-09-29
Remove dead code from previous GPU version

samhatfield:
As described in the comment, we pass(ed) `nflevl` via an environment variable so we didn't have to modify the ecTrans API, back when we had to know this in order to allocate GPU device buffers. Now that we do indeed grow those buffers automatically, we don't need this (and indeed this code has been "dead" for over a year now) :)


wdeconinck (approved):
Nice! I am happy that panned out as I planned it would :)

PR #317, opened 2025-10-16
Factorize duplication between trltog  and trgtol (cpu only)

dhaumont:
This PR is a refactoring to eliminate duplicated code between trltog and trgtol in the cpu branch.

The duplicated codehas been extracted into new routines in a new module trgl_mod, and they are called from trltog and trgtol.


samhatfield:
Thanks for this @dhaumont. Naturally it'll take some time to test, but assuming you've done your homework this could be a really nice restructuring.

Just a quick suggestion: `TRGTOL_PROLOG` and `TRLTOG_PROLOG` don't do much now - you could save even more complexity and line numbers by deleting them and moving the code into the parent subroutines?

samhatfield:
If I understand what you said the other day @dhaumont, you're well aware that this isn't finished, and just want some early feedback on the overall idea before you commit more time to finishing it.

I think in principle we'd be happy to accept a change like this. `TRLTOG`/`TRGTOL` are very much in the category of "too scary, don't touch it", and anything that simplifies them is welcome. It's good to reduce repetition, and I like how the actual communication routines are now shorter and to the point.

My only concern is that we're introducing a lot more variables (derived types) that are now wrapping the other variables. So it's sort of more complicated and less complicated at the same time.

With time maybe we can simplify even further without making the code more of a headache. Maybe we could work on this PR together, and keep it up to date with develop. Eventually we might decide not to merge it, but it wouldn't be a waste of time because we would learn a lot about `TRGTOL`/`TRLTOG` during the process, which are subroutines I've avoided up to now as they're too scary.

My question is, is this PR related to your FIELD API developments or is it orthogonal?

dhaumont:
Thanks a lot for taking the time to look at this @samhatfield, as well as for your comments.
- I agree that this refactoring is a bit scary, and this is a very good reason to try to simplify code and make it less intimidating :-).
- The introduction of new derived types to wrap variables is the main point of this refactoring.  Passing less arguments in function calls is a best practice that greatly simplifies the code and needs to be promoted. I don't really get your point that it makes things more complicated: there is nothing more than before, just a set of arguments that have been grouped together.
- This PR is completely orthogonal to FIELD API, but having less arguments in function calls should simplify the introduction of list of pointers later on
- it's a very good idea to work on this together and keep this development in a separated branch kept in sync with develop. When ready, we can ask other colleagues to use this branch in their day-to-day work, and eventually merge it when we are confident  that it's stable and error free.

dhaumont:
@samhatfield @wdeconinck The PR has now been properly tested and I no longer consider it a draft.
Could you update its status ?

dhaumont:
I've seen the compilation errors. I'm on it

samhatfield:
> I've seen the compilation errors. I'm on it

Looks like that line is just too long. We don't really have an official style guide for ecTrans (any guide would be broken in numerous places already...) but generally we try to limit lines to 100 characters.

dhaumont:
> > I've seen the compilation errors. I'm on it
> 
> Looks like that line is just too long. We don't really have an official style guide for ecTrans (any guide would be broken in numerous places already...) but generally we try to limit lines to 100 characters.

Indeed, It should be fixed now.

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -60,15 +63,13 @@ SUBROUTINE TRLTOG(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,KPTRGP,&
 USE PARKIND1  ,ONLY : JPIM     ,JPRB
 USE YOMHOOK   ,ONLY : LHOOK,   DR_HOOK, JPHOOK
 
-USE TPM_GEN         ,ONLY : NSTACK_MEMORY_TR
-USE TPM_DISTR       ,ONLY : D, NPRTRNS, NPROC
-USE TPM_TRANS       ,ONLY : NGPBLKS
-
+USE TPM_DISTR       ,ONLY : D
+USE TRGL_MOD, ONLY : TRGL_CTX, TRGL_VARS
```
samhatfield:
```suggestion
USE TRGL_MOD, ONLY : TRGL_CTX
```
Import not needed

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+  LOGICAL, ALLOCATABLE   :: LLUV(:)
+  LOGICAL, ALLOCATABLE   :: LLGP2(:)
+  LOGICAL, ALLOCATABLE   :: LLGP3A(:)
+  LOGICAL, ALLOCATABLE   :: LLGP3B(:)
+END TYPE TRGL_VARS
+
+CONTAINS
+
+SUBROUTINE ALLOCATE_CTX_CST(SELF)
+  USE TPM_DISTR       ,ONLY : D, MYSETW, NPRTRNS, MYPROC, NPROC
```
samhatfield:
```suggestion
  USE TPM_DISTR       ,ONLY : D, NPRTRNS, NPROC
```
Unused imports

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+  ALLOCATE (SELF%ISETWL(NPROC))
+  ALLOCATE (SELF%ISETVL(NPROC))
+
+END SUBROUTINE ALLOCATE_CTX_CST
+
+
+SUBROUTINE ALLOCATE_CTX_SR(SELF, KF_GP)
+  USE TPM_DISTR       ,ONLY : D, MYSETW, NPRTRNS, MYPROC
+  USE TPM_TRANS       ,ONLY : NGPBLKS
+  USE TPM_GEN         ,ONLY : NERR, NOUT
```
samhatfield:
```suggestion
  USE TPM_TRANS       ,ONLY : NGPBLKS
```
Unused imports

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+! nor does it submit to any jurisdiction.
+!
+
+MODULE TRGL_MOD
+
+USE PARKIND1, ONLY : JPIM
+
+IMPLICIT NONE
+
+TYPE TRGL_CTX
```
samhatfield: What does CTX actually mean? Could we have a more human-readable name for this type?

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+  INTEGER(KIND=JPIM), ALLOCATABLE :: IRECVTOT (:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: ISEND(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: IRECV(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: IINDEX(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: INDOFF(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: IGPTRSEND(:,:,:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: ISETAL(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: ISETBL(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: ISETWL(:)
+  INTEGER(KIND=JPIM), ALLOCATABLE :: ISETVL(:)
```
samhatfield: As far as I can see these variables are only used in `TRGL_PROLOG`, and then only as temporary containers. I think it's best if you just delete them and pass the local variables `ISETA` etc. directly to `PE2SET`.

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -164,244 +142,31 @@ SUBROUTINE TRLTOG_PROLOG(KF_FS,KF_GP,KVSET,&
 
 USE PARKIND1  ,ONLY : JPIM
 
-USE TPM_DISTR       ,ONLY : D, MYSETW, NPRTRNS, MYPROC, NPROC
-USE TPM_TRANS       ,ONLY : NGPBLKS
-
 USE INIGPTR_MOD     ,ONLY : INIGPTR
 USE PE2SET_MOD      ,ONLY : PE2SET
```
samhatfield:
```suggestion
```
Unused import

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -464,13 +229,16 @@ SUBROUTINE TRLTOG_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
 
-USE TPM_GEN         ,ONLY : NOUT, NTRANS_SYNC_LEVEL
-USE TPM_DISTR       ,ONLY : D, MYSETV, MYSETW, MTAGLG,      &
-     &                      NPRCIDS, NPRTRNS, MYPROC, NPROC
-USE TPM_TRANS       ,ONLY : LDIVGP, LSCDERS, LUVDER, LVORGP, NGPBLKS
+USE TPM_GEN         ,ONLY : NTRANS_SYNC_LEVEL, NSTACK_MEMORY_TR
+USE TPM_DISTR       ,ONLY : D, MTAGLG,      &
+     &                      NPRCIDS, MYPROC, NPROC
 
 USE PE2SET_MOD      ,ONLY : PE2SET
```
samhatfield:
```suggestion
```
Unused import

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -464,13 +229,16 @@ SUBROUTINE TRLTOG_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
 USE ABORT_TRANS_MOD ,ONLY : ABORT_TRANS
```
samhatfield:
```suggestion
```
Unused import

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -479,63 +247,61 @@ SUBROUTINE TRLTOG_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
 
 REAL(KIND=JPHOOK) :: ZHOOK_HANDLE_BAR
 
 !     ------------------------------------------------------------------
 
 !*       0.    Some initializations
 !              --------------------
+ASSOCIATE (KSENDCOUNT=>YDCTX%ISENDCOUNT, KRECVCOUNT=>YDCTX%IRECVCOUNT, KNSEND=>YDCTX%INSEND, KNRECV=>YDCTX%INRECV, &
+         & KSENDTOT=>YDCTX%ISENDTOT, KRECVTOT=>YDCTX%IRECVTOT, KSEND=>YDCTX%ISEND, KRECV=>YDCTX%IRECV, KINDEX=>YDCTX%IINDEX, &
+         & KNDOFF=>YDCTX%INDOFF, KGPTRSEND =>YDCTX%IGPTRSEND, KSETAL=>YDCTX%ISETAL, KSETBL=>YDCTX%ISETBL, KSETWL=>YDCTX%ISETWL)
```
samhatfield:
```suggestion
ASSOCIATE (KNSEND=>YDCTX%INSEND, KNRECV=>YDCTX%INRECV, &
         & KSENDTOT=>YDCTX%ISENDTOT, KRECVTOT=>YDCTX%IRECVTOT, KSEND=>YDCTX%ISEND, KRECV=>YDCTX%IRECV, KINDEX=>YDCTX%IINDEX, &
         & KNDOFF=>YDCTX%INDOFF)
```
Unused associations

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -557,268 +323,22 @@ SUBROUTINE TRLTOG_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
-      ENDDO
-    ENDDO
-    IOFFEW=IOFFEW+IGP3BPAR*IGP3BLEV
-  ENDIF
-ENDIF  
-
+YDCTX%LLINDER = .FALSE.
+YDCTX%LLPGPONLY = .FALSE.
+IF(PRESENT(KPTRGP))  YDCTX%LLINDER = .TRUE.
+IF(PRESENT(PGP))     YDCTX%LLPGPONLY = .TRUE.
```
samhatfield:
```suggestion
YDCTX%LLINDER = PRESENT(KPTRGP)
YDCTX%LLPGPONLY = PRESENT(PGP)
```

src/trans/cpu/internal/trltog_mod.F90
```diff
@@ -164,244 +142,31 @@ SUBROUTINE TRLTOG_PROLOG(KF_FS,KF_GP,KVSET,&
 
 USE PARKIND1  ,ONLY : JPIM
 
-USE TPM_DISTR       ,ONLY : D, MYSETW, NPRTRNS, MYPROC, NPROC
-USE TPM_TRANS       ,ONLY : NGPBLKS
-
 USE INIGPTR_MOD     ,ONLY : INIGPTR
```
samhatfield:
```suggestion
```
Unused import

src/trans/cpu/internal/trgl_mod.F90
samhatfield: My bad, I see that `ISETWL` and `ISETVL` _are_ useful, but `ISETAL` and `ISETBL` seem to be unnecessary.

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+  INTEGER(KIND=JPIM), ALLOCATABLE :: IGP3BLEVS(:)
+  LOGICAL, ALLOCATABLE   :: LLUV(:)
+  LOGICAL, ALLOCATABLE   :: LLGP2(:)
+  LOGICAL, ALLOCATABLE   :: LLGP3A(:)
+  LOGICAL, ALLOCATABLE   :: LLGP3B(:)
+END TYPE TRGL_VARS
+
+CONTAINS
+
+SUBROUTINE ALLOCATE_CTX_CST(SELF)
```
samhatfield: What does CST mean?

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+  ALLOCATE (SELF%IGPTRSEND(2,NGPBLKS,NPRTRNS))
+  ALLOCATE (SELF%ISETAL(NPROC))
+  ALLOCATE (SELF%ISETBL(NPROC))
+  ALLOCATE (SELF%ISETWL(NPROC))
+  ALLOCATE (SELF%ISETVL(NPROC))
+
+END SUBROUTINE ALLOCATE_CTX_CST
+
+
+SUBROUTINE ALLOCATE_CTX_SR(SELF, KF_GP)
```
samhatfield: Similar, what does SR mean?

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+            IGPTROFF=>YLVARS%IGPTROFF,&
+            IUVPARS=>YLVARS%IUVPARS, &
+            IGP2PARS=>YLVARS%IGP2PARS, LLUV=>YLVARS%LLUV, &
+            LLGP2=>YLVARS%LLGP2, LLGP3A=>YLVARS%LLGP3A, &
+            LLGP3B=>YLVARS%LLGP3B, IGP3APARS=>YLVARS%IGP3APARS, &
+            IGP3ALEVS=>YLVARS%IGP3ALEVS, IGP3BPARS=>YLVARS%IGP3BPARS, &
+            IGP3BLEVS=>YLVARS%IGP3BLEVS)
+  ASSOCIATE (KSENDCOUNT=>YDCTX%ISENDCOUNT, KRECVCOUNT=>YDCTX%IRECVCOUNT, KNSEND=>YDCTX%INSEND, KNRECV=>YDCTX%INRECV, &
+          & KSENDTOT=>YDCTX%ISENDTOT, KRECVTOT=>YDCTX%IRECVTOT, KSEND=>YDCTX%ISEND, KRECV=>YDCTX%IRECV, KINDEX=>YDCTX%IINDEX, &
+          & KNDOFF=>YDCTX%INDOFF, KGPTRSEND =>YDCTX%IGPTRSEND,IFLDS=>YDCTX%IFLDS,LLPGPONLY=>YDCTX%LLPGPONLY, LLINDER=>YDCTX%LLINDER)
```
samhatfield:
```suggestion
  ASSOCIATE(KINDEX=>YDCTX%IINDEX, KNDOFF=>YDCTX%INDOFF, KGPTRSEND=>YDCTX%IGPTRSEND, &
    &       IFLDS=>YDCTX%IFLDS, LLPGPONLY=>YDCTX%LLPGPONLY)
```
Unused associations

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+    IF(IFIRST > 0) THEN
+      ILAST = KGPTRSEND(2,JBLK,MYSETW)
+      IPOS=IPOS+ILAST-IFIRST+1
+    ENDIF
+  ENDDO
+  END ASSOCIATE
+  END ASSOCIATE
+END SUBROUTINE TRGL_INIT_OFF_VARS
+
+SUBROUTINE TGRL_INIT_PACKING_VARS(YDCTX,YLVARS, KVSET, KF_GP, PCOMBUFS)
```
samhatfield: `YLVARS` isn't used in this subroutine

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+        IPOSPLUS(INRS)=IPOSPLUS(INRS)+(ILAST-IFIRST+1)
+        IPOS=IPOS+(ILAST-IFIRST+1)
+      ENDIF
+    ENDDO
+    IF (PRESENT(PCOMBUFS)) THEN
+      PCOMBUFS(-1,INRS) = 1
+      PCOMBUFS(0,INRS) = IFLD
+    ENDIF
+  ENDDO
+!$OMP END PARALLEL DO
```
samhatfield:
```suggestion
  !$OMP END PARALLEL DO
```

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,784 @@
...
+            IGP2PARS=>YLVARS%IGP2PARS, LLUV=>YLVARS%LLUV, &
+            LLGP2=>YLVARS%LLGP2, LLGP3A=>YLVARS%LLGP3A, &
+            LLGP3B=>YLVARS%LLGP3B, IGP3APARS=>YLVARS%IGP3APARS, &
+            IGP3ALEVS=>YLVARS%IGP3ALEVS, IGP3BPARS=>YLVARS%IGP3BPARS, &
+            IGP3BLEVS=>YLVARS%IGP3BLEVS)
+  ASSOCIATE (KSENDCOUNT=>YDCTX%ISENDCOUNT, KRECVCOUNT=>YDCTX%IRECVCOUNT, KNSEND=>YDCTX%INSEND, KNRECV=>YDCTX%INRECV, &
+          & KSENDTOT=>YDCTX%ISENDTOT, KRECVTOT=>YDCTX%IRECVTOT, KSEND=>YDCTX%ISEND, KRECV=>YDCTX%IRECV, KINDEX=>YDCTX%IINDEX, &
+          & KNDOFF=>YDCTX%INDOFF, KGPTRSEND =>YDCTX%IGPTRSEND,&
+          & IFLDA=>YDCTX%IFLDA,IPOSPLUS=>YDCTX%IPOSPLUS, JPOS=>YDCTX%IJPOS, ISETW=>YDCTX%ISETW,LLPGPONLY=>YDCTX%LLPGPONLY, &
+          & LLINDER=>YDCTX%LLINDER)
```
samhatfield:
```suggestion
  ASSOCIATE(KGPTRSEND =>YDCTX%IGPTRSEND, IFLDA=>YDCTX%IFLDA, IPOSPLUS=>YDCTX%IPOSPLUS, &
    &       JPOS=>YDCTX%IJPOS, ISETW=>YDCTX%ISETW,LLPGPONLY=>YDCTX%LLPGPONLY, &
    &       LLINDER=>YDCTX%LLINDER)
```
Unused imports

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -56,66 +56,47 @@ SUBROUTINE TRGTOL(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,KPTRGP,&
 USE PARKIND1  ,ONLY : JPIM     ,JPRB
 USE YOMHOOK   ,ONLY : LHOOK,   DR_HOOK, JPHOOK
 
-USE TPM_GEN         ,ONLY : NSTACK_MEMORY_TR
-USE TPM_DISTR       ,ONLY : D, NPRTRNS, NPROC
-USE TPM_TRANS       ,ONLY : NGPBLKS
+USE TPM_DISTR       ,ONLY : D
+USE TRGL_MOD, ONLY : TRGL_CTX, TRGL_VARS
```
samhatfield:
```suggestion
USE TRGL_MOD, ONLY : TRGL_CTX
```

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -154,230 +135,32 @@ SUBROUTINE TRGTOL_PROLOG(KF_FS,KF_GP,KVSET,&
...
-
-
-USE PARKIND1  ,ONLY : JPIM 
-
-USE TPM_DISTR       ,ONLY : D, NPRTRNS, MYSETW, MYPROC, NPROC
-USE TPM_TRANS       ,ONLY : NGPBLKS
+USE PARKIND1  ,ONLY : JPIM
 
 USE INIGPTR_MOD     ,ONLY : INIGPTR
 USE PE2SET_MOD      ,ONLY : PE2SET
```
samhatfield:
```suggestion
```
Unused

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -434,82 +217,83 @@ SUBROUTINE TRGTOL_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
-USE TPM_GEN         ,ONLY : NOUT, NTRANS_SYNC_LEVEL
-USE TPM_DISTR       ,ONLY : D, NPRCIDS, NPRTRNS, MTAGGL,  &
-     &                      MYSETV, MYSETW, MYPROC, NPROC
-USE TPM_TRANS       ,ONLY : LDIVGP, LGPNORM, LSCDERS, LUVDER, LVORGP, NGPBLKS
+USE TPM_GEN         ,ONLY : NTRANS_SYNC_LEVEL, NSTACK_MEMORY_TR
+USE TPM_DISTR       ,ONLY : D, MTAGGL,  &
+     &                      NPRCIDS, MYPROC, NPROC
+USE TPM_TRANS       ,ONLY :  LGPNORM
 
 USE PE2SET_MOD      ,ONLY : PE2SET
```
samhatfield:
```suggestion
```
Unused

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -154,230 +135,32 @@ SUBROUTINE TRGTOL_PROLOG(KF_FS,KF_GP,KVSET,&
...
 
-
-
-USE PARKIND1  ,ONLY : JPIM 
-
-USE TPM_DISTR       ,ONLY : D, NPRTRNS, MYSETW, MYPROC, NPROC
-USE TPM_TRANS       ,ONLY : NGPBLKS
+USE PARKIND1  ,ONLY : JPIM
 
 USE INIGPTR_MOD     ,ONLY : INIGPTR
```
samhatfield:
```suggestion
```
Unused

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -434,82 +217,83 @@ SUBROUTINE TRGTOL_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
 
 REAL(KIND=JPHOOK) :: ZHOOK_HANDLE_BAR
 
 !     ------------------------------------------------------------------
 
 !*       0.    Some initializations
 !              --------------------
+ASSOCIATE (KSENDCOUNT=>YDCTX%ISENDCOUNT, KRECVCOUNT=>YDCTX%IRECVCOUNT, KNSEND=>YDCTX%INSEND, KNRECV=>YDCTX%INRECV, &
+& KSENDTOT=>YDCTX%ISENDTOT, KRECVTOT=>YDCTX%IRECVTOT, KSEND=>YDCTX%ISEND, KRECV=>YDCTX%IRECV, KINDEX=>YDCTX%IINDEX, &
+& KNDOFF=>YDCTX%INDOFF, KGPTRSEND =>YDCTX%IGPTRSEND, KSETAL=>YDCTX%ISETAL, KSETBL=>YDCTX%ISETBL, KSETWL=>YDCTX%ISETWL)
```
samhatfield:
```suggestion
ASSOCIATE(KNSEND=>YDCTX%INSEND, KNRECV=>YDCTX%INRECV, KSENDTOT=>YDCTX%ISENDTOT, &
  &       KRECVTOT=>YDCTX%IRECVTOT, KSEND=>YDCTX%ISEND, KRECV=>YDCTX%IRECV, KINDEX=>YDCTX%IINDEX, &
  &       KNDOFF=>YDCTX%INDOFF)
```

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -434,82 +217,83 @@ SUBROUTINE TRGTOL_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
+
+IF (NSTACK_MEMORY_TR == 0) THEN
+  CALL TRGL_ALLOCATE_HEAP_BUFFER(ZCOMBUFS_HEAP, YDCTX%ISENDCOUNT, YDCTX%INSEND)
+  CALL TRGL_ALLOCATE_HEAP_BUFFER(ZCOMBUFR_HEAP, YDCTX%IRECVCOUNT, YDCTX%INRECV)
+
+! Now, force the OS to allocate this shared array right now, not when it starts to be used which is
+! an OPEN-MP loop, that would cause a threads synchronization lock :
```
samhatfield:
```suggestion
  ! Now, force the OS to allocate this shared array right now, not when it starts to be used which is
  ! an OPEN-MP loop, that would cause a threads synchronization lock :
```

src/trans/cpu/internal/trgtol_mod.F90
```diff
@@ -541,271 +325,23 @@ SUBROUTINE TRGTOL_COMM(PGLAT,KF_FS,KF_GP,KF_SCALARS_G,KVSET,&
...
-      ENDDO
-    ENDDO
-    IOFFEW=IOFFEW+IGP3BPAR*IGP3BLEV
-  ENDIF
-ENDIF
-
+YDCTX%LLINDER = .FALSE.
+YDCTX%LLPGPONLY = .FALSE.
+IF(PRESENT(KPTRGP))  YDCTX%LLINDER = .TRUE.
+IF(PRESENT(PGP))     YDCTX%LLPGPONLY = .TRUE.
```
samhatfield:
```suggestion
YDCTX%LLINDER = PRESENT(KPTRGP)
YDCTX%LLPGPONLY = PRESENT(PGP)
```

samhatfield (commented):
Great work @dhaumont. I think this will nicely reduce the code volume. I have some small suggestions above. Could you go through them and let me know your reaction, or otherwise resolve them?

One thing I was thinking about was whether we anymore need the "prolog" and "comm" subroutine splitting anymore. The `_PROLOG` subroutines in particular are now very small. I think we can just delete those and instead put the calls directly in `TRGTOL`/`TRLTOG`. I tried that, then realised that it might also be nicer to put `TRGTOL_COMM` directly in `TRGTOL` as well. I tried that and it saves about 140 lines of code in each subroutine. What do you think?

dhaumont:
> Great work @dhaumont. I think this will nicely reduce the code volume. I have some small suggestions above. Could you go through them and let me know your reaction, or otherwise resolve them?
> 
> One thing I was thinking about was whether we anymore need the "prolog" and "comm" subroutine splitting anymore. The `_PROLOG` subroutines in particular are now very small. I think we can just delete those and instead put the calls directly in `TRGTOL`/`TRLTOG`. I tried that, then realised that it might also be nicer to put `TRGTOL_COMM` directly in `TRGTOL` as well. I tried that and it saves about 140 lines of code in each subroutine. What do you think?

Thank you for your thorough review, @samhatfield  ! I will implement the small changes you suggested. 

Regarding the `_trolog` and `_comm` functions. it would be very nice to get rid of them to further simplify the code. Since you have already done it, perhaps it makes sense if you could take over this part ? Alternatively, you can send me a patch with your changes, and I'll finalize everything.

samhatfield:
Yeah I can just submit a pull request to your branch. When you approve that, this one will be automatically updated.

src/trans/cpu/internal/trgl_mod.F90
dhaumont: Renamed int BUFFERS

dhaumont:
@samhatfield All your suggestions have been pushed. Looking forward to your PR into my branch

samhatfield:
Unfortunately it wasn't possible to collapse `TRGTOL` and `TRGTOL_COMM` into one subroutine, or at least _I_ couldn't find a way (same issue with `TRLTOG`). The issue is that the stack versions of `ZCOMBUFS/R` have a shape which depends on `YDBUFS%INSEND`, and this must first be defined in `TRGTOL`. So, you have an array whose dimensions must be known at compile time, but the dimensions are only defined at run time...

samhatfield:
Overall I think this is a nice change and I like how the communication routines are much shorter and the complicated index calculations are done elsewhere. Now `TRGTOL`/`TRLTOG` are mostly about communication which makes them easier to read IMO. I'd be happy to merge this, but I'd like to hear an opinion of @RyadElKhatibMF first. I also think we should take this chance to comment the code a bit more.

RyadElKhatibMF:
yes, it looks fine.

src/trans/cpu/internal/trgl_mod.F90
```diff
@@ -0,0 +1,746 @@
...
+
+SUBROUTINE TRGL_ALLOCATE_HEAP_BUFFER(Z_HEAP, S1, S2)
+  USE PARKIND1  ,ONLY : JPIM, JPRB
+
+  IMPLICIT NONE
+
+  REAL(KIND=JPRB), INTENT(INOUT), ALLOCATABLE :: Z_HEAP(:,:)
+  INTEGER(KIND=JPIM), INTENT(IN) :: S1, S2
+
+  IF (ALLOCATED(Z_HEAP) .AND. (S1 /= SIZE(Z_HEAP,1) .OR. S2 /= SIZE(Z_HEAP,2) )) THEN
```
samhatfield: `Z_HEAP`'s first dimension always starts at -1. Hence it will always have `S1`+2 elements. Won't `S1 /= SIZE(Z_HEAP,1)` therefore always return `.TRUE.`? I think the solution is to use `UBOUND` instead of `SIZE`.

dhaumont: This one is for Philippe @pmarguinaud

wdeconinck: @samhatfield According to the Intel documentation of [SIZE](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-2/size-function.html), it should count all elements in a dimension no matter if you changed the UBOUNDS or LBOUNDS:
```
SIZE = UBOUND - LBOUND + 1
```
Does this address your concern?

samhatfield: That's the problem. `Z_HEAP` is allocated like `ALLOCATE(Z_HEAP(-1:S1,S2))`. That means that `S1` is the upper bound. The size of that dimension is then `S1 + 2`, so the above inequality check will be true, always. `Z_HEAP` will be deallocated and allocated on every call rather than being reused as it should. If we use `UBOUND` then the bug is fixed.

wdeconinck: Indeed, very subtle.
The if-statement should then also check for `LBOUND(ZHEAP,1)` to be -1 to be sure.
If those conditions are not met then reallocation may happen. @dhaumont can you take care of that?

dhaumont: Yes, I'll take care of it. It's a bit beyond my Fortran skills, but it's a good opportunity to learn (the code is coming from the prototype of Philippe)

dhaumont: Solved here: https://github.com/ecmwf-ifs/ectrans/pull/317/commits/ab9eaa952f01ec1bc5711de68adea3414266ef15

Indeed, the size computation was always off by 2...thanks for spotting this, @samhatfield !

samhatfield:
I'll make one last check and then I think this is ready to merge.

samhatfield:
I realised after I started testing that there is still a bug in the checksum dumps for spectral fields. I'd like to test `trltog_clean` properly looking at checksums etc., but this isn't possible until the bug is fixed, so I opened a new PR [here](https://github.com/ecmwf-ifs/ectrans/pull/349). Let's review and merge that first, then we can merge develop into `trltog_clean` so we can properly test it?

samhatfield:
@dhaumont could you merge develop into this branch? This will make testing easier.

EDIT: nevermind, I did it myself.

samhatfield:
Okay, here are some benchmark results. I used ECMWF's HPC2020 (Atos) and ran a strong scaling test at T1279 from 1 up to 64 nodes:

number_of_nodes | tasks_per_node | cpus_per_task | inverse (develop) | direct (develop | both (develop) | inverse (trltog_clean) | direct (trltog_clean) | both (trltog_clean) | inverse (ratio) | direct (ratio) | both (ratio)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 16 | 8 | 2.3183 | 1.3415 | 3.6587 | 2.5742 | 1.4262 | 4.0014 | 1.11 | 1.06 | 1.09
2 | 16 | 8 | 1.1571 | 0.6278 | 1.7846 | 1.2158 | 0.6874 | 1.9032 | 1.05 | 1.09 | 1.07
4 | 16 | 8 | 0.6137 | 0.3383 | 0.9525 | 0.6378 | 0.3425 | 0.9808 | 1.04 | 1.01 | 1.03
8 | 16 | 8 | 0.3463 | 0.2408 | 0.5871 | 0.3484 | 0.2276 | 0.5755 | 1.01 | 0.95 | 0.98
16 | 16 | 8 | 0.1909 | 0.1206 | 0.3118 | 0.1878 | 0.1127 | 0.3004 | 0.98 | 0.93 | 0.96
32 | 16 | 8 | 0.1162 | 0.0728 | 0.1886 | 0.1103 | 0.0699 | 0.1801 | 0.95 | 0.96 | 0.95
64 | 16 | 8 | 0.0674 | 0.047 | 0.12 | 0.0673 | 0.0433 | 0.1107 | 1.00 | 0.92 | 0.92

All timings given in seconds, the median time over 10 iterations.

We can also focus on the regions we're touching, 157 and 158 (timings here are averages not median, in ms):

number_of_nodes | 157 (L to G) (develop) | 158 (G to L) (develop) | 157 (L to G) (trltog_clean) | 158 (G to L) (trltog_clean) | 157 (Lto G) (ratio) | 158 (G to L) ratio
-- | -- | -- | -- | -- | -- | --
1 | 1160.43 | 662.206 | 1404.958 | 725.849 | 1.21 | 1.10
2 | 492.139 | 240.925 | 568.197 | 257.623 | 1.15 | 1.07
4 | 241.849 | 105.255 | 261.985 | 106.215 | 1.08 | 1.01
8 | 143.09 | 62.947 | 146.389 | 59.1 | 1.02 | 0.94
16 | 82.544 | 36.567 | 79.607 | 34.496 | 0.96 | 0.94
32 | 48.499 | 22.171 | 45.072 | 21.293 | 0.93 | 0.96
64 | 29.305 | 15.622 | 29.491 | 15.156 | 1.01 | 0.97

It looks like this PR adds some extra initialisation overhead which can be significant for extreme cases of high problem size on low node count (we never really run T1279 on a single node). For actual cases, like T1279 on 64 nodes, there's no notable change in wall time. Does that make sense @dhaumont? Anything surprising?

samhatfield:
Note: I realise this PR is about code tidiness and not speed, I'm just trying to confirm that there's no significant impact on wall times.

dhaumont:
Thank you so much for these tests, Sam. The results are really informative. We should definitely also perform the same for https://github.com/ecmwf-ifs/ectrans/pull/314.

Yes, I'm surprised with the results, I didn't expect a difference at all, and you observed 5-10% which is a lot.
Are you comparing with heap buffers in the two configurations ?
Could it be that we do some extra memory reallocations ?
Otherwise the main difference are the allocation of  the buffers if the new types in `ALLOCATE_BUFFERS_CST` and `ALLOCATE_BUFFERS_SR` that are performed on the heap and were statically allocated on the stack before. Can this make such a big difference ?

dhaumont:
can you elaborate on this : "We can also focus on the regions we're touching, 157 and 158 (timings here are averages not median, in ms)" ? I'm not sure to understand what you measure exactly

samhatfield:
> Yes, I'm surprised with the results, I didn't expect a difference at all, and you observed 5-10% which is a lot. Are you comparing with heap buffers in the two configurations ? Could it be that we do some extra memory reallocations ? Otherwise the main difference are the allocation of the buffers if the new types in `ALLOCATE_BUFFERS_CST` and `ALLOCATE_BUFFERS_SR` that are performed on the heap and were statically allocated on the stack before. Can this make such a big difference ?

Yeah I pass `KOPT_MEMORY_TR = 0` so I must be using heap buffers. I'm not sure what to expect to be honest. Another change is that `ZCOMBUF{S,R}_STACK` are now _always_ allocated rather than only when `TRGTOL_COMM_STACK` is called. Could that make a difference?

> can you elaborate on this : "We can also focus on the regions we're touching, 157 and 158 (timings here are averages not median, in ms)" ? I'm not sure to understand what you measure exactly

Those are GSTATS regions. I'm taking those numbers from the end of the STDOUT of the benchmark run:

```
STATS FOR ALL TASKS
 NUM ROUTINE                                     CALLS      MEAN(ms)       MAX(ms)   FRAC(%)  UNBAL(%)
   0 PROGRAM        - Total                          1     54556.284     54556.338    100.00      0.00
   1 SETUP_TRANS0   - Setup ecTrans                  1        22.786        22.837      0.04      0.00
   2 SETUP_TRANS    - Setup ecTrans handle           1      4350.622      4350.728      7.97      0.00
   3 TIME STEP      - Time step                     10      3738.489      3738.489     68.53      0.00
   4 INV_TRANS      - Inverse transform             10      2318.278      2435.000     44.63      2.14
   5 DIR_TRANS      - Direct transform              10      1340.901      1399.966     25.66      1.08
   6 NORMS          - Norm comp. (optional)         10        79.304       145.520      2.67      1.21
 102 LTINV_CTL      - Inv. Legendre transform       10       414.146       425.113      7.79      0.20
 103 LTDIR_CTL      - Dir. Legendre transform       10       296.746       304.142      5.57      0.14
 106 FTDIR_CTL      - Dir. Fourier transform        10       212.835       233.692      4.28      0.38
 107 FTINV_CTL      - Inv. Fourier transform        10       516.628       580.303     10.64      1.17
 140 SULEG          - Comp. of Leg. poly.            1      1652.712      1686.581      3.09      0.06
 152 LTINV_CTL      - M to L transposition          10       155.048       160.288      2.94      0.10
 153 LTDIR_CTL      - L to M transposition          10       135.008       215.878      3.96      1.48
 157 FTINV_CTL      - L to G transposition          10      1160.430      1300.329     23.83      2.56
 158 FTDIR_CTL      - G to L transposition          10       662.206       749.749     13.74      1.60
 400 GSTATS         - GSTATS itself               1159         0.003         0.004      0.01      0.00
TOTAL MEASURED IMBALANCE =       0.0 SECONDS,  0.0 PERCENT
TOTAL WALLCLOCK TIME      54.556 CPU TIME    333.699 VECTOR TIME     333.699
```

samhatfield:
BTW I also ran the above benchmark suite with checksum dumping enabled, and all 14 cases have IDENTICAL checksums, both inverse and direct :)

wdeconinck:
It would be nice to understand the increase in setup time.

@samhatfield you say that we don't run this on a single node, which is true for the forecast but not necessarily for pre- and post-processing tools (e.g. sptogp and gptosp, python tools, ...).

So it would be nice to keep the performance or memory footprints the same (or better!)
Thank you @samhatfield btw to benchmark this!

dhaumont:
After discussion with Daan @ddegrauwe, the memory access speed is very different when you allocate  buffers on the stack or on the heap in Fortran. 
I don't really understand why, probably some optimizations that are disabled in the slower case. I will try to allocate the buffers on the stack again and see if it makes a difference.

wdeconinck:
Indeed stack allocation is quicker, but depending on the size may require a large stack size, which may not always be a nice thing. I am in favour of heap allocations in general because it is safer, but we should not allow current performance to degrade.

Are these buffers are allocated/reallocated often?

dhaumont:
They are re-allocated everytime we execute trltog or trgtol.
Those are small buffers, dimensioned to the number of procs.
For me, it should not hurt, so I'm investigating.

dhaumont:
I could also observe the same performance penalty on my tests.
As discussed before, I have implemented a stack based version, using POINTERS instead of ALLOCATABLE:
https://github.com/dhaumont/ectrans/tree/trltog_stack
Unfortunately, I don't have any performance improvement so far.

wdeconinck:
Perhaps it is unrelated to stack allocation then, and best to look with a profiler where differences could be coming from?

dhaumont:
Indeed. I just started to profile using map.
If you have other profilers in mind for ATOS, I'm interested

dhaumont:
I just fixed a bug introduced during the refactoring:
https://github.com/ecmwf-ifs/ectrans/pull/317/commits/591f30d9d7c7abd8599ee671b7d821ff81b08f5d
I don't see any performance penalty anymore.

@samhatfield Could you please relaunch your strong scaling benchmark ?
Thanks,

samhatfield:
number_of_nodes | tasks_per_node | cpus_per_task | inverse (develop) | direct (develop) | both (develop) | inverse (trltog_clean) | direct (trltog_clean) | both (trltog_clean) | inverse (ratio) | direct (ratio) | both (ratio)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 16 | 8 | 2.3213 | 1.3403 | 3.6618 | 2.5704 | 1.4256 | 3.9961 | 1.11 | 1.06 | 1.09
2 | 16 | 8 | 1.1477 | 0.6225 | 1.7701 | 1.2259 | 0.6853 | 1.9107 | 1.07 | 1.10 | 1.08
4 | 16 | 8 | 0.6245 | 0.3495 | 0.9732 | 0.639 | 0.3393 | 0.9788 | 1.02 | 0.97 | 1.01
8 | 16 | 8 | 0.3391 | 0.2384 | 0.5789 | 0.3505 | 0.2255 | 0.5753 | 1.03 | 0.95 | 0.99
16 | 16 | 8 | 0.1904 | 0.1123 | 0.3029 | 0.1915 | 0.112 | 0.3035 | 1.01 | 1.00 | 1.00
32 | 16 | 8 | 0.1206 | 0.0848 | 0.204 | 0.1155 | 0.0696 | 0.1861 | 0.96 | 0.82 | 0.91
64 | 16 | 8 | 0.0677 | 0.0493 | 0.1173 | 0.069 | 0.0475 | 0.1161 | 1.02 | 0.96 | 0.99

"develop" = 1cfdcef87073dafd17e4e091d99651dee86a12bb
"trltog_clean" = 591f30d9d7c7abd8599ee671b7d821ff81b08f5d

Looks like we're measuring different things @dhaumont because the performance is the same as before. Are you able to reproduce e.g. one of the cases above on ECMWF/Atos? I run with options
```
--norms --truncation 1279 --niter 10 --nlev 137 --nfld 1 --uvders --scders --meminfo
```
and `OMP_PLACES=cores`.

dhaumont:
Indeed, I was testing much smaller test cases.
Unfortunately, we are not able to reproduce your timings at all.

For 2 nodes, I'm using this setup:
```
export OMP_NUM_THREADS=8
export OMP_STACKSIZE=2G
export OMP_PLACES=cores
salloc -N 2 --mem=200G --qos=np --ntasks-per-node=16 --cpus-per-task=8 --time=01:00:00
```

Then, I run the benchmark like this:
```
 srun build-intel/bin/ectrans-benchmark-cpu-dp --norms --truncation 1279 --niter 10 --nlev 137 --nfld 1 --uvders --scders --meminfo
```
And the results are around 4.48s per time step, while you reported 1.7701s.

Do you see what is wrong here ? Could you provide us your complete slurm settings ?

samhatfield:
Sorry I never mentioned - I benchmark with single precision, can you try that?

dhaumont:
> Sorry I never mentioned - I benchmark with single precision, can you try that?

We also just figured it out with Daan after reading this : https://sites.ecmwf.int/docs/ectrans/page/benchmarking.html
Let's try that !

dhaumont:
I reproduced the benchmark with 2 and 4 nodes. The performance is the same for the different configurations (develop, trltog_clean and trltog_stack)
```
develop_2.txt : number_of_nodes:2, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 45.790
TIME STEP 3149.707
INV_TRANS - 1933.485
DIR_TRANS - 1152.207
SETUP_TRANS0 - 0.090
SETUP_TRANS - 3959.033


trltog_2.txt : number_of_nodes:2, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 45.710
TIME STEP 3121.278
INV_TRANS - 1917.861
DIR_TRANS - 1132.118
SETUP_TRANS0 - 0.207
SETUP_TRANS - 4142.525


stack_2.txt : number_of_nodes:2, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 45.246
TIME STEP 3094.295
INV_TRANS - 1920.487
DIR_TRANS - 1120.938
SETUP_TRANS0 - 0.202
SETUP_TRANS - 3917.210


develop_4.txt : number_of_nodes:4, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 19.598
TIME STEP 1211.485
INV_TRANS - 755.819
DIR_TRANS - 427.190
SETUP_TRANS0 - 118.913
SETUP_TRANS - 3008.922


trltog_4.txt : number_of_nodes:4, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 19.199
TIME STEP 1211.618
INV_TRANS - 751.439
DIR_TRANS - 429.551
SETUP_TRANS0 - 61.377
SETUP_TRANS - 2699.599


stack_4.txt : number_of_nodes:4, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 19.445
TIME STEP 1208.975
INV_TRANS - 751.532
DIR_TRANS - 433.370
SETUP_TRANS0 - 91.354
SETUP_TRANS - 2949.127
```

I don't know what can explain the differences with your benchmark

dhaumont:
Is the script generating the table from the results available somewhere ?

samhatfield:
Hmm those all still look a bit slow to me. Something like 1.8x too slow. Maybe we're building with different environments?

I generate all my results with JUBE which I'm happy to send instructions on, but before that you could try to reproduce one of my figures. Here's a Slurm script you can try:
```bash
#!/bin/bash
#SBATCH --qos=np
#SBATCH --time=01:00:00
#SBATCH --output=ectrans_benchmark.out
#SBATCH --error=ectrans_benchmark.err
#SBATCH --nodes=2
#SBATCH --ntasks-per-node=16
#SBATCH --cpus-per-task=8
#SBATCH --gpus-per-node=0
#SBATCH --hint=nomultithread
#SBATCH --mem=200G

set -x
set -uo pipefail

BUILD_DIR=/path/to/your/ectrans/build/directory

source env.sh

OMP_PLACES=cores OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK srun --cpus-per-task=8 \
    ${BUILD_DIR}/bin/ectrans-benchmark-cpu-sp \
    --norms --truncation 1279 --niter 10 --nlev 137 --nfld 1 --uvders --scders --meminfo
```
and the env.sh file:
```bash
module load prgenv/intel
module load cmake/3.25.2
module load intel/2021.4.0
module load hpcx-openmpi/2.9.0
module load intel-mkl/19.0.5

# Setting required for bit reproducibility with Intel MKL:
export MKL_CBWR=AUTO,STRICT
```
This should be used to build ecTrans and FIAT.

I just ran this and it gave these GSTATS:
```
STATS FOR ALL TASKS
 NUM ROUTINE                                     CALLS      MEAN(ms)       MAX(ms)   FRAC(%)  UNBAL(%)
   0 PROGRAM        - Total                          1     26733.096     26733.302    100.00      0.00
   1 SETUP_TRANS0   - Setup ecTrans                  1       151.476       152.195      0.57      0.00
   2 SETUP_TRANS    - Setup ecTrans handle           1      2183.069      2188.173      8.19      0.02
   3 TIME STEP      - Time step                     10      1795.972      1795.973     67.18      0.00
   4 INV_TRANS      - Inverse transform             10      1144.960      1208.943     45.22      2.39
   5 DIR_TRANS      - Direct transform              10       621.733       686.826     25.69      2.43
   6 NORMS          - Norm comp. (optional)         10        29.273        69.366      2.59      1.50
 102 LTINV_CTL      - Inv. Legendre transform       10       213.846       222.374      8.32      0.32
 103 LTDIR_CTL      - Dir. Legendre transform       10       161.768       172.031      6.44      0.38
 106 FTDIR_CTL      - Dir. Fourier transform        10       108.572       127.428      4.77      0.71
 107 FTINV_CTL      - Inv. Fourier transform        10       265.114       321.063     12.01      2.09
 140 SULEG          - Comp. of Leg. poly.            1       942.594       962.715      3.60      0.08
 152 LTINV_CTL      - M to L transposition          10       136.868       145.347      5.44      0.32
 153 LTDIR_CTL      - L to M transposition          10        92.415       167.517      6.27      2.81
 157 FTINV_CTL      - L to G transposition          10       493.075       588.640     22.02      3.57
 158 FTDIR_CTL      - G to L transposition          10       242.663       343.109     12.83      3.76
 400 GSTATS         - GSTATS itself                759         0.052         0.052      0.15      0.00
TOTAL MEASURED IMBALANCE =       0.0 SECONDS,  0.0 PERCENT
TOTAL WALLCLOCK TIME      26.733 CPU TIME    185.764 VECTOR TIME     185.764
```

dhaumont:
I have added `#SBATCH --hint=nomultithread` and the results are much closer to yours:
```
STATS FOR ALL TASKS
 NUM ROUTINE                                     CALLS      MEAN(ms)       MAX(ms)   FRAC(%)  UNBAL(%)
   0 PROGRAM        - Total                          1     28265.767     28265.897    100.00      0.00
   1 SETUP_TRANS0   - Setup ecTrans                  1         2.280         2.508      0.01      0.00
   2 SETUP_TRANS    - Setup ecTrans handle           1      3796.866      3799.175     13.44      0.01
   3 TIME STEP      - Time step                     10      1822.906      1822.907     64.49      0.00
   4 INV_TRANS      - Inverse transform             10      1143.672      1218.013     43.09      2.63
   5 DIR_TRANS      - Direct transform              10       648.608       712.445     25.21      2.26
   6 NORMS          - Norm comp. (optional)         10        30.618        66.306      2.35      1.26
 102 LTINV_CTL      - Inv. Legendre transform       10       216.020       225.664      7.98      0.34
 103 LTDIR_CTL      - Dir. Legendre transform       10       169.981       179.704      6.36      0.34
 106 FTDIR_CTL      - Dir. Fourier transform        10       109.752       129.333      4.58      0.69
 107 FTINV_CTL      - Inv. Fourier transform        10       267.405       323.926     11.46      2.00
 140 SULEG          - Comp. of Leg. poly.            1       818.627       843.274      2.98      0.09
 152 LTINV_CTL      - M to L transposition          10       136.392       144.683      5.12      0.29
 153 LTDIR_CTL      - L to M transposition          10        99.193       190.205      6.73      3.22
 157 FTINV_CTL      - L to G transposition          10       488.803       595.269     21.06      3.77
 158 FTDIR_CTL      - G to L transposition          10       253.228       353.191     12.50      3.54
 400 GSTATS         - GSTATS itself                759         0.002         0.002      0.01      0.00
TOTAL MEASURED IMBALANCE =       0.0 SECONDS,  0.0 PERCENT
TOTAL WALLCLOCK TIME      28.266 CPU TIME    188.179 VECTOR TIME     188.179
```
But there is still no difference for the 3 different configurations:
```
develop_2.txt : number_of_nodes:2, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 28.266
TIME STEP 1822.906
INV_TRANS - 1143.672
DIR_TRANS - 648.608
SETUP_TRANS0 - 2.280
SETUP_TRANS - 3796.866


trltog_2.txt : number_of_nodes:2, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 28.149
TIME STEP 1805.970
INV_TRANS - 1136.188
DIR_TRANS - 641.104
SETUP_TRANS0 - 121.229
SETUP_TRANS - 3826.623


stack_2.txt : number_of_nodes:2, tasks_per_node: 16, cpus_per_task=8
------------
WALLCLOCK TIME 28.307
TIME STEP 1818.698
INV_TRANS - 1153.497
DIR_TRANS - 636.538
SETUP_TRANS0 - 20.694
SETUP_TRANS - 3900.775

```

samhatfield:
Yes much closer :) Your SETUP_TRANS is still a bit slow for some reason but the rest looks good. I'm not sure why your benchmark of trltog_clean is better than mine though. Could you share the stdout for that run?

dhaumont:
The complete results:
[develop_2.txt](https://github.com/user-attachments/files/24943300/develop_2.txt)
[develop_4.txt](https://github.com/user-attachments/files/24943301/develop_4.txt)
[stack_2.txt](https://github.com/user-attachments/files/24943302/stack_2.txt)
[stack_4.txt](https://github.com/user-attachments/files/24943304/stack_4.txt)
[trltog_2.txt](https://github.com/user-attachments/files/24943305/trltog_2.txt)
[trltog_4.txt](https://github.com/user-attachments/files/24943306/trltog_4.txt)

Indeed, `SETUP_TRANS` is a bit slower than your's, but consistent accross my different runs

samhatfield:
I think, but can't confirm, that the difference we are trying to measure (if indeed there is a difference) is subject to the node choice. Your cases all choose the same node allocation for a given number of nodes. Mine is randomly selected. It matters because we're trying to benchmark a communication-heavy section of the code. If I get unlucky and am allocated nodes on either ends of the topology, I suppose the benchmark will look unfavourable (far apart nodes take longer to communicate).

It's good that you get the same performance regardless of branch when fixing the nodelist. This gives me confidence that the changes here are performance-neutral. However it's still seems a strange coincidence that I measure a systematic but slight decrease in performance when node selection is randomised. I'll look into it a bit more.

samhatfield:
Disregard the above - my setup is _still_ slower than yours, even when running on exactly the same nodes ! ...

dhaumont:
Could it be that there was something wrong with your latest build of my branch ? Your timings are similar to the ones I  had before the introduction of  591f30d9d7c7abd8599ee671b7d821ff81b08f5d

samhatfield:
Yes I'm using 591f30d9d7c7abd8599ee671b7d821ff81b08f5d - it's printed by SETUP_TRANS0. We seem to have the same performance for compute regions but different for communication regions. For example your 153 L to M averages at 92.622 ms per call, whereas mine is 136.687. Similarly your 157 L to G is 481.613 ms whereas mine is 567.449. What modules do you load when building and running ecTrans / FIAT?

dhaumont:
I'm using fs-bundle, with the default settings for ATOS: https://git.ecmwf.int/users/masd/repos/ifs-bundle/browse/arch/ecmwf/hpc2020/intel/2021.4.0/hpcx-openmpi/2.9.0/env.sh?at=refs%2Fheads%2Fdevelop

```
# Load modules
module_load prgenv/intel
module_load intel/2021.4.0
module_load hpcx-openmpi/2.9.0
module_load intel-mkl/19.0.5

# Don't load these modules if env.sh is used as part of the IFS runtime environment - only the modules above are required
if [[ ${IFS_RUNTIME_ENV:-unset} == "unset" ]]; then
  module_load python3/3.10.10-01
  module_load fftw/3.3.9
  module_load netcdf4/4.9.2
  module_load hdf5/1.14.3
  module_load eigen/3.4.0
  module_load qhull/8.1-alpha1
  module_load cmake/3.25.2
  module_load ninja/1.10.0
  module_load fcm/2019.05.0
  module_load aec/1.1.2
fi

```

dhaumont:
@samhatfield I just shared the ectrans-benchmak repository with you (https://github.com/dhaumont/ectrans-benchmark).
It contains bash scripts reproducing all my steps (clone of repositories, compilation, allocation of resources, benchmarks and results extraction.
The  `README.md` file contains more detailed instructions. Don't hesitate if you have any questions.

samhatfield:
Thanks @dhaumont. Looking at your build scripts, I think the only meaningful difference between our two setups is the build type: I use BIT, you use RELEASE. If I repeat my benchmark above with RELEASE, I get the following:

number_of_nodes | tasks_per_node | cpus_per_task | truncation | inverse (develop) | direct (develop) | both (develop) | inverse (trltog_clean) | direct (trltog_clean) | both (trltog_clean) | inverse (ratio) | direct (ratio) | both (ratio)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 16 | 8 | 1279 | 2.4065 | 1.4681 | 3.8732 | 2.4496 | 1.4944 | 3.943 | 1.02 | 1.02 | 1.02
2 | 16 | 8 | 1279 | 1.1339 | 0.6444 | 1.7786 | 1.1349 | 0.6433 | 1.7777 | 1.00 | 1.00 | 1.00
4 | 16 | 8 | 1279 | 0.6077 | 0.3412 | 0.9483 | 0.6104 | 0.353 | 0.9655 | 1.00 | 1.03 | 1.02
8 | 16 | 8 | 1279 | 0.3467 | 0.2226 | 0.5696 | 0.3418 | 0.2308 | 0.577 | 0.99 | 1.04 | 1.01
16 | 16 | 8 | 1279 | 0.1898 | 0.1244 | 0.3138 | 0.1906 | 0.1302 | 0.3214 | 1.00 | 1.05 | 1.02
32 | 16 | 8 | 1279 | 0.1136 | 0.0673 | 0.1813 | 0.1188 | 0.0703 | 0.1903 | 1.05 | 1.04 | 1.05
64 | 16 | 8 | 1279 | 0.0678 | 0.0503 | 0.1179 | 0.0673 | 0.0472 | 0.1163 | 0.99 | 0.94 | 0.99

In other words, performance parity between develop and trltog_clean. I think we can merge this now.

samhatfield (approved):

samhatfield:
I'll leave the final approval and squash-merge for @wdeconinck.

wdeconinck (approved):
Thanks everyone involved in pushing this through! Huge effort but worth it!

dhaumont:
Thanks to both of you for the great support, it was much more difficult than anticipated :-) !

PR #318, opened 2025-10-17
MPI Split communicator.

mo-joshuacolclough:
Add tests to ensure ectrans/transi works with split MPI communicators. Requires no changes to ectrans, but does require an extension to transi to allow for setting of the fiat default MPL communicator.

Added:
- Test which performs a simple inv trans such as in the [usage example](https://sites.ecmwf.int/docs/ectrans/page/usage.html).
- `trans_set_mpi_comm` method to set the user communicator prior to setup of trans. Includes some error checking to ensure correct usage (prior to trans_init. And if it's not, ignore if it's called with the same communicator that has already been set).
- A simple test to ensure trans_set_mpi_comm behaves as expected in transi.


tests/trans/test_example.F90
```diff
@@ -0,0 +1,190 @@
...
+integer(kind=JPIM), parameter, dimension(2) :: truncations = [79, 188]
+
+! MODE
+integer(kind=JPIM), parameter, dimension(2) :: Ms = [1, 3]
+integer(kind=JPIM), parameter, dimension(2) :: Ns = [2, 4]
+
+
+integer(kind=JPIM) :: num_spectral_elements, num_grid_points
+integer(kind=JPIM) :: g_num_spectral_elements, g_num_grid_points  ! global
+integer(kind=JPIM) :: mode_index
```
mo-joshuacolclough:
```suggestion
integer(kind=JPIM) :: spectral_coefficient_index
```

tests/trans/test_example.F90
```diff
@@ -0,0 +1,190 @@
...
+real(kind=JPRM), allocatable :: g_grid_point_field(:)
+
+character(len=1024) :: filename
+
+
+call MPI_Init(ierror)
+call MPI_Comm_rank(MPI_COMM_WORLD, rank, ierror)
+
+split_colour = get_split_group()
+split_key = rank
```
mo-joshuacolclough:
```suggestion
split_key = rank     ! Global rank
```

tests/trans/test_example.F90
```diff
@@ -0,0 +1,190 @@
...
+call MPI_Comm_split(MPI_COMM_WORLD, split_colour, split_key, split_comm, ierror)
+
+print*, "=== Rank ", rank, ", Setup on group", split_colour, "==="
+
+! Set MPL comm
+LMPLUSERCOMM = .true.
```
mo-joshuacolclough:
```suggestion
! Essentially arguments to MPL_INIT here...
LMPLUSERCOMM = .true.
```

tests/trans/test_example.F90
```diff
@@ -0,0 +1,190 @@
...
+MPLUSERCOMM = split_comm
+call MPL_INIT()
+
+
+rank = MPL_MYRANK()
```
mo-joshuacolclough:
```suggestion
split_rank = MPL_MYRANK()
```

tests/trans/test_example.F90
```diff
@@ -0,0 +1,190 @@
...
+
+num_latitudes = 2*(truncation + 1)
+num_longitudes = num_latitudes*2
+print*, ">>> GLOBAL NUM LON =>", num_longitudes, "| LAT =>", num_latitudes
+call setup_trans(KSMAX=truncation, KDGL=num_latitudes)
+
+! Get function space sizes
+call trans_inq(KSPEC2=num_spectral_elements, KGPTOT=num_grid_points)
+call trans_inq(KSPEC2G=g_num_spectral_elements, KGPTOTG=g_num_grid_points)
+print*,"Num spec = ", num_spectral_elements, "| Num grid points = ", num_grid_points, g_num_grid_points
```
mo-joshuacolclough:
```suggestion
print*,"R", world_rank, "C", split_rank, "Num spec = ", num_spectral_elements, "| Num grid points = ", num_grid_points, g_num_grid_points
```

mo-joshuacolclough (commented):
(From discussion with @MarekWlasak)

tests/transi/transi_test_example.c
```diff
@@ -0,0 +1,340 @@
...
+
+      MPI_Comm_split(base_comm, colour, rank, &split_comm);
+    }
+
+    if (colour == 0) {
+      printf("WORLD_RANK %d, g%d => L\n", world_rank, colour);
+      test_gptosptogp(split_comm, 320,161,159);
+    } else if (colour == 1) {
+      printf("WORLD_RANK %d, g%d => S\n", world_rank, colour);
+      test_gptosptogp(split_comm, 320/2,161/2,159/2);
```
mo-joshuacolclough: TODO: Check output here.

src/transi/transi.h
```diff
@@ -166,6 +167,8 @@ int trans_use_mpi(_bool);
  */
 int trans_init(void);
 
+int trans_set_mpi_comm(const MPI_Fint mpi_user_comm);
```
wdeconinck: Please change this to use a standard int as to not `#include  <mpi.h>`
```suggestion
int trans_set_mpi_comm(int);
```

tests/trans/test_example.F90
```diff
@@ -0,0 +1,190 @@
+! (C) Crown Copyright 2025- Met Office.
+! 
+! This software is licensed under the terms of the Apache Licence Version 2.0
+! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+!
+
+program test_example
```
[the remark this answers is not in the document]
wdeconinck: Also the name of the file then.

wdeconinck (commented):
Nice progress on building up some test cases.
You can keep the transi developments here for now. I find it useful as a discussion point. We can always split the PR in pieces later.

tests/trans/test_multiple_res.F90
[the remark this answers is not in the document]
wdeconinck: Possibly it can be merged with the split_comm case later, where you would also have more than one handle available.

src/transi/transi_module.F90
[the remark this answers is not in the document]
wdeconinck: For now I like it here.

wdeconinck:
> Thanks @samhatfield - we are working off of the tagged release we are using currently, to avoid any potential complications with using develop in our codebase... (Happy to deal with the headaches at a later date... :) ) Just while we get things working. I don't intend for this to get merged, just for sharing at the moment.
> 
> We are pleased so far that the changes needed seem to be quite minimal!

I still believe forking from a develop branch will be less painful, especially because 1.6.2 is quite old. Already this (draft, granted) PR is not mergeable due to merge conflicts. Version 1.7.0 is now out. Perhaps you can test this tagged release in the stack; That feedback is also valuable.
And then if you can, rebase your changes either on 1.7.0 or develop.

tests/trans/test_multiple_res.F90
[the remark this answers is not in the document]
mo-joshuacolclough: I removed it but kept hold of it on a separate branch. It was more a simple test for our understanding of how KRESOL is handled... :)

mo-joshuacolclough:
Thanks @samhatfield @wdeconinck - I have merged develop into the branch to bring it up-to-date, and I've tidied the changes and used MPL where I can. Please can you take another look when you get the chance!

samhatfield:
Thanks @mo-joshuacolclough! Yes we will take a look soon.

wdeconinck:
Hi @mo-joshuacolclough this looks great to me now, and a good starting point to begin the work of making the MPI communicator configurable per trans handle.

wdeconinck (approved):

samhatfield:
Couple of early comments:
- If you want to link a target (such as the new transi tests) against `MPI::MPI_C` you'll need to add `C` to the `COMPONENTS` of MPI in the `REQUIRED_PACKAGES` argument of `ecbuild_add_option( FEATURE MPI ...`
- I *think* that the proper way to set a new default communicator with MPL is through [`MPL_SETDFLT_COMM`](https://github.com/ecmwf-ifs/fiat/blob/develop/src/fiat/mpl/internal_f08/mpl_setdflt_comm_mod.F90). This means that instead of setting `LMPLUSERCOMM` and `MPLUSERCOMM`, you should simply call `MPL_SETDFLT_COMM(MPI_USER_COMM, DUMMY_COMM)`. This subroutine always returns the last default communicator as the second argument, but you can just pass a dummy argument and then ignore it. This affects `trans_set_mpi_comm` but also `test_split_mpi_comm`.
- I think `test_split_mpi_comm` can be written without any explicit MPI calls and this is a strong preference for us. Going forward we want all MPI functionality to be wrapped under the MPL layer. For example:
  ```diff
  -call MPI_Init(ierror)
  -call MPI_Comm_rank(MPI_COMM_WORLD, world_rank, ierror)
  -call MPI_Comm_size(MPI_COMM_WORLD, world_num_ranks, ierror)
  +CALL MPL_INIT()
  +WORLD_NUM_RANKS = MPL_NPROC()

   split_colour = get_split_group()
  -split_key = world_rank    ! Key used here is the WORLD rank.
  -call MPI_Comm_split(MPI_COMM_WORLD, split_colour, split_key, split_comm, ierror)
  +SPLIT_KEY = MPL_MYRANK() ! Key used here is the WORLD rank.

  -print*,"=== Rank ", world_rank, ", Setup on group", split_colour, "==="
  +PRINT *, "=== Rank ", SPLIT_KEY, ", Setup on group", SPLIT_COLOUR, "==="

  -! Set MPL comm. Global variables from fiat MPL_DATA_MODULE.
  -! Acts like arguments to MPL_INIT.
  -LMPLUSERCOMM = .true.
  -MPLUSERCOMM = split_comm
  -call MPL_INIT()
  +CALL MPL_COMM_SPLIT(MPL_COMM, SPLIT_COLOUR, SPLIT_KEY, SPLIT_COMM, IERROR)
  +CALL MPL_SETDFLT_COMM(SPLIT_COMM, DUMMY_COMM)
  ```
  This requires a bugfix PR to FIAT to be merged: https://github.com/ecmwf-ifs/fiat/pull/72.
  Note that the call to `MPI_FINALIZE` is also unnecessary.
- `get_split_group` can be modified to avoid MPI calls as well, just making sure not to add +1 when calculating `rank_ratio`. Though this function is probably unnecessary; you could just say `SPLIT_COLOUR = MERGE(0, 1, MPL_MYRANK() == 1)` or something.
- Style: we really have a messy situation with some upper, some lower case Fortran. transi sort of follows a more modern style, but the core library is still shouty Fortran, so I would say that test_split_mpi_comm.F90 should also be entirely uppercase.
- As for the transi test, there are again many calls to raw MPI, but I'm not sure what the preferred style is there. @wdeconinck, MPL has no C bindings right, so how should people use transi in an MPI environment?

samhatfield:
BTW I think I planted the seed of the `LMPLUSERCOMM` style in one of our meetings - if so sorry about that 😆 I literally only just discovered that `MPL_SETDFLT_COMM` exists 5 minutes ago.

mo-joshuacolclough:
Thanks Sam, can confirm that the test works with your changes - 0e68acf. Now updating the transi code.

wdeconinck:
I'm of the opinion that whatever can be done with MPL is good to do with MPL especially in the Fortran tests.

However! ectrans should also be able to be used within codes that don't use MPL and be able to be set up with regular MPI communicators and be interoperable with MPI routines and MPI comms. MPL should not be the ground truth for how MPI is run in packages that use ectrans. For this reason, I found that also previous implementation using MPI in the driver layer in Fortran was OK in my opinion.

For transi, the use of MPL is really an implementation detail of ectrans, and standard MPI calls are advised. In the context of Atlas these mpi communicators are managed and split via eckit.
transi library itself should not explicitly link with MPI::MPI_C, but unit tests that set up transi for MPI should be able to link with MPI::MPI_C indeed.

tests/trans/test_split_mpi_comm.F90
mo-joshuacolclough, line 1: Happy to update this all to upper case if that is the consensus? Cheers
@samhatfield @wdeconinck

CMakeLists.txt
```diff
@@ -41,9 +41,10 @@ ecbuild_enable_fortran( REQUIRED NO_MODULE_DIRECTORY )
 
 ecbuild_find_package( NAME fiat REQUIRED )
 
-# Inherit MPI feature from FIAT (if you don't want MPI, rebuild FIAT with ENABLE_MPI=OFF)
-set( HAVE_MPI ${fiat_HAVE_MPI} )
-set( ectrans_HAVE_MPI ${fiat_HAVE_MPI} )
+ecbuild_add_option( FEATURE MPI
+                    DESCRIPTION "Support for MPI distributed memory parallelism"
+                    REQUIRED_PACKAGES "MPI COMPONENTS C Fortran CXX"
```
mo-joshuacolclough: NOTE I needed to revert this cmake in order to include C MPI for the transi test (fiat includes `Fortran` only).

mo-joshuacolclough:
Turning this back into a draft PR... My Atlas update is being problematic. Bear with me!

tests/trans/test_split_mpi_comm.F90
samhatfield: That would be the "least messy" solution, thanks. But feel free to do that at the end before the final merge.

mo-joshuacolclough: Done, cheers

src/transi/transi_module.F90
```diff
@@ -658,7 +661,11 @@ function trans_set_mpi_comm(mpi_user_comm) bind(C,name="trans_set_mpi_comm") res
   if (is_init .and. mpi_user_comm == last_comm_set) return
 
   ! Confirm that this is called prior to MPL_INIT, to ensure correct setting of global vars.
-  if (is_init) then
+  !
+  ! If it is the case that trans_init has been called prior, then it will have been setup with
+  ! MPI_COMM_WORLD. So therefore in the special case that the mpi_user_comm matches WORLD,
+  ! then there is no issue.
+  if (is_init .and. mpi_user_comm /= MPI_COMM_WORLD) then
```
mo-joshuacolclough: Added this small amendment to allow current usage of `trans_init` in Atlas. Might there be issues with using WORLD here, or is this ok? @wdeconinck @samhatfield 

If so I can instead undo this and enforce that users don't call set_comm unless actually needing to set a communicator.

mo-joshuacolclough:
> Couple of early comments:
> 
>     * If you want to link a target (such as the new transi tests) against `MPI::MPI_C` you'll need to add `C` to the `COMPONENTS` of MPI in the `REQUIRED_PACKAGES` argument of `ecbuild_add_option( FEATURE MPI ...`
> 
>     * I _think_ that the proper way to set a new default communicator with MPL is through [`MPL_SETDFLT_COMM`](https://github.com/ecmwf-ifs/fiat/blob/develop/src/fiat/mpl/internal_f08/mpl_setdflt_comm_mod.F90). This means that instead of setting `LMPLUSERCOMM` and `MPLUSERCOMM`, you should simply call `MPL_SETDFLT_COMM(MPI_USER_COMM, DUMMY_COMM)`. This subroutine always returns the last default communicator as the second argument, but you can just pass a dummy argument and then ignore it. This affects `trans_set_mpi_comm` but also `test_split_mpi_comm`.
> 
>     * I think `test_split_mpi_comm` can be written without any explicit MPI calls and this is a strong preference for us. Going forward we want all MPI functionality to be wrapped under the MPL layer. For example:
>       ```diff
>       -call MPI_Init(ierror)
>       -call MPI_Comm_rank(MPI_COMM_WORLD, world_rank, ierror)
>       -call MPI_Comm_size(MPI_COMM_WORLD, world_num_ranks, ierror)
>       +CALL MPL_INIT()
>       +WORLD_NUM_RANKS = MPL_NPROC()
>       
>        split_colour = get_split_group()
>       -split_key = world_rank    ! Key used here is the WORLD rank.
>       -call MPI_Comm_split(MPI_COMM_WORLD, split_colour, split_key, split_comm, ierror)
>       +SPLIT_KEY = MPL_MYRANK() ! Key used here is the WORLD rank.
>       
>       -print*,"=== Rank ", world_rank, ", Setup on group", split_colour, "==="
>       +PRINT *, "=== Rank ", SPLIT_KEY, ", Setup on group", SPLIT_COLOUR, "==="
>       
>       -! Set MPL comm. Global variables from fiat MPL_DATA_MODULE.
>       -! Acts like arguments to MPL_INIT.
>       -LMPLUSERCOMM = .true.
>       -MPLUSERCOMM = split_comm
>       -call MPL_INIT()
>       +CALL MPL_COMM_SPLIT(MPL_COMM, SPLIT_COLOUR, SPLIT_KEY, SPLIT_COMM, IERROR)
>       +CALL MPL_SETDFLT_COMM(SPLIT_COMM, DUMMY_COMM)
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       This requires a bugfix PR to FIAT to be merged: [Make sure MPL_RANK is updated with new comm fiat#72](https://github.com/ecmwf-ifs/fiat/pull/72).
>       Note that the call to `MPI_FINALIZE` is also unnecessary.
> 
>     * `get_split_group` can be modified to avoid MPI calls as well, just making sure not to add +1 when calculating `rank_ratio`. Though this function is probably unnecessary; you could just say `SPLIT_COLOUR = MERGE(0, 1, MPL_MYRANK() == 1)` or something.
> 
>     * Style: we really have a messy situation with some upper, some lower case Fortran. transi sort of follows a more modern style, but the core library is still shouty Fortran, so I would say that test_split_mpi_comm.F90 should also be entirely uppercase.
> 
>     * As for the transi test, there are again many calls to raw MPI, but I'm not sure what the preferred style is there. @wdeconinck, MPL has no C bindings right, so how should people use transi in an MPI environment?

I am hoping my changes have now addressed these problems. @samhatfield  Please can you have another look and let me know. Cheers!

EDIT(24/02/26): I have missed `MPL_SETDFLT_COMM` in transi_module.F90. I am trying it now, but getting a strange error..

```
188: At line 79 of file /home/users/joshua.colclough/lj/ectrans__120226/fiat/src/fiat/mpl/internal_f08/mpl_setdflt_comm_mod.F90
188: Fortran runtime error: Index '1' of dimension 1 of array 'mpl_comm_oml' above upper bound of 0
```

Currently looking at this. @samhatfield

src/transi/transi.h
```diff
@@ -189,6 +189,8 @@ int trans_use_mpi(_bool);
  */
 int trans_init(void);
 
+int trans_set_mpi_comm(const int mpi_user_comm);
```
wdeconinck:
```suggestion
int trans_set_mpi_comm(int mpi_user_comm);
```
No need for an argument passed by value to be const.

src/transi/transi_module.F90
wdeconinck: We would not want to `use MPL_MPIF`  and `MPI_COMM_WORLD` directly. Could we avoid it?
Check following logic, and perhaps adapt.
```Fortran
if (MPL_NUMPROC == -1) then
    ! MPL is not yet initialised
else 
    ! MPL is already initialised
    if (mpl_user_comm /= MPL_COMM) then
       ! ERROR: MPL was already initialised with a different communicator than the requested one.
   endif
endif
```

I suppose you would need to do the actual checking within `trans_init()` above, so that you could exchange the order of `trans_use_mpi` and `trans_set_mpi_comm`.

wdeconinck (changes requested):
Hi @mo-joshuacolclough good work! I have a few requested changes.

mo-joshuacolclough:
> > Couple of early comments:
> > ```
> > * If you want to link a target (such as the new transi tests) against `MPI::MPI_C` you'll need to add `C` to the `COMPONENTS` of MPI in the `REQUIRED_PACKAGES` argument of `ecbuild_add_option( FEATURE MPI ...`
> > 
> > * I _think_ that the proper way to set a new default communicator with MPL is through [`MPL_SETDFLT_COMM`](https://github.com/ecmwf-ifs/fiat/blob/develop/src/fiat/mpl/internal_f08/mpl_setdflt_comm_mod.F90). This means that instead of setting `LMPLUSERCOMM` and `MPLUSERCOMM`, you should simply call `MPL_SETDFLT_COMM(MPI_USER_COMM, DUMMY_COMM)`. This subroutine always returns the last default communicator as the second argument, but you can just pass a dummy argument and then ignore it. This affects `trans_set_mpi_comm` but also `test_split_mpi_comm`.
> > 
> > * I think `test_split_mpi_comm` can be written without any explicit MPI calls and this is a strong preference for us. Going forward we want all MPI functionality to be wrapped under the MPL layer. For example:
> >   ```diff
> >   -call MPI_Init(ierror)
> >   -call MPI_Comm_rank(MPI_COMM_WORLD, world_rank, ierror)
> >   -call MPI_Comm_size(MPI_COMM_WORLD, world_num_ranks, ierror)
> >   +CALL MPL_INIT()
> >   +WORLD_NUM_RANKS = MPL_NPROC()
> >   
> >    split_colour = get_split_group()
> >   -split_key = world_rank    ! Key used here is the WORLD rank.
> >   -call MPI_Comm_split(MPI_COMM_WORLD, split_colour, split_key, split_comm, ierror)
> >   +SPLIT_KEY = MPL_MYRANK() ! Key used here is the WORLD rank.
> >   
> >   -print*,"=== Rank ", world_rank, ", Setup on group", split_colour, "==="
> >   +PRINT *, "=== Rank ", SPLIT_KEY, ", Setup on group", SPLIT_COLOUR, "==="
> >   
> >   -! Set MPL comm. Global variables from fiat MPL_DATA_MODULE.
> >   -! Acts like arguments to MPL_INIT.
> >   -LMPLUSERCOMM = .true.
> >   -MPLUSERCOMM = split_comm
> >   -call MPL_INIT()
> >   +CALL MPL_COMM_SPLIT(MPL_COMM, SPLIT_COLOUR, SPLIT_KEY, SPLIT_COMM, IERROR)
> >   +CALL MPL_SETDFLT_COMM(SPLIT_COMM, DUMMY_COMM)
> >   ```
> >   
> >   
> >       
> >         
> >       
> >   
> >         
> >       
> >   
> >       
> >     
> >   This requires a bugfix PR to FIAT to be merged: [Make sure MPL_RANK is updated with new comm fiat#72](https://github.com/ecmwf-ifs/fiat/pull/72).
> >   Note that the call to `MPI_FINALIZE` is also unnecessary.
> > 
> > * `get_split_group` can be modified to avoid MPI calls as well, just making sure not to add +1 when calculating `rank_ratio`. Though this function is probably unnecessary; you could just say `SPLIT_COLOUR = MERGE(0, 1, MPL_MYRANK() == 1)` or something.
> > 
> > * Style: we really have a messy situation with some upper, some lower case Fortran. transi sort of follows a more modern style, but the core library is still shouty Fortran, so I would say that test_split_mpi_comm.F90 should also be entirely uppercase.
> > 
> > * As for the transi test, there are again many calls to raw MPI, but I'm not sure what the preferred style is there. @wdeconinck, MPL has no C bindings right, so how should people use transi in an MPI environment?
> > ```
> 
> I am hoping my changes have now addressed these problems. @samhatfield Please can you have another look and let me know. Cheers!
> 
> EDIT(24/02/26): I have missed `MPL_SETDFLT_COMM` in transi_module.F90. I am trying it now, but getting a strange error..
> 
> ```
> 188: At line 79 of file /home/users/joshua.colclough/lj/ectrans__120226/fiat/src/fiat/mpl/internal_f08/mpl_setdflt_comm_mod.F90
> 188: Fortran runtime error: Index '1' of dimension 1 of array 'mpl_comm_oml' above upper bound of 0
> ```
> 
> Currently looking at this. @samhatfield

I've figured I need a `CALL MPL_INIT` prior to setting the comm, which fixes the ectrans test but breaks things in Atlas... :/

src/transi/transi.h
mo-joshuacolclough: Thanks done in 5165d1d

src/transi/transi_module.F90
mo-joshuacolclough: Thanks I've updated this now. The logic used to be that we had to ensure MPL_INIT was not called, because it used global variables set. Now I have switched to `MPL_SETDFLT_COMM` it means that MPL may or may not be setup at this point.
What is now important is if trans has been setup yet, and if so what communicator that has been done with (which will correspond with MPL_COMM).

mo-joshuacolclough:
@wdeconinck I'm running this with atlas `develop` and getting a floating point exception in trans. Wondering if you had any thoughts what could be happening? https://github.com/ecmwf/atlas/pull/346#issuecomment-3959178050

With `ctest -R atlas_test_trans$ -VV`.

```
Running case 13: test_trans_VorDivToUV ...
truncation = 1
vor:
0 0 1 0 0 0

[0] =========================================
[0] Floating point exception [FE_INVALID] (signal intercepted by atlas)
[0] -----------------------------------------
[0] BACKTRACE
[0] -----------------------------------------
[0] backtrace [5] stack has 27 addresses
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/libeckit.so+eckit::BackTrace::dump[abi:cxx11]())0x48
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/libatlas.so.0.45+atlas::backtrace[abi:cxx11]())0x18
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/libatlas.so.0.45+atlas::library::atlas_signal_handler(int, siginfo_t*, void*))0x1e9
[0] (/lib64/libc.so.6+)0x3fc30
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/../lib64/libectrans_dp.so+__vd2uv_mod_dp_MOD_vd2uv)0x231d
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/../lib64/libectrans_dp.so+)0x1ac283
[0] (/data/apps/spack/compilers/install/linux-rhel9-zen2/gcc-11.5.0/gcc-12.2.0-wwyf7gjcun7i6usswde5fnjsyfpsph33/lib64/libgomp.so.1+GOMP_parallel)0x46
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/../lib64/libectrans_dp.so+__vd2uv_ctl_mod_dp_MOD_vd2uv_ctl)0x3db
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/../lib64/libectrans_dp.so+vordiv_to_uv_dp_)0x2d09
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/libtransi_dp.so+trans_vordiv_to_UV)0x890
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/libatlas.so.0.45+atlas::trans::VorDivToUVIFS::execute(int, int, double const*, double const*, double*, double*, eckit::Configuration const&) const)0x8e
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/lib/libatlas.so.0.45+atlas::trans::VorDivToUV::execute(int, int, double const*, double const*, double*, double*, eckit::Configuration const&) const)0x5c
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)
[0] (/lib64/libc.so.6+)0x2a610
[0] (/lib64/libc.so.6+__libc_start_main)0x80
[0] (/home/users/joshua.colclough/lj/build-ectrans_atdev__230226-Debug/atlas/src/tests/trans/atlas_test_trans)

[0] end of backtrace dump ...
[0] =========================================
```

wdeconinck:
I will try to reproduce

CMakeLists.txt
samhatfield: I guess it could be added below instead, right @wdeconinck? I mean
```cmake
ecbuild_add_option( FEATURE TRANSI
                    DEFAULT ON
                    DESCRIPTION "Compile TransI C-interface to trans"
                    CONDITION HAVE_DOUBLE_PRECISION AND HAVE_CPU
                    REQUIRED_PACKAGES "MPI COMPONENTS C" )
```

src/transi/transi_module.F90
```diff
@@ -640,6 +649,41 @@ function trans_init() bind(C,name="trans_init") result(iret)
...
+      call MPL_INIT(KOUTPUT=0,KUNIT=trans_out,LDINFO=.False.)
+    end if
+
+    call MPL_SETDFLT_COMM(mpi_user_comm, dummy_comm)
+
+  ! Trans already initialised. If it has already been setup with the requested communicator
+  ! then there is no issue. Otherwise, the user is attempting to change the comm
+  ! mid-run which is not supported.
+  ! NOTE that MPL uses fortran indexing for MPI communicator numbers.
+  else if (mpi_user_comm + 1 /= MPL_COMM) then
```
samhatfield: I don't think that's correct. Rank indices are 1-indexed for MPL, so correspond to the MPI rank plus one, but MPL communicators seem to be identical to MPI communicators. At least, when I read [example MPL routines](https://github.com/ecmwf-ifs/fiat/blob/main/src/fiat/mpl/internal/mpl_setdflt_comm_mod.F90), I don't see that one is subtracted when they're passed to raw MPI routines.

tests/trans/test_split_mpi_comm.F90
```diff
@@ -0,0 +1,186 @@
...
+  DISPLS = 0
+  DO I=2, SPLIT_NUM_RANKS
+    DISPLS(I) = DISPLS(I - 1) + GRID_PARTITION_SIZES(I - 1)
+  END DO
+END IF
+
+CALL MPL_GATHERV(GRID_POINT_FIELD(:,1,1), PRECVBUF=G_GRID_POINT_FIELD, &
+                 KCOMM=SPLIT_COMM, &
+                 KRECVCOUNTS=GRID_PARTITION_SIZES, &
+                 KRECVDISPL=DISPLS)
```
samhatfield: Might it be easier to rely on [`GATH_GRID`](https://sites.ecmwf.int/docs/ectrans/page/api.html#GATH_GRID) here? I'm happy to do this.

samhatfield:
Looking good @mo-joshuacolclough! Just a few comments so far, but I'll continue reviewing tomorrow.

tests/trans/test_split_mpi_comm.F90
mo-joshuacolclough: Please that would be fantastic if you would be happy to, thanks. Will take me a lot longer...

src/transi/transi_module.F90
mo-joshuacolclough: I see. I was going to ask about this actually. Issues were happening where the comm was 1 index off, adding this fixed it. I will revert and have a closer look tomorrow.

wdeconinck:
@mo-joshuacolclough I could reproduce the SIGFPE and created a PR with the fix #372

src/transi/transi_module.F90
samhatfield: Yes, please do. I might be wrong!

tests/trans/test_split_mpi_comm.F90
samhatfield: There ya go.

src/transi/transi_module.F90
mo-joshuacolclough: It seems what is happening is that `MPL_SETDFLT_COMM` does not update the `MPL_COMM` value (which remains equal to `4`, whilst the mpi_user_comm is `3`).

mo-joshuacolclough: Setting MPL_COMM in fiat `MPL_SETDFLT_COMM` seems to fix things... See https://github.com/ecmwf-ifs/fiat/compare/develop...mo-joshuacolclough:fiat:feature/mpl_comm_setting. Not sure if this is the right thing to do?

mo-joshuacolclough:
> @mo-joshuacolclough I could reproduce the SIGFPE and created a PR with the fix #372

Thank you for the fix.

src/transi/transi_module.F90
wdeconinck: I would be happier if we did not change that because that seems to be a thread-safe routine. We would not want to have these modifications being called within a threaded region.
It seems that we should probably never reference `MPL_COMM`, but rather `MPL_COMM_OML(1)`

samhatfield: What's the actual point of `MPL_COMM`? Is it leftover dead code from a previous non-thread-safe MPL?

Rather than getting `MPL_COMM` (or `MPL_COMM_OML` or whatever) directly from Fiat, what about calling `MPL_COMM_COMPARE`? https://github.com/ecmwf-ifs/fiat/blob/main/src/fiat/mpl/internal/mpl_comm_compare_mod.F90

samhatfield: Nevermind. Subroutine above still requires you to pass both communicators.

samhatfield: Do we need a new subroutine like `MPL_GETDFLT_COMM`?

mo-joshuacolclough: I see. Thanks again for the help on this. Should it not be:

```Fotran
ITID = OML_MY_THREAD()
local_comm = MPL_COMM_OML(ITID)
```

wdeconinck: I think that would be equivalent as we're probably not in an openmp-threaded region, but surely good idea.

wdeconinck: > What's the actual point of MPL_COMM?

Perhaps it would be good to check use cases in the entire stack... 
In ifs-source in a simple superficial check I didn't find any.

In simple applications, tools, without specialised setup, using `MPL_COMM` instead of `MPL_COMM_OML(1)` reads a bit nicer I suppose.

samhatfield: Yes, can you try that @mo-joshuacolclough?
```fortran
else if (mpi_user_comm /= MPL_COMM_OML(OML_MY_THREAD())) then
```

mo-joshuacolclough: Working! 22b6b75

src/transi/transi_module.F90
```diff
@@ -640,6 +653,40 @@ function trans_init() bind(C,name="trans_init") result(iret)
...
+  !
+  ! If it is the case that trans_init has been called prior, ensure the comm here is the same
+  ! as what has been setup previously.
+  if (.not. is_init) then
+    ! MPL not yet initialised.
+    if (MPL_NUMPROC == -1) then
+      call MPL_INIT(KOUTPUT=0,KUNIT=trans_out,LDINFO=.False.)
+    end if
+
+    call MPL_SETDFLT_COMM(mpi_user_comm, dummy_comm)
```
wdeconinck:
```suggestion
    if (MPL_NUMPROC == -1) then
      LMPLUSERCOMM = .TRUE.           ! defined in MPL_MODULE
      MPLUSERCOMM = mpi_user_comm     ! defined in MPL_MODULE
      call MPL_INIT(KOUTPUT=0,KUNIT=trans_out,LDINFO=.False.)
    else
      call MPL_SETDFLT_COMM(mpi_user_comm, dummy_comm)
    end if
```

wdeconinck:
```suggestion
    if (MPL_NUMPROC == -1) then
      ! Set LMPLUSERCOMM and MPLUSERCOMM to be used in MPL_INIT when trans_init() is called
      LMPLUSERCOMM = .TRUE.           ! defined in MPL_MODULE
      MPLUSERCOMM = mpi_user_comm     ! defined in MPL_MODULE
    else
      call MPL_SETDFLT_COMM(mpi_user_comm, dummy_comm)
    end if
```

src/transi/transi_module.F90
```diff
@@ -640,6 +653,40 @@ function trans_init() bind(C,name="trans_init") result(iret)
...
+
+  ! Trans already initialised. If it has already been setup with the requested communicator
+  ! then there is no issue. Otherwise, the user is attempting to change the comm
+  ! mid-run which is not supported.
+  else if (mpi_user_comm /= MPL_COMM_OML(OML_MY_THREAD())) then
```
wdeconinck: You should make sure that MPL_COMM_OML is sized appropriately to avoid a SEGV.

wdeconinck (changes requested):
I will have a further look on the CMake configuration with MPI C. In the mean time, I think set_mpi_comm should not initialise MPL in case it was not done so.
Could you check my following suggestions are working as well?

wdeconinck:
The test ectrans_test_split_mpi_comm  is failing for the nvhpc GitHub runner with not enough slots available:
```
243/421 Test #243: ectrans_test_split_mpi_comm ......................................................***Failed    0.04 sec
  --------------------------------------------------------------------------
  There are not enough slots available in the system to satisfy the 4
  slots that were requested by the application:
  
    /home/runner/work/ectrans/ectrans/build/tests/ectrans_test_split_mpi_comm
  
  Either request fewer slots for your application, or make more slots
  available for use.
  
  A "slot" is the Open MPI term for an allocatable unit where we can
  launch a process.  The number of slots available are defined by the
  environment in which Open MPI processes are run:
  
    1. Hostfile, via "slots=N" clauses (N defaults to number of
       processor cores if not provided)
    2. The --host command line parameter, via a ":N" suffix on the
       hostname (N defaults to 1 if not provided)
    3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
    4. If none of a hostfile, the --host command line parameter, or an
       RM is present, Open MPI defaults to the number of processor cores
  
  In all the above cases, if you want Open MPI to default to the number
  of hardware threads instead of the number of processor cores, use the
  --use-hwthread-cpus option.
  
  Alternatively, you can use the --oversubscribe option to ignore the
  number of available slots when deciding the number of processes to
  launch.
  --------------------------------------------------------------------------
  ```

File: CMakeLists.txt (hunk 9 above)
Comment (wdeconinck) [reply CMakeLists.txt]: transi itself does not require MPI, it relies strictly on MPL which is compiled with Fortran MPI.

It is only the tests that splits the MPI comm that requires MPI directly.
I have just pushed 18d33ebc that add the MPI C component only in this test which requires it.

**Comment by samhatfield** [comment]
> The test ectrans_test_split_mpi_comm is failing for the nvhpc GitHub runner with not enough slots available:
> 
> ```
> 243/421 Test #243: ectrans_test_split_mpi_comm ......................................................***Failed    0.04 sec
>   --------------------------------------------------------------------------
>   There are not enough slots available in the system to satisfy the 4
>   slots that were requested by the application:
>   
>     /home/runner/work/ectrans/ectrans/build/tests/ectrans_test_split_mpi_comm
>   
>   Either request fewer slots for your application, or make more slots
>   available for use.
>   
>   A "slot" is the Open MPI term for an allocatable unit where we can
>   launch a process.  The number of slots available are defined by the
>   environment in which Open MPI processes are run:
>   
>     1. Hostfile, via "slots=N" clauses (N defaults to number of
>        processor cores if not provided)
>     2. The --host command line parameter, via a ":N" suffix on the
>        hostname (N defaults to 1 if not provided)
>     3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
>     4. If none of a hostfile, the --host command line parameter, or an
>        RM is present, Open MPI defaults to the number of processor cores
>   
>   In all the above cases, if you want Open MPI to default to the number
>   of hardware threads instead of the number of processor cores, use the
>   --use-hwthread-cpus option.
>   
>   Alternatively, you can use the --oversubscribe option to ignore the
>   number of available slots when deciding the number of processes to
>   launch.
>   --------------------------------------------------------------------------
> ```

I made sure 4 tasks are requested on AC but it didn't help. Not sure why.

**Comment by wdeconinck** [comment]
It's not the hpc runner, but the github runner , just like macOS.
Either we need to add the --oversubscribe to the nvhpc build as well, or we reduce the number of required MPI ranks in the test.

File: src/transi/transi_module.F90 (hunk 14 above)
Comment (mo-joshuacolclough) [reply src/transi/transi_module.F90]: Thanks done in a6a11eb

File: src/transi/transi_module.F90 (hunk 15 above)
Comment (mo-joshuacolclough) [reply src/transi/transi_module.F90]: Thanks I've added some logic. 410d536. Does that seem ok?

File: src/transi/transi_module.F90 (hunk 14 above)
Comment (wdeconinck) [reply src/transi/transi_module.F90]: Great! Thanks!

File: src/transi/transi_module.F90 (hunk 15 above)
Comment (wdeconinck) [reply src/transi/transi_module.F90]: Looks OK! thanks!

**Comment by wdeconinck** [comment]
It seems indeed the hpc runners as well. As you know we cannot change the hpc workflows in pull requests. First we need to update the workflow in develop and then merge it here.

**Review by wdeconinck — APPROVED** [review]
All in order for me. 
Thanks @mo-joshuacolclough !

Please do a final merge of develop and check with atlas.

**Comment by samhatfield** [comment]
> It seems indeed the hpc runners as well. As you know we cannot change the hpc workflows in pull requests. First we need to update the workflow in develop and then merge it here.

I always forget that!

The LUMI CI seems to actually be passing. I checked the action logs manually on LUMI. I think the pipe just got interrupted for some reason.

**Comment by mo-joshuacolclough** [comment]
Thanks both - updated the branch, I'm just taking a look at an issue with `atlas_test_trans` on my custom_comm branch.

**Comment by mo-joshuacolclough** [comment]
@wdeconinck I have tested that this passes with Atlas develop, and fixed the feature/custom_comm Atlas branch.

**Review by wdeconinck — APPROVED** [review]

File: tests/CMakeLists.txt [hunk 16]
```diff
@@ -297,6 +297,16 @@ if( HAVE_ETRANS )
 
 endif()
 
+if( HAVE_CPU AND HAVE_MPI )
```
samhatfield, line 313:
```suggestion
# --------------------------------------------------------------------------------------------------
# Add test for split-communicator compatibility of ecTrans
# --------------------------------------------------------------------------------------------------

if( HAVE_CPU AND HAVE_MPI )
```

tests/CMakeLists.txt
```diff
@@ -297,6 +297,20 @@ if( HAVE_ETRANS )
...
+
+if( HAVE_CPU AND HAVE_MPI )
+  ecbuild_add_test(TARGET ectrans_test_split_mpi_comm
+    SOURCES trans/test_split_mpi_comm.F90
+    LIBS trans_sp parkind_sp
+    LINKER_LANGUAGE Fortran
+    MPI 4
+    OMP 1
+  )
+endif()
```
samhatfield, line 321: This whole code block needs to be protected with an `if( HAVE_SINGLE_PRECISION )` as otherwise we can't guarantee the `trans_sp` library is available.

Any reason you wrote this test only for single precision @mo-joshuacolclough?

mo-joshuacolclough: My mistake, should I change it to `LIBS trans_dp parkind_dp`?

samhatfield: `DOUBLE_PRECISION` is also a feature so can be switched off. That means we can't guarantee `trans_dp` is available either.

Since this test is precision-agnostic I suggest we just pick SP if only SP is turned on, DP if only DP is turned on, and SP if both are turned on (totally arbitrary).

```suggestion
# It doesn't really matter which precision we use for this test, as long as we use one that's
# actually available
if( HAVE_SINGLE_PRECISION )
  set( trans_lib trans_sp )
  set( parkind_lib parkind_sp )
else()
  set( trans_lib trans_dp )
  set( parkind_lib parkind_dp )
endif()
if( HAVE_CPU AND HAVE_MPI )
  ecbuild_add_test(TARGET ectrans_test_split_mpi_comm
    SOURCES trans/test_split_mpi_comm.F90
    LIBS ${trans_lib} ${parkind_lib}
    LINKER_LANGUAGE Fortran
    MPI 4
    OMP 1
  )
endif()
```
Also, in `TEST_SPLIT_MPI_COMM`, you need to find-replace all `JPRM` (single precision) with `JPRB` (working precision).

mo-joshuacolclough: Thank you for the pointer, I've commited this & updated the types in the test.

samhatfield:
Congratulations @mo-joshuacolclough! Thanks for your patience during the review. Perhaps we should have a catchup call sometime?

PR #319, opened 2025-10-20
Disable FPEs for the Fourier transform

samhatfield:
This is to prevent spurious FPEs from occurring when making use of vector instructions.


marsdeno (approved):
Can't test right now, but looks ok. 
Only question on my side would be, given expected the use case, would this not also be desirable for BLAS calls?

samhatfield:
Following external discussions we've decided not to merge this.

PR #320, opened 2025-10-21
Tidy SUPOLF

samhatfield:
I've started to tidy up and document SUPOLF. This subroutine is used to generate the associated Legendre polynomials at a particular latitude and zonal wavenumber. Documentation here: https://sites.ecmwf.int/docs/ectrans/page/design/legendre_polynomials.html.

### DIFF
--- a/src/trans/common/internal/supolf_mod.F90
+++ b/src/trans/common/internal/supolf_mod.F90
@@ -94,6 +94,10 @@ DDL(K) = REAL(2 * K * (K + 1) - 2 * KM**2 - 1, JPRD) / REAL((2 * K - 1) * (2 * K
 ZEPS  = EPSILON(ZSCALE)
 ICORR3 = 0
 
+! This parameter determines which polynomials are computed
+! ICHEAP = 1 : all polynomials
+! ICHEAP = 2 : only even polynomials
+! ICHEAP = 3 : only odd polynomials
 ICHEAP = 1
 IF (PRESENT(KCHEAP)) THEN
   ICHEAP = KCHEAP


src/trans/common/internal/supolf_mod.F90
```diff
@@ -221,7 +222,7 @@ SUBROUTINE SUPOLF(KM,KNSMAX,DDMU,DDPOL,KCHEAP)
 
     DO JN=KM,KNSMAX,2
       DO JCORR=1,ICORR(JN)
-        DDPOL(JN)=DDPOL(JN)/ZSCALE
+        DDPOL(JN)=MAX(DDPOL(JN)/ZSCALE, ZEPS)
```
wdeconinck: I think this modification is equivalent to what is in the following lines.
Perhaps you can revert this modification.

samhatfield: Wooooops I think this was an error!

samhatfield:
Getting there. I've added some comments to clarify the logic which handles very small polynomial values. There seems to be some clever logic for extending the range of double-precision reals, in which potential underflows are preemptively detected and a rescaling is applied. This is undone at the end. This will help in cases where the first polynomials in a series are too small to be stored but later ones (computed from the earlier ones) can be stored. I think in practice this will rarely be the case but I haven't thought it through properly yet.

wdeconinck (approved):
For me this is good to go as soon as you checked bit-identicality. So I'll approve already

samhatfield:
Bit identicality confirmed with this program:

<details>

```fortran
PROGRAM CHECK_POLY_BIT_ID

USE PARKIND1, ONLY: JPIM, JPRB, JPIB
USE MPL_MODULE, ONLY: MPL_INIT, MPL_NPROC, MPL_MYRANK, MPL_END

IMPLICIT NONE

#include "setup_trans0.h"
#include "setup_trans.h"
#include "trans_inq.h"
#include "trans_end.h"

INTEGER(KIND=JPIM), PARAMETER :: TRUNCATION = 7999
INTEGER(KIND=JPIM), PARAMETER :: NLAT = 2 * (TRUNCATION + 1)

INTEGER(KIND=JPIM) :: LOEN(NLAT)
REAL(KIND=JPRB), ALLOCATABLE :: POLYS(:,:)

INTEGER(KIND=JPIM) :: IL, IM, POLYS_SIZE, NPROC, MYPROC
INTEGER(KIND=JPIB) :: CHECKSUM

CALL MPL_INIT
NPROC  = MPL_NPROC()
MYPROC = MPL_MYRANK()

IF (MYPROC == 1) WRITE(*,*) "NPROC = ", NPROC

DO IL = 1, NLAT / 2
  LOEN(IL) = 20 + 4 * (IL - 1)
  LOEN(NLAT - IL + 1) = LOEN(IL)
ENDDO

CALL SETUP_TRANS0(KPRINTLEV=0, KPRGPNS=NPROC, KPRTRW=NPROC)
CALL SETUP_TRANS(KSMAX=TRUNCATION, KDGL=NLAT, KLOEN=LOEN)

CALL TRANS_INQ(KSPOLEGL=POLYS_SIZE)

ALLOCATE(POLYS(NLAT / 2, POLYS_SIZE))

CALL TRANS_INQ(PRPNM=POLYS)

CALL CRC64(POLYS, INT(SIZE(POLYS) * KIND(POLYS), 8), CHECKSUM)
WRITE(1000+MYPROC,'(Z16.16)') CHECKSUM

CALL TRANS_END
CALL MPL_END

END PROGRAM
```

</details>

This has each MPI task calculate a CRC64 checksum of its polynomial array (this is of course decomposed among W-set members) and write this to a separate file. I ran this program for develop and tidy_supolf at TCO1279, 3999, and 7999, and verified that all checksums are identical. Good to merge.

PR #324, opened 2025-10-25
Allocate variable before subroutine calls

DJDavies2:
While testing ectrans I ran in to runtime failures in some ctests relating to allocatable arrays being passed to dump_checksums when they weren't allocated.

The solution I have implemented here is to allocate them to a zero-sized array, then deallocate them when after the subroutine returns.  This isn't very clean I know.  Is there a better way of fixing this?


samhatfield:
Well spotted @DJDavies2. I am not sure why the 6 or so compilers we test through CI don't have a problem with this. The proper solution is to declare `ivsetsc` as an optional argument of `dump_checksums` and simply not pass it for call mode 2. I'll prepare a fix later.

samhatfield:
Alternative PR opened [#328](https://github.com/ecmwf-ifs/ectrans/pull/328).

PR #325, opened 2025-10-25
Initialise pointer in ectrans_memory

DJDavies2:
While testing I ran into failures related to use of an uninitialized pointer c_label.  Actually it was being passed to the associated intrinsic; this is invalid Fortran if the pointer is in an undefined state, as it is at the start.  I have fixed this by initializing it in its declaration.

### DIFF
--- a/src/programs/util/ectrans_memory.F90
+++ b/src/programs/util/ectrans_memory.F90
@@ -81,7 +81,7 @@ end type
 
 type(allocator_t) :: allocator
 
-character(kind=c_char), pointer, private :: c_label(:)
+character(kind=c_char), pointer, private :: c_label(:) => null()
 
 interface
     function c_allocate_var(bytes) result(ptr) bind(c, name="ectrans_memory_allocate_var")


wdeconinck:
Hi @DJDavies2 thanks again (and sorry) for finding these issues, I suppose with NAG compiler? We really ought to have a CI with this... Do you know the best approach for this? I suppose it is licensed. Really appreciate it!
I have amended the PR with one more commit. This is good to merge.

DJDavies2:
Sorry, I don't know anything about CIs, I'm just running on our own machines.  Yes using NAG would require purchasing licenses, but I don't know how much it costs.

One option might be to build with the compilers you do have but with more aggressive runtime checking options.  It might catch some of these stuff.  It won't catch everything though, it will be slow and may have bugs of its own.

I am happy to submit fixes as and when I come across them.

src/programs/util/ectrans_memory.F90
```diff
@@ -364,6 +364,7 @@ subroutine set_label(label)
     subroutine unset_label()
         if (associated(c_label)) then
             deallocate(c_label)
+            c_label => null()
```
DJDavies2: I don't think this is necessary.  Pointers that are deallocated are automatically set to null.  There are cases where pointers can become undefined in the middle of a program, but this situation is fine.

wdeconinck: Okay I will revert this. Thanks for explaining.

wdeconinck (approved):

PR #326, opened 2025-10-25
Fix some IO problems in ectrans-benchmark

DJDavies2:
While testing I found some IO issues in ectrans-benchmark:

  * Use of Z formating for non-integer variables
  * Missing commas in write statements

For the Z formatting I have used transfer to convert the values to integers.  I have also inserted the missing commas.


samhatfield:
Seems reasonable to me, let's test it out.

wdeconinck (approved):
Looks good to me, Thanks!

PR #328, opened 2025-10-31
Fix dump checksums unallocated array issue

samhatfield:
We currently pass `ivsetsc` and `ivsetsc2` always to `dump_checksums`, even when they're not needed. This is a problem especially for `ivsetsc` because this is only allocated for call mode 1, not 2. Passing an unallocated array when calling a subroutine is a bit dodgy, so this PR fixes the situation by making them optional arguments.

@DJDavies2 could you give this a spin with NAG?

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -644,7 +644,7 @@ do jstep = 1, iters+iters_warmup
       write (checksums_filename,'(A)') trim(cchecksums_path)//'_inv_trans.checksums'
       call dump_checksums(filename=checksums_filename, noutdump=noutdump, jstep=jstep, &
         &                 myproc=myproc, nproma=nproma, ngptotg=ngptotg, ivset=ivset, &
-        &                 ivsetsc=ivsetsc, ivsetsc2=ivsetsc2, nspec2g=nspec2g, zgp=zgp)
+        &                 ivsetsc=ivsetsc, nspec2g=nspec2g, zgp=zgp)
     endif
   else
     call inv_trans(pspvor=zspvor, pspdiv=zspdiv, pspsc3a=zspsc3a, pspsc2=zspsc2, pgpuv=zgpuv, &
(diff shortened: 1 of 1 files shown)


src/programs/ectrans-benchmark.F90
```diff
@@ -1473,8 +1469,8 @@ subroutine dump_checksums(filename, noutdump,                      &
   integer(kind=jpim), intent(in) :: ngptotg
   integer(kind=jpim), intent(in) :: nspec2g
   integer(kind=jpim), intent(in) :: ivset(:)
-  integer(kind=jpim), intent(in) :: ivsetsc(:)
-  integer(kind=jpim), intent(in) :: ivsetsc2(:)
+  integer(kind=jpim), intent(in), optional :: ivsetsc(:)
+  integer(kind=jpim), intent(in), optional :: ivsetsc2(:)
```
wdeconinck, line 1470: Okay to be optional, but the code further does not assert on the presence (`if (present(ivsetsc)`) if needed.
And I don't see changes at the call site, not passing these arrays.
-- edit -- scratch last sentence. it was obfuscated in the full diff with the style changes; I should have looked at the commit individually.

DJDavies2:
I think this works for me although it isn't up to date with other NAG related changes (e.g. the format stuff) so I have had to patch my working copy a little bit.  The runtime errors I was encountering seem to have gone away.  Eyeballing the code indicates that ivsetc is only passed when it is allocated.  I think they will only be accessed in dump_checksums when they are present as although their presence is not tested directly, the accesses are hidden behind tests for the presence of other optional arguments that in practice prevent invalid access.

In fact  I wonder whether the arguments ivetsc on line 635 and ivsetsc2 on line 662 are actually needed, as the relevant other arguments (zspscalar/zspsc2) are not present, so ivetsc/ivetsc2 will not be accessed in these cases.  However I have not tested this.

From my point of view I think this change is fine.

samhatfield:
Thanks for the feedback both. I have refactored `dump_checksums` so that all passed arguments are checked for consistency. The subroutine has less flexibility now, but it's only called in one of 4 ways anyway, so I think this makes sense.

wdeconinck (approved):
Looks good!

PR #330, opened 2025-11-06
Tweak CMake features

samhatfield:
I started documenting all the [features](https://sites.ecmwf.int/docs/ectrans/page/cmake_features.html) of ecTrans and realised there were some inconsistencies. Trying to address them all here, but if there's anything else you've been meaning to fix or change, feel free to add to this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,7 @@ endif()
 ecbuild_add_option( FEATURE GPU
                     DEFAULT OFF
                     DESCRIPTION "Compile GPU version of ectrans (Requires OpenACC or sufficient OpenMP offloading support)"
-                    CONDITION (HAVE_HIP OR HAVE_CUDA) AND (HAVE_ACC OR HAVE_OMP) )
+                    CONDITION (HAVE_HIP OR HAVE_CUDA) AND (HAVE_ACC OR HAVE_OMP) AND HAVE_MPI )
 
 # Check CPU or GPU is enabled, and if not, abort
 if( (NOT HAVE_CPU) AND (NOT HAVE_GPU) )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ ecbuild_enable_fortran( REQUIRED NO_MODULE_DIRECTORY )
 ecbuild_find_package( NAME fiat REQUIRED )
 
 ecbuild_add_option( FEATURE MPI
+                    DEFAULT ON
                     DESCRIPTION "Support for MPI distributed memory parallelism"
                     REQUIRED_PACKAGES "MPI COMPONENTS Fortran CXX"
                     CONDITION fiat_HAVE_MPI )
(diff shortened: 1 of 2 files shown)


wdeconinck:
Regarding the MPI feature I think things have gotten out of control over time.

Normally, the CPU version at least, should strictly be using MPL so it is in fact not a controllable feature.
Instead we should just define these variables
```cmake
set( HAVE_MPI ${fiat_HAVE_MPI} )
set( ectrans_HAVE_MPI ${fiat_HAVE_MPI} ) # we need to define both
```
We define these variables to hide the FIAT implementation detail, and ectrans_HAVE_MPI gets exported for introspection to downstream packages.

The GPU code path does link directly with `MPI::MPI_Fortran` [here](https://github.com/ecmwf-ifs/ectrans/blob/da8220a54f8a506b11a9e20aab44f687c33c017f/src/trans/gpu/CMakeLists.txt#L151), I think to allow GPU-aware MPI.
We should therefore only search for MPI when we want to enable the GPU_AWARE_MPI feature, and only if we HAVE_GPU and HAVE_MPI.
```cmake
ecbuild_add_option( FEATURE GPU_AWARE_MPI
                    DEFAULT ON
                    CONDITION HAVE_GPU AND HAVE_MPI            # This means MPI is enabled via FIAT
                    REQUIRED_PACKAGES "MPI COMPONENTS Fortran" # Removed CXX... why was this there?
                    DESCRIPTION "Enable CUDA-aware )
```
I removed the CXX component. Not sure why it is needed. Perhaps a leftover no longer needed?
And in the above mentioned place  [here](https://github.com/ecmwf-ifs/ectrans/blob/da8220a54f8a506b11a9e20aab44f687c33c017f/src/trans/gpu/CMakeLists.txt#L151) we need to only link with `MPI::MPI_Fortran` in case GPU_AWARE_MPI is enabled.
```cmake
                               # ...
                               $<${HAVE_GPU_AWARE_MPI}:MPI::MPI_Fortran>
```
Perhaps (@marsdeno needs to confirm) we can forget linking directly with MPI::MPI_Fortran when FIAT is compiled with MPI_F08, so that this can be deprecated over time?

marsdeno:
Can confirm that with MPI_F08-enabled FIAT, we no longer need a direct link against MPI::MPI_Fortran

samhatfield:
Good points @wdeconinck. We don't need a separate MPI feature for ecTrans do we? I've deleted that and put in your other suggestions.

wdeconinck (approved):
Nice work.
When fiat is released with MPI_F08 we should be able to introspect the fiat version (or MPI_F08 feature) and avoid linking with MPI::MPI_Fortran directly.

samhatfield:
I realised that MPI is not actually required for the GPU version if you just want to run with one task / GPU. One now needs to build FIAT without MPI (rather than disabling ecTrans' MPI feature) and run with `ECTRANS_USE_MPI=0` exported. I removed this condition so we still have the freedom to debug on a single task / GPU without requiring an MPI library (admittedly a bit of an academic concern nowadays).

PR #332, opened 2025-11-11
Correct test properties for test with FLT option

samhatfield:
We weren't setting the properties for the FLT test properly.


wdeconinck (approved):
copy paste error 😱

PR #331, opened 2025-11-07
Remove references to unused FSPGL_PROC in inverse

samhatfield:
The use of `FSPGL_PROC` (user-passed Fourier space calculation callback function) is inconsistent in the adjoint code path. `FSPGL_PROC` is received by `INV_TRANSAD` but then not used for anything. `INV_TRANS_CTLAD` has no `FSPGL_PROC` argument. However `LTINV_CTLAD` _does_, and passes it to `LTINVAD`. There it isn't used because the code calling it is commented out, and has been commented out for 20 years. So let's just delete this. That should resolve #329.

We could also delete the argument from `INV_TRANSAD` but this would have to be done carefully because we do pass this argument in calls to `INV_TRANSAD` in the IFS.

Thanks @DJDavies2 for raising [#329](https://github.com/ecmwf-ifs/ectrans/issues/329).

Issue #329, opened 2025-11-01
Failure to compile with NAG


DJDavies2:
### What happened?

Trying to compile ectrans with NAG fails with this failure:

Error: /home/users/david.davies/cylc-run/mo-bundle-415/share/make_mo_base__spice_nag/ectrans/src/trans/cpu/generated/ectrans_dp/internal/ltinv_ctlad_mod.F90: FSPGL_PROC is not a dummy argument - cannot be OPTIONAL

### What are the steps to reproduce the bug?

Compile with NAG

### Version

develop

### Platform (OS and architecture)

Linux

### Relevant log output

```shell

```

PR #333, opened 2025-11-12
Patches for nag

DJDavies2:
For #329 and #331.

I'm not sure exaclty what NAG has problems with here, but the way I did it was to create dual ltinv/ltinv_ad subroutines, one of which has the optional procedure argument and one of which does not.  Then in inv_trans_ctl use an if statement to select one instead of relying on the optionality being automatically passed down.

It is very ugly and I would understand if it wasn't wanted as a solution.


samhatfield:
Hmm, yes, not ideal. There's this funny hint in the code too (I didn't write that):
```
  !!!WARNING!!! Duplication of code besides the FSPGL_PROC argument.
              ! It seems that gfortran 10 does not retain the value
              ! of FSPGL_PROC within the OMP region.
```
Do you know why this wasn't an issue with NAG before? I don't think we've changed any of this code recently.

DJDavies2:
I have only fairly recently been trying to compile ectrans with NAG, I haven't tried with older versions.

samhatfield:
Fair enough. I'm going to go out on a limb and trust NAG here, I think we must be doing something non-standards compliant with `FSPGL_PROC`. However it's difficult for me to experiment with a solution, not having access to the NAG compiler.

I've just had a go at implementing an explicit interface for `FSPGL_PROC`, branch here:

https://github.com/ecmwf-ifs/ectrans/tree/FSPGL_PROC_explicit_interface

Could you try this to see if it resolves the compiler errors with NAG?

DJDavies2:
> Fair enough. I'm going to go out on a limb and trust NAG here, I think we must be doing something non-standards compliant with `FSPGL_PROC`. However it's difficult for me to experiment with a solution, not having access to the NAG compiler.
> 
> I've just had a go at implementing an explicit interface for `FSPGL_PROC`, branch here:
> 
> https://github.com/ecmwf-ifs/ectrans/tree/FSPGL_PROC_explicit_interface
> 
> Could you try this to see if it resolves the compiler errors with NAG?

I tried this branch, yes that seems to work.  Thanks for that.

It is better to have explicit interfances anyway I think.

samhatfield:
> I tried this branch, yes that seems to work. Thanks for that.
> 
> It is better to have explicit interfances anyway I think.

I wholeheartedly agree. I do need to check this is compatible with the IFS before merging though.

I'm going to close this PR since we've found our solution for NAG.

PR #334, opened 2025-11-14
Make sure MPI is found when GPU and MPI enabled

samhatfield:
The GPU version still has raw MPI calls so we must find MPI in the case when GPU and MPI features are enabled.

Note that disabling MPI is done through FIAT's MPI feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,10 @@ if( (NOT HAVE_CPU) AND (NOT HAVE_GPU) )
   ecbuild_critical("Please enable one or both of the CPU and GPU features")
 endif()
 
+if( HAVE_GPU AND HAVE_MPI )
+  ecbuild_find_package( NAME MPI REQUIRED COMPONENTS Fortran )
+endif()
+
 if( HAVE_GPU )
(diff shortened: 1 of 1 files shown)


samhatfield:
@dhaumont

dhaumont (commented):
LGTM!

dhaumont (approved):

samhatfield:
Actually @dhaumont, @wdeconinck and I think we have a better solution which will pave the way towards eventually switching to pure MPL calls. This will bring the GPU version in line with the CPU version. Will try to draft another PR today.

dhaumont:
I understand. But as the problem has been identified and we have a workaround,  I would push this PR to fix the broken build, and prevent other people from wasting as much time as I did.

samhatfield:
> I understand. But as the problem has been identified and we have a workaround, I would push this PR to fix the broken build, and prevent other people from wasting as much time as I did.

It's not a huge amount of coding required - I'm already testing the new branch now.

samhatfield:
PR https://github.com/ecmwf-ifs/ectrans/pull/335 is the preferred solution.

PR #339, opened 2025-11-20
Split dump_checksums_pgp into 4 methods

dhaumont:
This PR proposes splitting `dump_checksums_pgp` into 4 different methods.
This simplifies the code of `dump_checksums_pgp`, and reduces the risk of errors when using it:
- removal of the optional arguments
- removal of `iconfig` from dump_checksums
- simplification of the allocation of gspfld and gpfld
- drastic reduction in the number of if statements in dump_checksums*

Unfortunately, this slightly lenghtens the cause due to Fortran verbosity, but I believe this is an acceptable price to pay for reducing code complexity

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -42,6 +42,7 @@ real(kind=jprd) :: zmaxerrg
 integer(kind=jpim), parameter :: nerr     = 0 ! Unit number for STDERR
 integer(kind=jpim), parameter :: nout     = 6 ! Unit number for STDOUT
 integer(kind=jpim), parameter :: noutdump = 7 ! Unit number for field output
+integer(kind=jpim), parameter :: noutdump_checksum = 8 ! Unit number for dump_checksun
 
 ! Default parameters
 integer(kind=jpim) :: iters   = 10  ! Number of iterations for transform test
@@ -642,8 +643,8 @@ do jstep = 1, iters+iters_warmup
       iend = ngptot - nproma * (ngpblks - 1)
       zgp (iend+1:, :, ngpblks) = 0
       write (checksums_filename,'(A)') trim(cchecksums_path)//'_inv_trans.checksums'
-      call dump_checksums(filename=checksums_filename, noutdump=noutdump, jstep=jstep, &
-        &                 myproc=myproc, nproma=nproma, ngptotg=ngptotg, zgp=zgp)
+      call dump_checksums_pgp(filename=checksums_filename, noutdump=noutdump_checksum, jstep=jstep, &
+        &                     myproc=myproc, nproma=nproma, ngptotg=ngptotg, zgp=zgp)
     endif
   else
     call inv_trans(pspvor=zspvor, pspdiv=zspdiv, pspsc3a=zspsc3a, pspsc2=zspsc2, pgpuv=zgpuv, &
@@ -658,9 +659,9 @@ do jstep = 1, iters+iters_warmup
       zgp3a (iend+1:, :, :, ngpblks) = 0
       zgp2 (iend+1:, :, ngpblks) = 0
       write (checksums_filename,'(A)') trim(cchecksums_path)//'_inv_trans.checksums'
-      call dump_checksums(filename=checksums_filename, noutdump=noutdump, jstep=jstep, &
-        &                 myproc=myproc, nproma=nproma, ngptotg=ngptotg, zgpuv=zgpuv, zgp3a=zgp3a, &
-        &                 zgp2=zgp2)
+      call dump_checksums_pgp_uv_3a_2(filename=checksums_filename, noutdump=noutdump_checksum, jstep=jstep, &
+        &                             myproc=myproc, nproma=nproma, ngptotg=ngptotg, zgpuv=zgpuv, zgp3a=zgp3a, &
+        &                             zgp2=zgp2)
     endif
   endif
   call gstats(4,1)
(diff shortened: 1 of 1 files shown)


samhatfield:
Hi @dhaumont, how does this PR interact with https://github.com/ecmwf-ifs/ectrans/pull/314? There too you have "split dump_checksuns into 4 functions" but it has a different commit hash to the one in this PR. Once this PR is merged, will you delete the corresponding commit in https://github.com/ecmwf-ifs/ectrans/pull/314 and merge develop to it again?

dhaumont:
Hi @samhatfield. Yes, it is exactly the same code, but I just reimplemented it here such that it is easier for you to review.  I will just merge this into my branch later on.
This change is completely optional, we can keep it or not, it is up to you.

wdeconinck:
I'm happy it is split out from the other PR, where it can be removed then to really make targeted contributions tested in isolation.
I have no strong opinion on this feature, but I think it's better to not have to rely on too many "if(present)"

samhatfield:
I definitely agree that the current `dump_checksums` looks very clumsy and could be improved. Still, I wonder if it's possible  to go even further than what's in this PR. There's still a lot of repetition and I wonder if we can reduce it by creating a generic function for dumping checksums of a single field, a bit like `dump_gridpoint_field`.

Also I think it's best to have just a single `gath_grid` rather than one for each level. That subroutine is designed to work well with a batch of fields rather than just one at a time. The same applies to `gath_spec` which in fact doesn't really make sense when applied to a single level (when `NPRTRV>=2`). In fact as I mentioned elsewhere, I think that's a bug.

EDIT: unfortunately I think doing all `gath_grid`s at the same time before the loop over levels would result in passing a slice of `gfld(jfld,:)` which is a horrible strided access.

dhaumont:
Yes, I also agree that we can even go further. Unfortunately, I'm afraid I will not have the time myself to implement the improvements suggested  in the coming weeks. If you think this PR is already a step into the right direction, can't we keep the improvements for later, and already merge the PR as it is ? It is quite safe, in my opinion

samhatfield:
It's definitely a step in the right direction. I've got some minor edits to make to e.g. whitespace -> do you mind if I make changes and force push them to your branch? I have the privilege as a maintainer of the repository.

After this PR there will still be a bug in the calls to `gath_spec`. I'll have to open another PR afterwards to fix them.

samhatfield:
One other thing @dhaumont - in order to fix the bug I mentioned in how `gath_spec` is called, I don't think it will be possible to print the checksums of each individual level in spectral space anymore.

The reason is because we need to call `gath_spec` once for all levels and collect into a global array with the levels on the first dimension. Then, if we wanted to compute a checksum for one level at a time, we would have to slice the array like `(jlev,:)` which is a strided access and is incredibly slow.

Instead we would have to compute and print the checksum for the whole array. Would that be okay?

dhaumont:
> It's definitely a step in the right direction. I've got some minor edits to make to e.g. whitespace -> do you mind if I make changes and force push them to your branch? I have the privilege as a maintainer of the repository.

Yes, of course I don't mind if you take over and make the necessary changes. You can even rename the functions if you want. Thanks for your help to finalize this

dhaumont:
> One other thing @dhaumont - in order to fix the bug I mentioned in how `gath_spec` is called, I don't think it will be possible to print the checksums of each individual level in spectral space anymore.
Instead we would have to compute and print the checksum for the whole array. Would that be okay?

@samhatfield I understand, and we need to prioritize the fastest solution for production. A single checksum for the entire array is therefore acceptable to check for regressions. However, more detailed level-by-level debugging is also very valuable. We could enable its on demand during debuging. Given that we already have the output_grid_point capability,  what additional debugging information could be useful to output for debugging ?

src/programs/ectrans-benchmark.F90
```diff
@@ -1452,190 +1453,277 @@ end subroutine dump_gridpoint_field
...
-                        & zspsc3a, zspsc2)
+subroutine open_dump_checksums_file(filename, noutdump, jstep)
+
+  character(len=*),   intent(in) :: filename
+  integer(kind=jpim), intent(in) :: noutdump ! unit number for output file
+  integer(kind=jpim), intent(in) :: jstep
+  logical :: exist
+
+  exist = .false.
+  if (jstep > 1) inquire(file=trim(filename), exist=exist)
```
samhatfield, line 1464: Any reason we still need this if statement?

samhatfield: Oh I understand. Nevermind.

samhatfield:
> @samhatfield I understand, and we need to prioritize the fastest solution for production. A single checksum for the entire array is therefore acceptable to check for regressions. However, more detailed level-by-level debugging is also very valuable. We could enable its on demand during debuging. Given that we already have the output_grid_point capability, what additional debugging information could be useful to output for debugging ?

How long is a piece of string? :D My philosophy is that the code should only have in it very common debugging features and anything that would be necessary for automation.

dhaumont:
> How long is a piece of string? :D My philosophy is that the code should only have in it very common debugging features and anything that would be necessary for automation.

So I learned a new expression, thanks ! :-D
As we've already discussed, checksums are useful for verifying: 1. the absence of non regression  and 2.  bit reproducibility between different MPI runs. 
Regarding point 2., sufficient granularity is necessary to easily identify and correct bugs. Since the main difference between MPI runs lies in the distribution of levels between the different procs, it is essential to have a checksums for each level. However, given the slowness you mentionned, a good compromise would be to calculate only a global checksum by default, and allow detailed checksums to be activated on demand, similar to the "verbose mode" used during debugging.

wdeconinck:
Some kind `dump_checksums(LCOMBINED=T/F)` argument to the routine could be a compromised solution? controllable with extra command line argument, which by default is such that we combine the checksums for max performance.

samhatfield:
Let me have a think about possible solutions. Certainly we need to fix the bug in `gath_spec` ASAP so I will focus on that. In the mean time, this PR is ready for merging IMO.

dhaumont:
> Let me have a think about possible solutions. Certainly we need to fix the bug in `gath_spec` ASAP so I will focus on that. In the mean time, this PR is ready for merging IMO.

@samhatfield Thanks Sam. Can you validate your review in Github, to reflect this statement ?
@wdeconinck @marsdeno Is it OK also for you to merge ?

samhatfield (approved):

wdeconinck (approved):
Thanks @dhaumont !!!

PR #340, opened 2025-11-26
Fix adjoint TRMTOL and TRLTOM

samhatfield:
I noticed some issues with these two subroutines, so here are a few simple commits to fix some potential bugs and make the two routines more symmetric.


samhatfield:
PR #342 makes this one redundant.

PR #341, opened 2025-12-01
Replace MPI tests 2 task case with 4 task case

samhatfield:
The 4 task case is a much better test as it actually activates the NPRTRV > 1 code paths.


samhatfield:
Lots of failures as I thought. I think these are due to the `gath_spec` bug I mentioned in other issues.

samhatfield:
Rather than merge this PR, I propose to open another PR where various issues related to bit-reproducibility for the 4 task MPI case are resolved. In fact [`fix_gath_spec_bug`](https://github.com/ecmwf-ifs/ectrans/tree/fix_gath_spec_bug) does exactly this but this branch builds on https://github.com/ecmwf-ifs/ectrans/pull/339 so that must be merged first.

PR #342, opened 2025-12-04
Remove redundant adjoint GPU communication routines

samhatfield:
In the CPU code path, we don't have "AD" versions for `TRLTOG`, `TRLTOM` etc. Instead we simply use the inverse routine (e.g. G->L instead of L->G) with the input / output arguments swapped. In the GPU code path though we have additional subroutines `TRLTOMAD` and `TRMTOLAD`. This PR is for investigating whether these routines can be eliminated. If this succeeds, then we can close PR https://github.com/ecmwf-ifs/ectrans/pull/340 unmerged.


samhatfield:
Any thoughts on this @l90lpa? I was hoping I could also delete `TRLTOMAD_PACK_UNPACK` and `TRMTOLAD_PACK_UNPACK` and somehow reuse the non-adjoint versions, saving another ~700 lines of code, but the adjoint packing/unpacking routines do seem to be genuinely different. I might therefore undo commit [f995241](https://github.com/ecmwf-ifs/ectrans/pull/342/commits/f995241500def352c1c1dcd11f1cdb16254a5106).

l90lpa:
Hi @samhatfield, sorry for the slow reply. Yes, I had originally wanted to take the approach you're discussing, however, the issue I ran into was that `trltom_pack/trltom/trltom_unpack` and `trmtol_pack/trmtol/trmtol_unpack` don't just do transposition but also have some over responsibilities built in that only apply to one transform direction. If I recall correct, `trltom_pack` isn't just packing a buffer, it is also applying quadrature weights. And again, if I recall correctly somewhere in `trmtol_pack/unpack` it handles frequency padding. There might be some other bits I forgot. At the time, I did look and is seemed possible to break up these routines to avoid needing additional adjoint code, but this would have required more work outside of the scope of the PRs than I had time for (and I wasn't sure if the GPU code had been written this way intentionally).

samhatfield:
Thanks for the reply @l90lpa. I am cleaning up some old PRs. At the time I thought this wasn't important and put it on the backburner, but now that I've looked at it again, I can't see a reason to remove the two adjoint L<->M transposition routines. Each appears to be identical to the corresponding non-adjoint routine in the opposite transform direction so it seems like a no brainer.

As you say, the packing routines are more complicated, so for now I won't touch those.

This is ready for review.

samhatfield:
Wow, zero comments from Copilot...

samhatfield:
And this @wdeconinck 🙏

src/trans/gpu/internal/dir_trans_ctlad_mod.F90
```diff
@@ -153,7 +153,7 @@ SUBROUTINE DIR_TRANS_CTLAD(KF_UV_G,KF_SCALARS_G,KF_GP,KF_FS,KF_UV,KF_SCALARS,&
 
       CALL GSTATS(153,0)
       CALL TRLTOMAD_UNPACK(ALLOCATOR,HTRLTOM_UNPACK,FOUBUF,ZINPS,ZINPA,ZINPS0,ZINPA0,KF_FS,KF_UV)
-      CALL TRLTOMAD(ALLOCATOR,HTRLTOM,FOUBUF_IN,FOUBUF,KF_FS)
+      CALL TRMTOL(ALLOCATOR,HTRMTOL,FOUBUF,FOUBUF_IN,KF_FS)
```
wdeconinck, line 157: Can we add a comment here (and also for inv_trans) so we don't think this is a typo, explaining that the inverse can be used here for the adjoint.
Perhaps also in the CPU code path if it is not there.

wdeconinck (approved):
Approved, pending resolution of the comments.
Just go ahead and merge when ready.

PR #343, opened 2025-12-05
Update CUFFT error codes

samhatfield:
Quoting Nvidia docs (https://docs.nvidia.com/cuda/cufft/):

Starting from CUDA 13.0:

The following error codes have been removed:
CUFFT_INCOMPLETE_PARAMETER_LIST, CUFFT_PARSE_ERROR, CUFFT_LICENSE_ERROR.

Starting from CUDA 12.9:

The following error codes are deprecated and will be removed in a future release: CUFFT_INCOMPLETE_PARAMETER_LIST, CUFFT_PARSE_ERROR, CUFFT_LICENSE_ERROR.

### DIFF
--- a/src/trans/gpu/algor/hicfft_cuda.h
+++ b/src/trans/gpu/algor/hicfft_cuda.h
@@ -118,6 +118,9 @@ inline static const char * _fftGetErrorEnum(cufftResult error)
         case CUFFT_NOT_SUPPORTED:
         return "CUFFT_NOT_SUPPORTED";
 
+// These are only available from 13.0 onwards
+// https://docs.nvidia.com/cuda/archive/12.9.1/cufft/index.html#return-value-cufftresult
+#if defined(CUDART_VERSION) && CUDART_VERSION >= 13000
         case CUFFT_MISSING_DEPENDENCY:
         return "CUFFT_MISSING_DEPENDENCY";
 
@@ -129,6 +132,7 @@ inline static const char * _fftGetErrorEnum(cufftResult error)
 
         case CUFFT_NVSHMEM_FAILURE:
         return "CUFFT_NVSHMEM_FAILURE";
+#endif
 
 // These are deprecated from CUDA 13.0 onwards
 // https://docs.nvidia.com/cuda/cufft/#deprecated-functionality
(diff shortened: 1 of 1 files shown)


src/trans/gpu/algor/hicfft_cuda.h
```diff
@@ -106,15 +106,9 @@ inline static const char * _fftGetErrorEnum(cufftResult error)
         case CUFFT_UNALIGNED_DATA:
         return "CUFFT_UNALIGNED_DATA";
 
-        case CUFFT_INCOMPLETE_PARAMETER_LIST:
-        return "CUFFT_INCOMPLETE_PARAMETER_LIST";
-
```
wdeconinck, line 111: We could wrap this:
```c++
#if defined(CUDART_VERSION) && CUDART_VERSION < 13000
        case CUFFT_INCOMPLETE_PARAMETER_LIST:
        return "CUFFT_INCOMPLETE_PARAMETER_LIST";
#endif
```

src/trans/gpu/algor/hicfft_cuda.h
```diff
@@ -106,15 +106,9 @@ inline static const char * _fftGetErrorEnum(cufftResult error)
...
         case CUFFT_INVALID_DEVICE:
         return "CUFFT_INVALID_DEVICE";
 
-        case CUFFT_PARSE_ERROR:
-        return "CUFFT_PARSE_ERROR";
-
```
wdeconinck, line 117: Same here. Also put link to https://docs.nvidia.com/cuda/cufft/#deprecated-functionality

wdeconinck (changes requested):

src/trans/gpu/algor/hicfft_cuda.h
lukasm91: there are also a few cases missing, some of them obviously not important, but maybe cross check with https://docs.nvidia.com/cuda/cufft/index.html#return-value-cufftresult.  (e.g. CUFFT_MISSING_DEPENDENCY and later are missing).

samhatfield:
Missing error codes added, and deprecated ones added back with CUDART version check.

I assume `CUDART_VERSION` is automatically in scope when compiling with `nvcc`?

lukasm91:
nvcc includes `cuda_runtime.h` by default, yes. That indirectly includes `cuda_runtime_api.h` which defines that macro. If you were not using a cu but a cpp file and only use CUDA Runtime API, you would need to make sure that this header is included directly or indirectly.

wdeconinck (approved):
Thanks Sam!

PR #344, opened 2025-12-05
Delete unused gpu/INIGPTR subroutine

samhatfield:
This is precision agnostic so pretty straightforward to move to common. I just had to modify the signature so that NPROMA and NGPBLKS are passed as arguments. This is necessary because TPM_TRANS is not in common, so we cannot simply take these variables from that module.

Note that the GPU version could have been deleted in any case as it isn't used at all.


wdeconinck:
I see the reason for doing this, but on the other hand, "common" is also intended to be common between cpu and gpu code paths.
As you found out, it is not shared with the gpu code path, so I am not entirely happy with the move to common as a code cleanup.
For the gpu code path, we also have a [ectrans_gpu_common](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/CMakeLists.txt#L11) library for precision independent routines. For the cpu code path we did not do this. This would be the right approach here as well but I am not sure it is worth the extra complexity for one file currently?

samhatfield:
That's fair enough. So would you say the common library is for stuff *actually* used by both CPU and GPU backends, not just stuff that can work regardless of the precision and the backend? In that case `INIGPTR` should not go in common.

We can still delete `INIGPTR` from the GPU backend.

samhatfield:
PR updated so it just deletes the unused gpu/INIGPTR subroutine.

wdeconinck (approved):

PR #346, opened 2025-12-05
Remove redundant LDFOU2AD

samhatfield:
It's literally exactly the same as LDFOU2.


wdeconinck (approved):
I suppose we have an adjoint test that would have failed?

samhatfield:
> I suppose we have an adjoint test that would have failed?

In this case `LDFOU2` is _supposed_ to be the same as the adjoint version, so it wouldn't have failed. We would just have two identical subroutines, so this PR is just for removing some duplicated code.

wdeconinck:
What I meant is that if this were somehow not identical, then a CI test should have failed. All good.

PR #347, opened 2025-12-22
Update macOS and clang versions in workflows

samhatfield:
[macos 13 is deprecated](https://github.com/actions/runner-images/issues/13046).

### DIFF
--- a/.github/workflows/build-ectrans4py.yml
+++ b/.github/workflows/build-ectrans4py.yml
@@ -47,8 +47,8 @@ jobs:
 
           - name: macos
             # Xcode compiler requires empty environment variables, so we pass null (~) here
-            os: macos-latest
-            compiler: clang-14
+            os: macos-15
+            compiler: clang-18
             compiler_cc: ~
             compiler_cxx: ~
             compiler_fc: gfortran-13
--- a/.github/workflows/build-ectrans4py.yml
+++ b/.github/workflows/build-ectrans4py.yml
@@ -47,7 +47,7 @@ jobs:
 
           - name: macos
             # Xcode compiler requires empty environment variables, so we pass null (~) here
-            os: macos-13
+            os: macos-latest
             compiler: clang-14
             compiler_cc: ~
             compiler_cxx: ~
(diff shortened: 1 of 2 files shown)


.github/workflows/build-ectrans4py.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
           - name: macos
             # Xcode compiler requires empty environment variables, so we pass null (~) here
-            os: macos-13
+            os: macos-latest
             compiler: clang-14
```
wdeconinck: `clang-14` should be replaced with just `clang` as it is also not up to date. With `latest` we don't really know what it is.

.github/workflows/build.yml
```diff
@@ -77,7 +77,7 @@ jobs:
 
           - name: macos
             # Xcode compiler requires empty environment variables, so we pass null (~) here
-            os: macos-13
+            os: macos-latest
             compiler: clang-14
```
wdeconinck: Same here.

wdeconinck (changes requested):

.github/workflows/build-ectrans4py.yml
samhatfield: Is `compiler` only used for labelling the cache for `Cache Dependencies` or is it used for something important?

Maybe we should actually pin the macos version to `macos-15`, until that's deprecated, so we know exactly which OS we're targeting.

wdeconinck: Indeed, it is better to be explicit.

samhatfield: Okay, done.

samhatfield:
I don't quite understand the compiler suite on macOS. We are asking for gfortran-13, and according to the CMake output, that's what we're getting for the Fortran compiler. But the C and C++ compilers are simply `~`, apparently because Xcode requires empty environment variables. Does that mean we can't control the clang version for these compilers? In any case, we're apparently using clang-17, not 18, so I need to update the label.

wdeconinck:
Indeed, the C/C++ compiler are provided by the OS / developer tools.
Homebrew package manager is based on this as well.
Fortran is not part of the OS and is installed via homebrew. CMake identifies it as AppleClang 17 so I think clang-17 is correct.
Thanks for spotting!

samhatfield:
Okay to merge then?

wdeconinck (approved):

PR #349, opened 2026-01-13
Fix gath spec bug

samhatfield:
As mentioned in https://github.com/ecmwf-ifs/ectrans/pull/339, there is a bug in the way we call `gath_spec` in the benchmark program, when checksums are dumped. The checksum is computed per global level, but spectral arrays are decomposed in the levels dimension, so this doesn't make sense. This bug leads to spurious checksum differences between benchmark runs with different `NPRTRV` values.

This PR fixes the bug and permits us to include `NPRTRV /= 1` cases in the test suite. I've included several modifications to the test suite and necessary changes to the CI to ensure reproducibility with these changes.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -1613,6 +1613,7 @@ subroutine dump_checksums_psp(filename, noutdump,       &
   call gath_spec(pspecg=gspfld(1:numfld,:), kfgathg=numfld, kto=[(1, i = 1, numfld)], kvset=ivset, &
     &            pspec=zspvor)
   if (myproc == 1) then
+    icrc = 0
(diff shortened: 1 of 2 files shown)


samhatfield:
[`macos-15` only has 3 cores](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories) :( So we can't run 4-task tests on that platform.

samhatfield:
For neatness I removed commits pertaining to 4-task tests from this PR and put them in a followup PR, #350. This PR should be merged first.

ddegrauwe (approved):
Thanks for spotting this! Looks good to me.

wdeconinck (approved):
Nice @samhatfield !

PR #352, opened 2026-01-14
Improve ectrans4py

samhatfield:
This PR implements a few improvements to ectrans4py, notably to `get_legendre_assets`. Most importantly, it allows one to call this function as many times as you like with different truncations, without having to know the value of "KMAXRESOL" in advance. It does this by tearing down every time, which allows you to call `SETUP_TRANS0` with each invocation. This may be necessary for eventual integration of the spectral transform in [anemoi-core](https://github.com/ecmwf/anemoi-core) which may use ectrans4py to generate Legendre polynomials.


samhatfield:
Ugh, what a mess. What I've done is make `get_legendre_assets` tear down the ecTrans context after its finished, so it can be called again without a clash of state. That works. The problem is that the other ectrans4py functions leave their state "alive", presumably so that they can be called multiple times without having to initialise each time. Makes sense, but it makes it impossible (as far as I can see) to call those functions in the same program as `get_legendre_assets`. That includes the ectrans4py test suite, which is why the tests fail.

samhatfield:
It would help if there was a "tear down" function for ectrans4py, i.e. an inverse for `spec_setup4py.F90`. That's not currently possible because the "is initialised" state is stored as a `SAVE` variable in `SPEC_SETUP4PY`. So I can't access the state from another subroutine.

samhatfield:
I will close this for now because I can't see a way forward without rewriting ectrans4py from scratch, and I don't want to upset those who are using it.

PR #355, opened 2026-02-04
Add API tests for SETUP_TRANS

samhatfield:
This PR adds API tests for `SETUP_TRANS`, `DIR_TRANS`, and `GPNORM_TRANS`.


samhatfield:
Actually, it's easier if I open one PR at a time. This PR will just add tests for SETUP_TRANS.

tests/trans/api/setup_trans/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+                         MPIxOMPS "0x1" "0x8" "1x1" "1x8" "2x1" "2x8"
+                         BACKENDS "cpu_sp" "cpu_dp" "gpu_sp" "gpu_dp" )
+
+# The FLT is not implemented for the GPU backend
+get_property( all_tests DIRECTORY PROPERTY TESTS )
+foreach( test IN LISTS all_tests )
+  if( (test MATCHES "flt") AND (test MATCHES "gpu") )
+    set_property( TEST "${test}" PROPERTY DISABLED TRUE )
+  endif()
+endforeach()
```
wdeconinck: Interesting pattern, which I think will show up more often.

Rather than having this boiler-plate, could we not add an argument to the `generate_api_tes_suite` macro?
e.g. the following EXCLUDES argument.
```cmake
generate_api_test_suite( SUITE_NAME setup_trans_test_suite
                         TESTS ${test_list}
                         WILL_FAIL_LIST ${will_fail_list}
                         MPIxOMPS "0x1" "0x8" "1x1" "1x8" "2x1" "2x8"
                         BACKENDS "cpu_sp" "cpu_dp" "gpu_sp" "gpu_dp"
                         EXCLUDES "flt.*gpu" ) # regex is used in `if ( test MATCHES regex )`
```
Rather than disable the test, you can just exclude it from being added.

tests/trans/api/setup_trans/setup_trans_test_suite.F90
```diff
@@ -0,0 +1,272 @@
...
+  ! Set up MPI
+  LUSE_MPI = DETECT_MPIRUN()
+  IF (LUSE_MPI) THEN
+    CALL MPL_INIT
+    NPROC = MPL_NPROC()
+  ELSE
+    NPROC = 1
+  ENDIF
+END SUBROUTINE SETUP_TEST
+
```
wdeconinck, line 55: Suggestions here:
You can make NPROC and LUSE_MPI module variables, being setup via SETUP_TEST.
I would add the setup_trans0 also in SETUP_TEST as it is common.
And I would add another `END_TEST` routine as well, which takes care of the TRANS_END and MPL_END.

wdeconinck (changes requested):
Nice work!!! I have some suggested changes (see other comments)

There is probably also an interplay with the `LEQ_REGIONS` flag in  SETUP_TRANS0.
For reduced global grids we usually have `LEQ_REGIONS=.TRUE.`, for LAM or regular grids I think `LEQ_REGIONS=.FALSE.` should be used.

Sidenote:  with the other work on MPI comms per trans handle, I think some of the SETUP_TRANS0 arguments for parallelisation should go into SETUP_TRANS, but that is the future. These tests here will be useful to ensure backwards compatibility when that is being developed.

There are many more cases to test here of course. You can check [transi_module.F90](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/transi/transi_module.F90) for a lot of options on how to call SETUP_TRANS.
The most important ones:

- `LDGRIDONLY` for a grid-point only setup
- `LDSPSETUPONLY` for a spectral only setup
- `LDSPLIT` (TRUE for a better load-balance, FALSE for cleaner domain partition edges)

tests/trans/api/setup_trans/setup_trans_test_suite.F90
samhatfield: Sounds good, I have done most of this. As for LUSE_MPI, maybe it's better just to remove this and call DETECT_MPIRUN every time?

tests/trans/api/setup_trans/setup_trans_test_suite.F90
```diff
@@ -0,0 +1,287 @@
...
+CONTAINS
+
+!---------------------------------------------------------------------------------------------------
+
+! Setup fixture
+SUBROUTINE SETUP_TEST(LUSE_MPI, LCALL_SETUP_TRANS0)
+  USE UTIL, ONLY: DETECT_MPIRUN
+  USE MPL_MODULE, ONLY: MPL_INIT, MPL_NPROC
+
+  LOGICAL, INTENT(OUT) :: LUSE_MPI
```
wdeconinck: I would suggest to add LUSE_MPI as a module variable. Essentially the module acts as the fixture, and the setup_test and end_test are the setup and teardown of the fixture.
LUSE_MPI is then an implementation that the actual tests do not need to concern itself with.
So not local variable LUSE_MPI in the test.

wdeconinck:
I just realised another important missing test is the capability for multiple resolutions simultaneously.

tests/trans/api/setup_trans/setup_trans_test_suite.F90
samhatfield: Yes very nice

tests/trans/api/setup_trans/CMakeLists.txt
samhatfield: Nice, I made this change. Not sure if there's a better way to do "if test matches any in list ${excludes}".

samhatfield:
> I just realised another important missing test is the capability for multiple resolutions simultaneously.

Done.

wdeconinck (approved):
Great!!! That looks so neat now.

tests/trans/api/CMakeLists.txt
```diff
@@ -84,6 +85,13 @@ function( generate_api_test_suite )
           continue()
         endif()
 
+        # If this test is in the exclusion list, skip it
+        foreach( exclude IN_LIST excludes )
```
wdeconinck:
```suggestion
        foreach( exclude ${excludes} )
```
No need for IN_LISTS or IN LISTS or such.

samhatfield:
GPU FLT tests are now being run again...

tests/trans/api/CMakeLists.txt
wdeconinck, line 96: This needs to be:
```cmake
set(skip_test FALSE)
foreach( exclude ${excludes} )
  if( test MATCHES ${exclude} )
      set(skip_test TRUE)
      break()
  endif()
endforeach()
if(skip_test)
  continue()
endif()
```
you were just continuing the inner loop.

samhatfield: Yes that's one issue, I didn't think of that. However it still won't work, because `test` doesn't contain the backend (nor MPIxOMP config). It's just the base name of the test, to be suffixed with this info. So we need to do the exclusion test on the full test name.

PR #357, opened 2026-02-05
Find files in bundle

DJDavies2:
This seems to work for me in a bundle.  However it is difficult for me to test outside of a bundle.  Does this work for you?

### DIFF
--- a/tests/trans/api/CMakeLists.txt
+++ b/tests/trans/api/CMakeLists.txt
@@ -57,7 +57,7 @@ function( generate_api_test_suite )
                             SOURCES
                                 ${suite_name}${suffix}.c
                                 ${suite_name}.F90
-                                ${CMAKE_SOURCE_DIR}/tests/trans/api/util.F90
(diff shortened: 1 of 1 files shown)


samhatfield:
Thanks for finding and solving this @DJDavies2, I was literally just about to open a bugfix PR myself.

samhatfield:
While your proposed change does resolve the configure-time error, I don't think it will work if `generate_api_test_suite` is invoked in a directory not one level below this CMakeLists.txt file.

samhatfield:
Having thought about it, I think the issue above is minor, and we already have this pattern in a few other places, so I'm going to approve this.

samhatfield (approved):

wdeconinck:
This was indeed a bug.
It should not have been `${CMAKE_SOURCE_DIR}`, which is the top-level project dir (i.e. the bundle dir in case of a bundle), but rather `${PROJECT_SOURCE_DIR}`, which is the directory of the current project.
Using a relative path as suggested here also works as a fix.

PR #359, opened 2026-02-10
Properly handle unsupported SETUP_TRANS GPU features

samhatfield:
Now that we have API tests for SETUP_TRANS, we should clarify which features are and aren't supported for GPUs. This PR adds abort statements when the following arguments are passed to SETUP_TRANS for GPU:
- `LDSPSETUPONLY`: currently this is ignored with a warning
- `LDLL`: currently this is handled with an abort followed by some pointless logic
- `LDGRIDONLY`: currently this is ignored with a warning

Personally I think the "warn and carry on" approach is only appropriate for arguments which don't have relevance for the GPU implementation, e.g. "use FFTW". When a feature is requested and the functionality just isn't implemented at all, we should abort (this is what we currently do when the FLT is requested for the GPU version).


wdeconinck:
Hi Sam,
I agree with the removal of pointless code after an abort. 👍

However the workarounds for LDSPSETUPONLY and LDGRIDONLY should not be deleted and not abort. Let me explain.
The use of LDSPSETUPONLY or LDGRIDONLY is to speed up `setup_trans` in terms of computation and memory when the other "space" is not needed.
Because a single "space" is not yet implemented for GPUs, and I would not want to abort because of this, the workaround which I implemented was to create a minimal "other space", taking not too much memory or computations.
This was the quickest workaround I could do rather than dive in deeper to actually implement it. It should probably be not too much work, following CPU code path logic, but hey, this was minimally invasive.

samhatfield:
Fair enough. So perhaps I should try to actually implement these features for the GPU version?

wdeconinck:
> Fair enough. So perhaps I should try to actually implement these features for the GPU version?

Indeed... 😇

wdeconinck (changes requested):
Please either restore the mentioned workarounds, or repurpose PR to implement GPU version.

wdeconinck:
@samhatfield I don't think it is as easy to implement it. with LDGRIDONLY you have to assume there's no NSMAX for instance, and for LDSPSETUPONLY no grid information (e.g. NDGL, NLOEN, domain decomposition etc.)
Importantly, no grid-to/from-spectral transform structures should be computed: no legpol etc.

The reason for these is:
- LDGRIDONLY: Only compute structures needed for a parallel domain decomposition. We should still be able call DIST_GRID and GATH_GRID and GPNORM.  Nothing related to spectral data can be called (INV_TRANS, DIR_TRANS, DIST_SPEC, GATH_SPEC, SPECNORM)
- LDSPSETUPONLY: Only compute structures needed for understanding spectral fields. We should still be able to call DIST_SPEC, GATH_SPEC and SPECNORM, but nothing related to grid (INV_TRANS, DIR_TRANS, DIST_GRID, GATH_GRID, GPNORM)

So a unit test should be able to verify that these still work.

The implementation as is currently in the develop branch makes a tiny GRID for LDSPSETUPONLY and a tiny NSMAX for LDGRIDONLY to minimise computations and storage, providing a workable version in each space (grid/spectral). I prefer to have that workaround than an abort with "not implemented" message.

samhatfield:
Yeah I started work and realised it was more complicated than I expected :/

PR #360, opened 2026-02-10
Add missing linker language to generate_api_test_suite

awnawab:
A small fix to `generate_api_test_suite` to restore the ability to link ectrans statically in IFS builds.


samhatfield:
Thanks @awnawab!

samhatfield (approved):

wdeconinck:
We should probably have a CI config for static builds.

PR #365, opened 2026-02-16
Fix feature for etrans

samhatfield:
Previously it wasn't possible to enable etrans at all when the GPU feature is enabled, even the CPU version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,10 +171,14 @@ ecbuild_add_option( FEATURE ETRANS
                     DESCRIPTION "Include Limited-Area-Model Transforms" )
 
 # Note: ETRANS GPU does not support OpenMP yet or FFT graphs yet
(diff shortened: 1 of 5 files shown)


samhatfield:
Note: now, when `ENABLE_ETRANS=ON` and `ENABLE_GPU=ON` and `ENABLE_OMP=ON` (which is how we configure ecTrans for AMD GPUs), a GPU version of etrans will *not* be built. That version doesn't yet support OpenMP offloading. Perhaps we should issue a warning?

wdeconinck:
> Note: now, when `ENABLE_ETRANS=ON` and `ENABLE_GPU=ON` and `ENABLE_OMP=ON` (which is how we configure ecTrans for AMD GPUs), a GPU version of etrans will _not_ be built. That version doesn't yet support OpenMP offloading. Perhaps we should issue a warning?

I think a warning is in order indeed! Nothing fatal though.

wdeconinck (approved):
Looks great!

PR #366, opened 2026-02-18
Fix compiler flags for ifx

awnawab:
A small fix to enable compilation with ifx.


samhatfield:
Nice, thanks @awnawab. Any idea how difficult it would be to add ifx testing to the CI?

samhatfield (approved):

cmake/ectrans_compile_options.cmake
```diff
@@ -34,6 +34,7 @@ elseif( CMAKE_Fortran_COMPILER_ID MATCHES "Cray" )
 elseif( CMAKE_Fortran_COMPILER_ID MATCHES "IntelLLVM" )
   ecbuild_add_fortran_flags("-march=core-avx2 -no-fma" BUILD BIT)
   ecbuild_add_fortran_flags("-fp-model precise -fp-speculation=safe")
+  set( NO_FORTRAN_MAIN_FLAG "-nofor-main")
```
samhatfield:
```suggestion
  set( NO_FORTRAN_MAIN_FLAG "-nofor-main" )
```

awnawab:
Thanks @samhatfield! It's not too bad at all, you can follow the logic for intel modern here in field_api: https://github.com/ecmwf-ifs/field_api/blob/main/.github/workflows/build.yml#L39

PR #372, opened 2026-02-26
Reproduce and fix vordiv_to_uv FPE reported in ecmwf/atlas#346

wdeconinck:
A test in atlas that uses `VORDIV_TO_UV` fails with a `FE_INVALID` FPE in `SUBROUTINE VD2UV`, see  ecmwf/atlas#346
The issue is only present for DEBUG builds, where uninitialised values are set to NaN.

This PR's goal is to reproduce and fix the issue using following approach:

- [X] Create test: A new test suite for vordiv_to_uv has been created with the failing test included
- [X] Enable floating point exceptions: A mechanism for enabling FPE (also on macOS) has been added and can be enabled for testing. Later, see if we can implement the FPE on macOS also in DR_HOOK and use DR_HOOK signal handling.
- [x] Fix test, setting uninitialised values to zero.


tests/trans/api/vordiv_to_uv/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+# API tests for SETUP_TRANS
+
+# Define list of SETUP_TRANS tests
+set( test_list
+  ectrans_test_trans_api_vordiv_to_uv_t1
+)
+
+# Declare tests that WILL_FAIL
+set( will_fail_list
```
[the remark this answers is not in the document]
wdeconinck: intention is to fail for now in draft mode

tests/trans/api/vordiv_to_uv/vordiv_to_uv_test_suite.F90
```diff
@@ -0,0 +1,91 @@
...
+
+  RSPVOR = 1.0_JPRB
+  RSPDIV = 1.0_JPRB
+
+  CALL SETUP_TEST
+  CALL VORDIV_TO_UV(PSPVOR=RSPVOR, PSPDIV=RSPDIV, PSPU=RSPU, PSPV=RSPV, KSMAX=TRUNCATION)
+
+  ! RSPU and RSPV contain NaNs if FPE trapping is not working.
+  write(0,*) 'PSPU =', RSPU
+  write(0,*) 'PSPV =', RSPV
```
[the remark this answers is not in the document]
wdeconinck: Won't do

tests/trans/api/fpe_trapping.cc
```diff
@@ -0,0 +1,195 @@
...
+#else
+static int ectrans_test_feenableexcept(unsigned int excepts) {
+    return 0;
+}
+static int ectrans_test_fedisableexcept(unsigned int excepts) {
+    return 0;
+}
+#endif
+
+[[noreturn]] void ectrans_test_signal_handler(int signum, siginfo_t* si, [[maybe_unused]] void* ucontext) {
```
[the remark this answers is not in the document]
wdeconinck: On non-apple-silicon it is unused so keep it.

tests/trans/api/vordiv_to_uv/vordiv_to_uv_test_suite.F90
```diff
@@ -0,0 +1,91 @@
...
+  CALL END_TEST
+
+  RET = 0
```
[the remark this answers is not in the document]
wdeconinck: Test should crash and never reach this when test fails.

src/trans/cpu/internal/prfi1b_mod.F90
```diff
@@ -111,7 +111,7 @@ SUBROUTINE PRFI1B(KM,PIA,PSPEC,KFIELDS,KFLDPTR)
 DO JFLD=1,2*KFIELDS
   PIA(1,JFLD) = 0.0_JPRB
   PIA(2,JFLD) = 0.0_JPRB
-  PIA(ILCM+3,JFLD) = 0.0_JPRB
+  PIA(ILCM+3:,JFLD) = 0.0_JPRB
```
samhatfield, line 114: The effect of this is that all elements of `PIA` not set by the logic above are zeroed. Why not remove this loop and simply add
```fortran
PIA(:,:) = 0.0_JPRB
```
to the start of the subroutine?

wdeconinck: In order to avoid going through all of the memory again for just a few values

src/trans/cpu/internal/vdtuv_mod.F90
```diff
@@ -109,13 +109,16 @@ SUBROUTINE VDTUV(KM,KFIELD,PEPSNM,PVOR,PDIV,PU,PV)
...
     IR = 2*J-1
+    II = IR+1
     DO JI=2,ISMAX+3-KM
       PU(JI,IR) = +&
        &ZN(JI+1)*ZEPSNM(JI)*ZLAPIN(JI+1)*PVOR(JI+1,IR)-&
        &ZN(JI-2)*ZEPSNM(JI-1)*ZLAPIN(JI-1)*PVOR(JI-1,IR)
       PV(JI,IR) = -&
        &ZN(JI+1)*ZEPSNM(JI)*ZLAPIN(JI+1)*PDIV(JI+1,IR)+&
        &ZN(JI-2)*ZEPSNM(JI-1)*ZLAPIN(JI-1)*PDIV(JI-1,IR)
+      PU(JI,II) = 0.0_JPRB
```
samhatfield, line 121:
```suggestion
      ! Imaginary components are always zero for KM = 0
      PU(JI,II) = 0.0_JPRB
```

tests/trans/api/vordiv_to_uv/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+)
+
+generate_api_test_suite( SUITE_NAME vordiv_to_uv_test_suite
+                         TESTS ${test_list}
+                         WILL_FAIL_LIST ${will_fail_list}
+                         MPIxOMPS 0x1 0x8
```
samhatfield, line 23: I guess we might as well add the full set of configs?

```suggestion
                         MPIxOMPS 0x1 0x8 1x1 1x8 2x1 2x8
```

samhatfield: Actually no. Let's complete the `VORDIV_TO_UV` test suite in another PR.

wdeconinck: This test does not involve MPI besides the setup, so perhaps overkill to add too many tests at this point.
Also we're not verifying the results. This was a reproducer which was crashing before in some cases with FE_INVALID.
More work, later, is would be to verify the results, which could also test with more MPI ranks.

tests/trans/api/util.F90
```diff
@@ -49,4 +49,14 @@ LOGICAL FUNCTION DETECT_MPIRUN()
   ENDIF
 END FUNCTION
 
+SUBROUTINE ENABLE_FPE()
+  USE, INTRINSIC :: ISO_C_BINDING, ONLY: C_INT
```
samhatfield:
```suggestion
```
Unnecessary import?

tests/trans/api/CMakeLists.txt
```diff
@@ -6,6 +6,29 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+
+cmake_push_check_state(RESET)
```
samhatfield, line 10:
```suggestion
# Check whether we have feenableexcept and fedisableexcept for controlling individual floating-point exceptions
cmake_push_check_state(RESET)
```

tests/trans/api/fpe_trapping.cc
```diff
@@ -0,0 +1,196 @@
+// This file takes care of enabling floating point exception trapping for the API tests,
```
samhatfield, line 9:
```suggestion
/* (C) Copyright 2026- ECMWF.

This software is licensed under the terms of the Apache Licence Version 2.0
which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
In applying this licence, ECMWF does not waive the privileges and immunities
granted to it by virtue of its status as an intergovernmental organisation
nor does it submit to any jurisdiction. */

// This file takes care of enabling floating point exception trapping for the API tests,
```

samhatfield (approved):
Nice job @wdeconinck 👍🏻

Approved with minor comments above.

Here's my summary:
- `ZIA` is declared in `LTINV` with dimensions `(R%NLEI1, KLEI2)`. The first dimension is the "tallest" column of a spectral array (containing all total wavenumbers) plus some padding for computing gradients with stencil operations. This array is filled by `PRFI1B`, but is only exclusively filled for `KM = 0`. For higher zonal wavenumbers, values go uninitialised which triggered errors for debug builds, as those values are set to NaN and thereby trigger floating-point exceptions. **Question: those values still shouldn't be touched, so where exactly was the FPE occuring? In `VD2UV`?** This PR makes sure all elements of `ZIA` have values, i.e. are at least zeroed.
- In `VDTUV`, U and V wind are computed from vorticity and divergence. For `KM = 0`, only the real elements of the U and V wind arrays were filled. The imaginary components were not filled. Those are by definition zero, so this PR does that explicitly.
- A basic test suite is added for `VORDIV_TO_UV` which should trigger the FPE when the above two arrays are not filled properly. Now that they are, the test passes. This suite can be expanded later to give better coverage of `VORDIV_TO_UV`.

tests/trans/api/util.F90
wdeconinck: indeed :)

src/trans/cpu/internal/vdtuv_mod.F90
wdeconinck: Good to document!

tests/trans/api/CMakeLists.txt
wdeconinck: I will add your suggested documentation.

wdeconinck:
> **Question: those values still shouldn't be touched, so where exactly was the FPE occuring? In `VD2UV`?** 

The problem was in this block:
https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/vd2uv_mod.F90#L139-L142

tests/trans/api/fpe_trapping.cc
wdeconinck: Done

tests/trans/api/CMakeLists.txt
wdeconinck: Done.

tests/trans/api/util.F90
wdeconinck: Done.

src/trans/cpu/internal/vdtuv_mod.F90
wdeconinck: Done

PR #375, opened 2026-03-02
Fix some bugs and tidy the benchmark program

samhatfield:
Here are a few small things that Claude spotted.

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -1664,13 +1664,14 @@ subroutine dump_checksums_psp_3a_2(filename, noutdump,  &
   integer(kind=jpim), intent(in) :: noutdump ! unit number for output file
   integer(kind=jpim), intent(in) :: jstep    ! time step
   integer(kind=jpim), intent(in) :: myproc   ! mpi rank
-  integer(kind=jpim), intent(in), optional :: nspec2g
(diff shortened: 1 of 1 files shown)


src/programs/ectrans-benchmark.F90
```diff
@@ -1314,6 +1316,10 @@ subroutine get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, n
     iarg = iarg + 1
   end do
 
+  if (icall_mode == 1 .and. ldump_values) then
+    call parsing_failed("The --dump-values option is not compatible with --callmode 1")
```
wdeconinck: Not standing in the way of "good to merge", but why is this not compatible?

wdeconinck (approved):

src/programs/ectrans-benchmark.F90
samhatfield: I just thought it would be too much effort to implement, but I realise it was actually quite straightforward! I think we should refactor the field dumping at some point though because it's a bit arbitrary. Do we really need to dump all fields when they're basically all treated the same by ecTrans?

wdeconinck: I don't think we need to dump-values at all, unless for debugging.
And indeed perhaps you want to be more selective on which fields or level.

PR #379, opened 2026-03-16
Ensure trans_set_mpi_comm compatibility with congruent comms

wdeconinck:
This PR fixes an issue of compatibility with using `trans_set_mpi_comm` with where the same comm as was already set-up before is set up. This is detected and then `trans_set_mpi_comm` is supposed to be a no-op.
It is possible these comms are however not detected correctly to be the same:

Typically with MPL_INIT the communicator used is `MPL_COMM_OML(1)`, created with `MPL_CREATE_LOCOMM`.
When created with `MPI_COMM_WORLD` argument, then ` MPL_COMM_OML(1)`  is a communicator which is by default **congruent** with MPI_COMM_WORLD but not necessarily the same.
The value returned by MPL_COMM_OML(1) is 3, whereas MPI_COMM_WORLD = 0.

NB. : Congruent means the same processes and same order, but not identical context. Messages sent cannot be exchanged between congruent communicators.

This is needed to avoid changing API in https://github.com/ecmwf/atlas/pull/346

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,7 +39,7 @@ ecbuild_enable_fortran( REQUIRED NO_MODULE_DIRECTORY )
 
 ### Find (optional) dependencies
 
-ecbuild_find_package( NAME fiat REQUIRED )
+ecbuild_find_package( NAME fiat VERSION 1.3.0 REQUIRED )
 
 # Inherit MPI feature from FIAT (if you don't want MPI, rebuild FIAT with ENABLE_MPI=OFF)
 set( HAVE_MPI ${fiat_HAVE_MPI} )
--- a/src/transi/transi_module.F90
+++ b/src/transi/transi_module.F90
@@ -46,7 +46,8 @@ use MPL_module, only: &
   MPL_SETDFLT_COMM, &
   MPL_COMM_OML, &
   LMPLUSERCOMM, &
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ ecbuild_enable_fortran( REQUIRED NO_MODULE_DIRECTORY )
 
 ### Find (optional) dependencies
 
-ecbuild_find_package( NAME fiat REQUIRED )
+ecbuild_find_package( NAME fiat VERSION 1.3.0 REQUIRED )
```
wdeconinck, line 42: `MPL_COMM_COMPARE` was introduced in fiat 1.3.0

mo-joshuacolclough:
Thank you for this!

samhatfield:
So to clarify: we **do** want to permit a communicator congruent with `MPL_COMM_OML(1)` to be passed to `trans_set_mpi_comm`, without error?

wdeconinck:
> So to clarify: we **do** want to permit a communicator congruent with `MPL_COMM_OML(1)` to be passed to `trans_set_mpi_comm`, without error?

Correct. In principle this means we don’t need to change anything with an already initialized MPL. This is a side effect of using MPL_COMM_OML.

samhatfield (approved):

PR #382, opened 2026-03-20
missing routine (horiz_field) in precision renamings

AlexandreMary:
Thanks @RyadElKhatibMF for spotting the miss
(cherry picked from commit 7318ac83492181daec9ec5aa87d4a01a55531bc3)


AlexandreMary:
@marsdeno as mentioned today

samhatfield:
Thanks @AlexandreMary! Did you mean to put this change in src/**etrans**/sedrenames.txt, nor src/trans/sedrenames.txt?

AlexandreMary:
Indeed, sorry ! Force-pushed to fix that

wdeconinck (approved):
Good spot.
If this all works nicely and it is urgent, we could immediately make a hotfix release 1.8.1, @AlexandreMary ?

AlexandreMary:
It is necessary for the pre-CY51, which is bundled with 1.6.2 at the moment (like 50R1), so no urgency.

PR #384, opened 2026-04-16
Enable GPU_STATIC and FIELD_API features in build-hpc

samhatfield:
PR https://github.com/ecmwf-ifs/ectrans/pull/314 begins the implementation of a Field API interface to ecTrans. This work must be tested using the build-hpc suite. But because the branch is hosted in an external repository, the build-hpc action is not taken from the source branch of the PR but rather the target branch (develop) (for security reasons). So, this PR enables the two features which enable the Field API feature to be tested in build-hpc. We commit them to develop, and then the build-hpc action run in PR #314 will inherit them. These two features should not have any effect when another branch is tested.


wdeconinck (approved):
Perfectly fine!

PR #386, opened 2026-04-21
Update NVHPC version to 26

samhatfield:
This PR updates the NVHPC version to 26 (currently 26.1 but we could consider 26.3).

At some point nvfortran with check bounds enabled became a lot stricter about how it handles zero-sized arrays. These arrays crop up in a few places in ecTrans, especially when running with NPRTRV /= 1 and with NPROC > 1. It's possible for a task to have no resident data for a spectral field if it hasn't been assigned any through IVSET. So we pass around arrays with shapes like `(0, n)`. Mathematically this shouldn't be a problem, as long as we index them appropriately (the alternative would be guards all over the place disabling subroutine calls when no local fields are present). But nvfortran now flags accesses to these arrays sometimes as "out of bounds", e.g. `PSP(:,:)` where the first dimension is zero sized.

So this PR attempts to resolve all of the offending cases as well. A few of my fixes are a bit hacky so I'm open to suggestions for better ways around this.


src/etrans/cpu/internal/eftdir_ctl_mod.F90
```diff
@@ -173,7 +173,7 @@ SUBROUTINE EFTDIR_CTL(KF_UV_G,KF_SCALARS_G,KF_GP,KF_FS,KF_GPB, &
 ! Transposition
 
 CALL GSTATS(158,0)
-CALL TRGTOL(ZGTF,KF_FS,KF_GP,KF_SCALARS_G,IVSET,KPTRGP,&
```
wdeconinck, line 176: Add a comment here and elsewhere such as:

> The trivial use of LBOUND and UBOUND in ZGTF argument was added to overcome a NVHPC 26 run-time problem when ZGTF is a zero-sized array, see https://github.com/ecmwf-ifs/ectrans/pull/386. Please do not remove.

src/programs/ectrans-lam-benchmark.F90
```diff
@@ -1646,7 +1646,9 @@ subroutine dump_checksums(filename, noutdump, &
         call crc64(gspfld(1:numfld,:), int(size(gspfld(1:numfld,:)) * kind(gspfld), 8), icrc)
         write(noutdump, '(a,"(",i0,") = ",z16.16)') "sp3d", jfld, icrc
       else
-        call egath_spec(kfgathg=numfld, kto=[(1, i = 1, numfld)], kvset=ivset, pspec=sp3d(:,:,jfld))
+        if (size(sp3d) > 0) then
+          call egath_spec(kfgathg=numfld, kto=[(1, i = 1, numfld)], kvset=ivset, pspec=sp3d(:,:,jfld))
+        endif
```
wdeconinck: This could be a problem causing an MPI deadlock if it is expected that this rank is going to be communicating within this routine. (SEND/RECEIVE or BARRIER etc..)
When size(sp3d) > 0 on some ranks and size(sp3d) == 0 on others.

wdeconinck (changes requested):

src/trans/cpu/internal/read_legpol_mod.F90
```diff
@@ -162,7 +162,7 @@ SUBROUTINE READ_LEGPOL
           CALL ABORT_TRANS('READ_LEGPOL:BYTES_IO_READ FAILED')
         ENDIF
         ALLOCATE(S%FA(IMLOC)%RPNMA(IDGLU,ILA))
-        S%FA(IMLOC)%RPNMA(:,:) = RESHAPE(ZBUF,(/IDGLU,ILA/))
+        IF (ISIZE > 0) S%FA(IMLOC)%RPNMA(:,:) = RESHAPE(ZBUF,(/IDGLU,ILA/))
```
dhaumont: When do you expect to have ISIZE < 0 ? Should'nt we implement a workaround in that case, or at least log a warning message ?

src/trans/cpu/internal/spnormc_mod.F90
```diff
@@ -47,7 +47,7 @@ SUBROUTINE SPNORMC(PSM,KFLD_G,KVSET,KMASTER,KSMAX,PGM)
 ITAG = 100
 
 IF (NPROC > 1.AND.MYPROC /= KMASTER) THEN
-  CALL MPL_SEND(PSM(:,:),KDEST=NPRCIDS(KMASTER),KTAG=ITAG,&
+  CALL MPL_SEND(PSM,KDEST=NPRCIDS(KMASTER),KTAG=ITAG,&
```
dhaumont: I believe PSM(:,: ) was required for some other compilers, otherwise it would not be there, no ?

src/trans/cpu/internal/read_legpol_mod.F90
samhatfield: ISIZE will never be negative but it can be zero. That's because IDGLU (the number of latitudes that are included in the Legendre transform for the given zonal wavenumber `IM`) can be zero. (In short: some latitudes don't have enough points to represent wavenumbers above a certain threshold without aliasing).

The case that triggers this is unusual: in the transi I/O test, a transform from a 320x161 lonlat grid to T511 spectral representation is performed. For this case modes above wavenumber 160 can't be represented. So for `IM > 160`, `IDGLU` is zero.

src/trans/cpu/internal/spnormc_mod.F90
samhatfield: Ah, interesting, didn't know that. I thought the coder was just following our style guide, which I think recommends always passing colon indices even when every element is requested.

src/trans/cpu/internal/read_legpol_mod.F90
dhaumont: Thanks for the explanation

src/trans/cpu/internal/spnormc_mod.F90
dhaumont: Not 100% sure, it was more a question. Maybe @ddegrauwe or @RyadElKhatibMF  will know better ?

src/programs/ectrans-lam-benchmark.F90
samhatfield: Damn, I think you're right. The other tricks (`LBOUND` etc.) didn't work here. Let me investigate further.

lukasm91:
@samhatfield  This is a known regression and should be fixed in 26.3.

This provides some context:
https://forums.developer.nvidia.com/t/a-bug-of-nvfortran-26-1-regarding-empty-arrays/358946/5

samhatfield:
> @samhatfield This is a known regression and should be fixed in 26.3.
> 
> This provides some context: https://forums.developer.nvidia.com/t/a-bug-of-nvfortran-26-1-regarding-empty-arrays/358946/5

😮 we were assuming 26.1 and 26.3 behave similarly. So maybe none of these changes are needed??

Edit: let's see... reverting all workarounds.

wdeconinck:
I thought I was testing all this with 26.3 as well, having these issues.

lukasm91:
It is at least worth testing... if you still see issues with 26.3, I am happy to also have a look. If it is a compiler bug, I would prefer fixing the compiler rather than adding workarounds :D

samhatfield:
> It is at least worth testing... if you still see issues with 26.3, I am happy to also have a look. If it is a compiler bug, I would prefer fixing the compiler rather than adding workarounds :D

Yes, issue still present in 26.3, see https://github.com/ecmwf-ifs/ectrans/actions/runs/24828523187/job/72670702893?pr=386.

lukasm91:
Thanks, I will have a look!

wdeconinck:
Thanks Lukas for being proactive about this! Always nicer to have a more robust compiler without workarounds.

samhatfield:
When 26.5 is released soon (next month perhaps) we will revive this PR.

wdeconinck:
@samhatfield could we just turn off bounds-checking for those compiler versions? Would that work for now?

samhatfield:
> @samhatfield could we just turn off bounds-checking for those compiler versions? Would that work for now?

I'm not actually sure where ecTrans gets its debug flags from: `-Mlarge_arrays -g -O0 -Mbounds -traceback`.

`-Mlarge_arrays` and `-traceback` are added explicitly by us, but who sets `-g -O0 -Mbounds`? Is it CMake itself? It's not ecBuild.

lukasm91:
Yes it comes from -DCMAKE_BUILD_TYPE=Debug
https://github.com/Kitware/CMake/blob/61510710b12ddff97c2c22e146cdfb399daa2dad/Modules/Compiler/PGI-Fortran.cmake#L13

samhatfield:
`-Mbounds` is no longer passed to (Debug, linux nvhpc-26.3). I guess that's because we now wipe all pre-defined flags and build the list from scratch and `-Mbounds` [isn't on the list](https://github.com/ecmwf-ifs/ectrans/blob/develop/cmake/ectrans_compile_flags.cmake#L12).

So I guess we can merge this? @wdeconinck @lukasm91

wdeconinck:
The move to "overwrite" has changed quite a bit what we have been doing, and indeed this PR should now be merged.

We need probably another PR that reinstates bounds checking, fpe trapping, init_snan etc,
And maybe be more aggressive on optimisations?
And since we're now requiring ecbuild 3.14, we have access to following variables defined in [ecbuild_compile_options](https://github.com/ecmwf/ecbuild/blob/develop/cmake/ecbuild_compile_options.cmake), at least for Fortran:
  - ``ECBUILD_Fortran_COMPILE_OPTIONS_REAL4`` : Convert all unqualified REALs to 32 bit (single precision)
  - ``ECBUILD_Fortran_COMPILE_OPTIONS_REAL8`` : Convert all unqualified REALs to 64 bit (double precision)
  - ``ECBUILD_Fortran_COMPILE_OPTIONS_CHECK_BOUNDS`` : Bounds checking compile options
  - ``ECBUILD_Fortran_COMPILE_OPTIONS_INIT_SNAN`` : Compile options to initialize REAL's with signaling NaN
  - ``ECBUILD_Fortran_COMPILE_OPTIONS_FPE_TRAP`` : Compile options to trap floating-point-exceptions
It will be nice to start using these, and extend the list. Note that these variables can be overridden in a toolchain!

samhatfield:
> The move to "overwrite" has changed quite a bit what we have been doing, and indeed this PR should now be merged.
> 
> We need probably another PR that reinstates bounds checking, fpe trapping, init_snan etc, And maybe be more aggressive on optimisations? And since we're now requiring ecbuild 3.14, we have access to following variables defined in [ecbuild_compile_options](https://github.com/ecmwf/ecbuild/blob/develop/cmake/ecbuild_compile_options.cmake), at least for Fortran:
> 
> * `ECBUILD_Fortran_COMPILE_OPTIONS_REAL4` : Convert all unqualified REALs to 32 bit (single precision)
> * `ECBUILD_Fortran_COMPILE_OPTIONS_REAL8` : Convert all unqualified REALs to 64 bit (double precision)
> * `ECBUILD_Fortran_COMPILE_OPTIONS_CHECK_BOUNDS` : Bounds checking compile options
> * `ECBUILD_Fortran_COMPILE_OPTIONS_INIT_SNAN` : Compile options to initialize REAL's with signaling NaN
> * `ECBUILD_Fortran_COMPILE_OPTIONS_FPE_TRAP` : Compile options to trap floating-point-exceptions
>   It will be nice to start using these, and extend the list. Note that these variables can be overridden in a toolchain!

Sounds good. Happy to do that in another PR.

wdeconinck (approved):

PR #388, opened 2026-04-24
WIP : Branch for reworking how the hipblas handle is created

PaulMullowney:
Move the hipblas handle creation into setup_trans and destroyed in trans_end. Attach the handle to tpm_fields_gpu.


samhatfield:
Looks good to me. Does "WIP" mean this is not finished?

src/trans/gpu/algor/hicblas_mod.F90
```diff
@@ -12,6 +12,20 @@ MODULE HICBLAS_MOD
 IMPLICIT NONE
 
 INTERFACE
+  SUBROUTINE HIP_BLAS_CREATE( &
+    HANDLE             &
+  &) BIND(C, NAME='hipblasCreate')
```
samhatfield: This symbol doesn't exist in CUDA contexts. We should create an `extern(C)` function in https://github.com/PaulMullowney/ectrans/blob/hipblas_handle_rework/src/trans/gpu/algor/hicblas_gemm.hip.cpp called `hicblasCreate` which simply calls `hipblasCreate`. `hip` symbols in this file are redefined as `cuda` symbols thanks to https://github.com/PaulMullowney/ectrans/blob/hipblas_handle_rework/src/trans/gpu/algor/hicblas.h.

src/trans/gpu/algor/hicblas_mod.F90
```diff
@@ -12,6 +12,20 @@ MODULE HICBLAS_MOD
...
+    USE ISO_C_BINDING, ONLY: C_PTR
+    TYPE(C_PTR), INTENT(INOUT) :: HANDLE
+  END SUBROUTINE HIP_BLAS_CREATE
+
+  SUBROUTINE HIP_BLAS_DESTROY( &
+    HANDLE             &
+  &) BIND(C, NAME='hipblasDestroy')
```
samhatfield: Same for this.

samhatfield:
It's not as straightforward as I imagined. Which prompts the question @PaulMullowney: what prompted this PR in the first place?

PaulMullowney:
> It's not as straightforward as I imagined. Which prompts the question @PaulMullowney: what prompted this PR in the first place?

Months ago, I was having all sorts of troubles on 300A with the handles. I will try to reproduce.

lukasm91:
I think the "problem" with this approach is that it assumes that there is only one "blas" handle type, which so far we did not assume (there can be several kind of handles, like one for cutlass, one for cublas, they are not exposed). The user of the library doesn't need to know that these handle even exist in the first place.

samhatfield:
> I think the "problem" with this approach is that it assumes that there is only one "blas" handle type, which so far we did not assume (there can be several kind of handles, like one for cutlass, one for cublas, they are not exposed). The user of the library doesn't need to know that these handle even exist in the first place.

I hadn't though of that, thanks for raising it. Also this PR is not really cleaner as the handle is still essentially a static variable managed by `TPM_FIELDS`. We'll only merge this if we really have to.

samhatfield:
Following work last week with @PaulMullowney we are able to build the develop branch on an MI300A successfully with ROCm-AFAR 22.3.0. So this PR is no longer needed and can be closed, right @PaulMullowney?

PaulMullowney:
> Following work last week with @PaulMullowney we are able to build the develop branch on an MI300A successfully with ROCm-AFAR 22.3.0. So this PR is no longer needed and can be closed, right @PaulMullowney?

Yes. I am able to build and run successfully and run on multiple ranks using the system modules for openmpi. I can reproduce the issue with a custom build of openmpi. I don't think this is worth diving into more.

PR #390, opened 2026-04-27
Make adjoint utils library platform dependent

samhatfield:
This library calls ecTrans externals which are NOT in the common library, so we need to be explicit about which version we want.

Very important when we have one of CPU or GPU features disabled.


wdeconinck (approved):
This looks good.
The CI failing in installing MPI with nvhpc and intel is strange. This should be unrelated to this PR though.

samhatfield:
We have passing tests on all GPU systems, and the failing tests have nothing to do with this branch as @wdeconinck points out, so I'm merging.

PR #393, opened 2026-05-05
Remove legacy compiler workarounds

samhatfield:
I am currently refactoring the ecTrans compile flags setting functionality in line with the convention laid down by ECMWF's internal FORGE project, described [here](https://confluence.ecmwf.int/display/~naan/IFS+compiler+flags) for those who have access. According to the convention, per-source-file compile flags should be provided in a separate file following a few guidelines. We are blessed in only having two source files with custom flags in ecTrans which are to accommodate two old compilers: Intel 18 and Cray 8.

Before I spend effort to do the plumbing to conform to the FORGE convention, I wanted to first check whether we even need these anymore. If we can delete these, then that will save some time.

The workaround for the Cray compiler is presumably targeting ECMWF's old Cray XC40 HPC system, taken offline 3 years ago. I wondered if the change might be relevant for DestinE digital twins on LUMI as well, but the source file in question is only used for data assimilation (adjoint code). Any thoughts @wdeconinck?

The Intel workaround targets Intel 18. ECMWF's current HPC only provides Intel >= 19 so we have no way to test whether this workaround is still necessary. Maybe @DJDavies2 could comment, since he raised this issue originally?


wdeconinck:
I'm t thinking this is OK, but we should in the README state which compilers are no longer supported.

samhatfield:
> I'm t thinking this is OK, but we should in the README state which compilers are no longer supported.

Good idea.

samhatfield:
Done.

I have been quite strict with some of the compiler versions. My reasoning is that we can't offer support if we don't have access to a machine with that compiler version. For CCE, for example, LUMI now only provides CCE 19.0.0 (I think maybe earlier versions can be accessed through module hacks).

wdeconinck:
The listed ecbuild version is too old. It should be updated to 3.8

wdeconinck (approved):
Approved pending the change in README regarding ecbuild.
Nice work!

README.md
```diff
@@ -15,6 +15,32 @@ License
 ecTrans is distributed under the Apache License Version 2.0.
 See `LICENSE` file for details.
 
+Requirements
+------------
+
+Generally, ecTrans has the following requirements:
+- [CMake](https://cmake.org/) >= 3.25
+- [ecBuild](https://github.com/ecmwf/ecbuild) >= 1.0.2
```
samhatfield:
```suggestion
- [ecBuild](https://github.com/ecmwf/ecbuild) >= 3.8
```

samhatfield:
I'm going to leave this open for a bit in case @DJDavies2 wants to object to the enforcement of Intel >= 19. I think he's the only one I know who might be using an earlier version still.

DJDavies2:
We don't try with Intel 18 anymore, only Intel 19.  For Cray we use 15.  So from my point of view this change is no problem.

samhatfield:
> We don't try with Intel 18 anymore, only Intel 19. For Cray we use 15. So from my point of view this change is no problem.

We're also increasing our minimum for CCE to 19, as mentioned in https://github.com/ecmwf-ifs/ectrans/issues/398. That doesn't mean we won't help you of course, but we do have a lack of platforms for testing earlier versions, so our support will be limited.

samhatfield:
The LUMI test is failing but LUMI is a mess at the moment so I'm not surprised. This PR is not related to support for LUMI so I am going to merge anyway.

PR #396, opened 2026-05-07
Remove compile flag overloads for ftinv/ftdir GPU

samhatfield:
I found a couple of other instances of `set_source_file_properties` which I'm skeptical of. These were introduced by [this commit](https://github.com/samhatfield/ectrans/commit/dfe7153c8b9e18cbb1a4bb54ad6bd9b5139202ad#diff-672173bb8f228a6159f3b3b58aaf4c4b04aa1c173778d6682cc2e2bb67fcd941R46-R47) 4 years ago. I can't think of a reason why you'd need these overrides, so let's just delete them and see how the build-hpc tests respond.


samhatfield:
Again, LUMI tests failing probably due to a timeout. I'm not able to even navigate to the folder where the job runs at the moment.

samhatfield:
build-hpc / ci-hpc (ac-gpu nvhpc) is passing, so I'm going to say "these compile flags weren't needed".

wdeconinck (approved):
Very good to remove this indeed. Good to be critical.

PR #397, opened 2026-05-07
Implement new IFS compile flags convention

samhatfield:
Builds on #393. That should be merged first.

This PR implements the new compile flags convention for IFS projects, specified [here](https://confluence.ecmwf.int/display/~naan/IFS+compiler+flags).

Since ecTrans no longer has per-source-file flags thanks to #393 (and hopefully also #396) this didn't involve much work.

At some point we should review the compile flag workarounds and determine which ones are still necessary, given the minimum compiler versions we specified in #393.


wdeconinck (approved):
Nice!

PR #404, opened 2026-05-19
Optionally device resident grid-point fields

awnawab:
This PR contributes an optional argument, which can be used to disable data-transfers for grid-point fields in order to keep them device resident across calls to `INV/DIR_TRANS`. Given we rely on the OpenACC/Cuda runtime API in FIELD_API (to handle discontiguous arrays), we cannot rely reliably on the OpenACC reference counters as is proposed in #131.


wdeconinck (approved):
@awnawab Thank you for this contribution. That will be very helpful
Copilot has some nitpicking on some missing documentation which is a good idea to do.
I will approve already regardless!

PR #405, opened 2026-05-19
Modernise TPM_FFTW

wdeconinck:
### Description

1. This contribution moves the calls to the legacy FFTW API (`SFFTW_*` and `DFFTW_*`) to the FFTW3 API (`FFTW_*` and `FFTWF_*`).
2. Further an extensive refactoring and documentation has been added in order to improve maintenance, and especially clarity. Duplicate code paths in EXEC_EFFTW and EXEC_FFTW have been removed. The only difference was the transposition of the PREEL argument.
3. Magic constants like 123456 and 999999 have extracted as module parameters.
4. Generic interfaces for FFTW functions have been added which avoid code like `IF(JPRB==JPRD)` obfuscating the control flow.

### Contributor Declaration


wdeconinck:
It came to my attention that there is a FFTW API for FFT992 here: https://github.com/pmarguinaud/fftw992 which does not implement the modern FFTW API. I will adapt this PR to still use the legacy API.

src/trans/cpu/internal/tpm_fftw.F90
```diff
@@ -241,267 +376,322 @@ SUBROUTINE DESTROY_PLANS_FFTW
...
 
 #include "abor1.intfb.h"
 
-IF (LHOOK) CALL DR_HOOK('EXEC_EFFTW',0,ZHOOK_HANDLE)
+IF (LHOOK) CALL DR_HOOK(CDNAME,0,ZHOOK_HANDLE)
 
-IF ( (KTYPE /= -1) .AND. (KTYPE /=1) ) THEN
-  CALL ABOR1('TPM_FFTW:EXEC_EFFTW : WRONG VALUE KTYPE')
+IF ( (KTYPE /= NTYPE_R2C) .AND. (KTYPE /= NTYPE_C2R) ) THEN
+  CALL ABOR1('TPM_FFTW:'//TRIM(CDNAME)//' : WRONG VALUE KTYPE')
```
samhatfield: Should we be using `ABORT_TRANS`?

src/trans/cpu/internal/tpm_fftw.F90
```diff
@@ -41,34 +42,166 @@ MODULE TPM_FFTW
 PUBLIC CREATE_PLAN_FFTW, DESTROY_PLAN_FFTW, DESTROY_PLANS_FFTW, INIT_PLANS_FFTW, &
       & FFTW_RESOL, TW, EXEC_FFTW, EXEC_EFFTW
```
samhatfield: I think we can make some of these `PRIVATE`, e.g. `CREATE_PLAN_FFTW`.

src/trans/cpu/internal/tpm_fftw.F90
samhatfield:
```suggestion
```
Might as well get rid of all these.

samhatfield:
```suggestion
IF( CURR_FFTW_PLAN%NPLAN_ID /= NPLAN_ID_UNINITIALISED )THEN
WRITE(*,'("CREATE_PLAN_FFTW.4: PLAN_ID=",I10)')CURR_FFTW_PLAN%NPLAN_ID
CALL ABOR1('CREATE_PLAN_FFTW.4: NPLAN_ID /= NPLAN_ID_UNINITIALISED')
ENDIF
```

src/trans/cpu/internal/tpm_fftw.F90
wdeconinck: I will try to see and make it private if possible. That is a good suggestion.

src/trans/cpu/internal/tpm_fftw.F90
wdeconinck: Yes the magic number should be replaced. Oversight! Good catch!

samhatfield:
Just had a brief look. Looks good and very nice work. Let me know when you want a full review by marking the PR as ready.

src/trans/cpu/internal/tpm_fftw.F90
wdeconinck: Done.

src/trans/cpu/internal/tpm_fftw.F90
wdeconinck: Donel

wdeconinck: Done.

wdeconinck:
TPM_FFTW has been adapted to use the LEGACY interface again.
There is a hardcoded `#define LEGACY_FFTW_INTERFACE 1` in `tpm_fftw.F90` which can be toggled 0/1.
Once we're confident to switch to the MODERN FFTW interfaces, we can easily toggle this, and remove the added code.

wdeconinck:
@samhatfield This is now ready for final review.
@pmarguinaud I have verified this compiles and runs with https://github.com/pmarguinaud/fftw992

src/trans/cpu/internal/tpm_fftw.F90
```diff
@@ -127,106 +308,89 @@ SUBROUTINE CREATE_PLAN_FFTW(KPLAN,KTYPE,KN,KLOT)
...
-      TW%FFTW_PLANS(KN)%NPLAN_ID=999999
+      CALL TPM_FFTW_DESTROY_PLAN(TW%FFTW_PLANS(KN)%NPLAN)
+      TW%FFTW_PLANS(KN)%NPLAN_ID=NPLAN_ID_DESTROYED
+       ! mark the plan record as destroyed but keep it in the list to be recycled by the next plan creation with the same KN,
+       ! which avoids costly deallocation and reallocation of the FFTW_PLAN record and preserves the linked list structure
+       ! without needing to update any pointers
       START_FFTW_PLAN=>TW%FFTW_PLANS(KN)
       TW%FFTW_PLANS(KN)=TW%FFTW_PLANS(KN)%NEXT_PLAN
       ! DEALLOCATE(START_FFTW_PLAN)
       TW%N_PLANS(KN)=TW%N_PLANS(KN)-1
```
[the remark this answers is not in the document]
samhatfield: I agree with Copilot that there's something not quite right with the cache eviction mechanism which you have copied over from the previous `TPM_FFTW` (so not blaming you @wdeconinck).

This code is about 10 years old so if it is wrong, it can't be in a catastrophic way. I suspect that the cache eviction mechanism has probably never been exercised.

Firstly, what is `NPLAN_ID_DESTROYED` for? It doesn't do anything.

Secondly the head node is never deallocated -> Line 328 has always been commented out.

Thirdly the new head node is declared through a naive copy of a derived type (line 327). The derived type in question (`FFTW_PLAN`) has a pointer member (`NEXT_PLAN`). How is this being treated by the copy? Seems dangerous to me.

All in all I think we need to take a thorough look at the cache eviction mechanism to make sure it's working, including testing it. However it clearly isn't urgent, so maybe we should leave that to a future PR?

samhatfield (changes requested):
Approved, subject to resolution of comment above.

src/trans/cpu/internal/tpm_fftw.F90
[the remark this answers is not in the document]
wdeconinck: I think `NPLAN_ID` is not used at all besides for some internal error checking. I feel we can just remove it.
I have not dared touching this at this point, to introduce minimal changes.
Perhaps we can tackle this in a follow-up PR?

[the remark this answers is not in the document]
samhatfield: Yes let's do it in another PR.

samhatfield (approved):

wdeconinck:
Thanks for the review. Merging.

PR #408, opened 2026-05-28
Check if fftw_resol is allocated before attempting to deallocate

pardallio:
### Description

if etrans_end is called and fftw_resol is not allocated it will fail with
```

forrtl: severe (153): allocatable array or pointer is not allocated
Image              PC                Routine            Line        Source             
gl_DP              0000000000430743  Unknown               Unknown  Unknown
libectrans_lam_dp  0000150422B4E034  etrans_end_dp_            111  etrans_end.F90
```

this check avoids this error
### Contributor Declaration


dhaumont:
@ddegrauwe Is it a reason why there is no test on FFTW_RESOL, like for the others arrays ?

ddegrauwe (approved):
Looks good to me. The same is done for the global transforms.
Out of curiositiy: how do you trigger this bug? By calling etrans_end after calling trans_end or something?

dhaumont:
There was a workaround for this in gl: an extra allocation of this array was performed, such that the deallocation won't crash.
But the workaround wont't compile anymore with the latest version of ectrans, which renamed the module with _sp and _dp

dhaumont:
@wdeconinck @samhatfield Will this be possible to merge this quickly ? I believe it is safe

ddegrauwe:
> There was a workaround for this in gl: an extra allocation of this array was performed, such that the deallocation won't crash. But the workaround wont't compile anymore with the latest version of ectrans, which renamed the module with _sp and _dp

Ok, I checked the code of gl, and there in fact are several calls to etrans_end from the same program. Should be easy to fix there, but a safety in ectrans doesn't harm.

dhaumont:
> > There was a workaround for this in gl: an extra allocation of this array was performed, such that the deallocation won't crash. But the workaround wont't compile anymore with the latest version of ectrans, which renamed the module with _sp and _dp
> 
> Ok, I checked the code of gl, and there in fact are several calls to etrans_end from the same program. Should be easy to fix there, but a safety in ectrans doesn't harm.

Not that easy, you have to introduce a lot of ifdefs to manage sp and dp

pardallio:
> Looks good to me. The same is done for the global transforms. Out of curiositiy: how do you trigger this bug? By calling etrans_end after calling trans_end or something?

I don't know exactly why this happens. I believe there is probably some fuzzy login in gl calling etrans_end twice, but I couldn't find it

wdeconinck (approved):
Perfectly fine!

dhaumont:
Thanks a lot @wdeconinck and @ddegrauwe  for the swift reaction ! Highly appreciated

Issue #392, opened 2026-05-05
CPU SP performance imbalance (IM=0 path) linked to allocations and SP→DP copies in LEDIR


CarlosPenaDePedro:
## Context

- Organisation: Barcelona Supercomputing Center (BSC)
- Project: DestinE ClimateDT

## Summary

We observe a significant OpenMP thread imbalance in the single-precision CPU version of `ectrans` when running with `IM=0`.

The issue was first detected on MN5 GPP and also appears on LUMI-C, but the reproducer below is for MN5.

The imbalance becomes visible when GEMM is fast enough, e.g. with:

```bash
export MKL_CBWR=AUTO
```

On MN5 this selects AVX512 behavior and makes the SGEMM calls sufficiently fast that the overhead from temporary allocation and single-to-double copies in `ledir_mod.F90` becomes a relevant fraction of runtime.

## Versions

```text
ecbuild: 3.8.5
fiat:    1.4.1
ectrans: 1.8.0
```

## Platform

```text
System: MN5 GPP partition
Nodes: 100
Cores per node: 112
Compiler: Intel ifort
MPI: Intel MPI 2021.10.0
MKL: 2023.2.0
FFTW: 3.3.10
```

Modules:

```bash
module load intel/2023.2.0 impi/2021.10.0 mkl/2023.2.0 ucx/1.16.0 fftw/3.3.10 cmake
export FC=ifort
```

## Reproducer

Submission script:

```bash
#!/bin/bash
#SBATCH -J ectrans-tco2559
#SBATCH --qos=gp_bsces
#SBATCH --account=bsc32
#SBATCH -N 100
#SBATCH --ntasks-per-node=14
#SBATCH --cpus-per-task=8
#SBATCH --hint=nomultithread
#SBATCH --time=00:30:00

export OMP_NUM_THREADS=8
export OMP_PLACES=cores
export OMP_PROC_BIND=close

module load intel/2023.2.0 impi/2021.10.0 mkl/2023.2.0 ucx/1.16.0 fftw/3.3.10

export DR_HOOK=1
export DR_HOOK_OPT=prof

# Important: needed for GEMM to be fast enough for the imbalance to show clearly.
# On MN5 this selects AVX512 behavior.
export MKL_CBWR=AUTO

srun /gpfs/scratch/ehpc01/bsc032799/ectrans-1.8/ectrans/bin/ectrans-benchmark-cpu-sp \
  --truncation 1279 \
  --vordiv \
  --scders \
[...]

PR #407, opened 2026-05-27
Specify compiler flags explicitly

samhatfield:
https://github.com/ecmwf-ifs/ectrans/pull/397 implemented new conventions regarding compiler flags, and how they are handled by ecBuild. A call to `ecbuild_override_compiler_flags` was added to the top level ecTrans CMakeLists.txt. However I misunderstood the scope of these changes. This call was _supposed_ to wipe all flags previously declared, including those set by ecBuild. It didn't because of a bug in ecBuild at the time. With the release of ecBuild 3.14.2 (incorporating https://github.com/ecmwf/ecbuild/pull/145), this bug was fixed and default compiler flags declared by ecBuild were totally wiped by the call to `ecbuild_override_compiler_flags`. This triggered some GitHub-hosted actions to fail, specifically for the Intel and Nvidia compilers. For Intel, some tests are stack-hungry so will segfault if you don't provide `-heap-arrays`. For Nvidia, we got the same old math dispatching issue.

This PR resolves this issue by explicitly declaring default compiler flags within ecTrans itself, replicating what was previously done by ecBuild. Defaults are provided for several build types which should be valid for all compilers.

I've also refactored the `-tp=haswell` flag passing for Nvidia by having it passed through a toolchain file rather than through environment variables. The latter doesn't work and I'm not sure why, but in any case I think a toolchain file is tidier.

Resolves https://github.com/ecmwf-ifs/ectrans/issues/403.

PR #409, opened 2026-05-28
Ledir IM=0 DP matrix caching 

CarlosPenaDePedro:
### Description

This PR adds persistent double-precision caches for the KM=0 DGEMM path in LEDIR.

Specifically, it introduces persistent allocatable arrays:

RPNMA_DGEMM
RPNMS_DGEMM

to avoid repeated allocations/recomputations in the DGEMM execution path for the KM=0 case. This is intended to reduce overhead and help improve load imbalance observed in this path.

More information in issue #392 

The change passes the GitHub Actions test suite.

### Contributor Declaration


samhatfield:
Thanks for this amazing work @CarlosPenaDePedro - we will start the review when we have time, but definitely well before the October deadline you mention 🙏🏻

src/trans/cpu/internal/ledir_mod.F90
```diff
@@ -115,64 +115,78 @@ SUBROUTINE LEDIR(KM,KMLOC,KFC,KIFC,KSL,KDGLU,KLED2,PAIA,PSIA,POA1,PW)
...
                 ENDDO
              ENDDO
-             DO I3=1,I1
-                DO I4=1,I2
-                   ZRPNMA(I3,I4) = S%FA(KMLOC)%RPNMA(I3,I4)
-                END DO
-             END DO
-             CALL GEMM('T','N',ILA,KIFC,KDGLU,1.0_JPRD,ZRPNMA,KDGLU,&
+
+             CALL GEMM('T','N',ILA,KIFC,KDGLU,1.0_JPRD,S%RPNMA_DGEMM,KDGLU,&
```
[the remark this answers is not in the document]
samhatfield: Oof, and that's why we use Copilot. I will take a look at this.

src/trans/cpu/internal/tpm_flt.F90
```diff
@@ -42,6 +42,11 @@ MODULE TPM_FLT
 
 TYPE FLT_TYPE_WRAP
 TYPE(FLT_TYPE),ALLOCATABLE :: FA(:)
+
+! Persistent double-precision caches for the KM=0 DGEMM path in LEDIR
+REAL(KIND=JPRD), ALLOCATABLE :: RPNMA_DGEMM(:,:)
+REAL(KIND=JPRD), ALLOCATABLE :: RPNMS_DGEMM(:,:)
```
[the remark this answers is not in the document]
samhatfield: I think this is also genuine. Could you implement that change @CarlosPenaDePedro?

samhatfield:
Don't worry about the failing tests. Once https://github.com/ecmwf-ifs/ectrans/pull/407 is merged, you can then rebase this again against develop and that should resolve them.

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -690,14 +692,30 @@ SUBROUTINE SULEG
...
+
+          LLCACHEDGEMM = (.NOT. LLDOUBLE) .AND. (IM == 0)
+
+          IF (LLCACHEDGEMM) THEN
+            IF (ALLOCATED(S%RPNMA_DGEMM)) THEN
+              IF (SIZE(S%RPNMA_DGEMM,1) /= IDGLU .OR. &
+                  SIZE(S%RPNMA_DGEMM,2) /= ILA) THEN
+                DEALLOCATE(S%RPNMA_DGEMM)
+              ENDIF
+            ENDIF
```
samhatfield: Is this logic strictly necessary? `S` is a pointer to the "current resolution's" element in `FLT_RESOL`. It should be "fresh" and uninitialised.

src/trans/cpu/internal/tpm_flt.F90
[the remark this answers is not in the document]
CarlosPenaDePedro: Yes!

src/trans/cpu/internal/suleg_mod.F90
CarlosPenaDePedro: Good point. I added that check just to be safe in case this object could be reused, but if S is always a fresh FLT_RESOL entry here, then it isn't needed. I'll simplify it.

samhatfield:
Could you rebase against develop again @CarlosPenaDePedro? That should fix the failing tests.

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -710,13 +720,19 @@ SUBROUTINE SULEG
...
+                  ENDIF
                 ENDDO
               ENDDO
             ENDIF
           ELSE
             DO JGL=1,IDGLU
               DO JI=1,ILA
                 S%FA(IMLOC)%RPNMA(JGL,ILA-JI+1)=ZRCVBUFV((JGL-1)*ILA+JI,JSETV)
+                IF (LLCACHEDGEMM) THEN
+                  S%RPNMA_DGEMM(JGL,ILA-JI+1)=REAL(ZRCVBUFV((JGL-1)*ILA+JI,JSETV),JPRD)
```
samhatfield:
```suggestion
                  S%RPNMA_DGEMM(JGL,ILA-JI+1)=ZRCVBUFV((JGL-1)*ILA+JI,JSETV)
```
`ZRCVBUFV` is already double precision so no cast required.

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -750,13 +774,19 @@ SUBROUTINE SULEG
...
+                  ENDIF
                 ENDDO
               ENDDO
             ENDIF
           ELSE
             DO JI=1,ILA
               DO JGL=1,IDGLU
                 S%FA(IMLOC)%RPNMA(JGL,JI) = F%RPNM(ISL+JGL-1,D%NPMS(IM)+IA+(JI-1)*2)
+                IF (LLCACHEDGEMM) THEN
+                  S%RPNMA_DGEMM(JGL,JI) = REAL(F%RPNM(ISL+JGL-1,D%NPMS(IM)+IA+(JI-1)*2),JPRD)
```
samhatfield:
```suggestion
                  S%RPNMA_DGEMM(JGL,JI) = F%RPNM(ISL+JGL-1,D%NPMS(IM)+IA+(JI-1)*2)
```

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -989,13 +1027,19 @@ SUBROUTINE SULEG
...
+                  ENDIF
                 ENDDO
               ENDDO
             ENDIF
           ELSE
             DO JGL=1,IDGLU
               DO JI=1,ILS
                 S%FA(IMLOC)%RPNMS(JGL,ILS-JI+1)=ZRCVBUFV((JGL-1)*ILS+JI,JSETV)
+                IF (LLCACHEDGEMM) THEN
+                  S%RPNMS_DGEMM(JGL,ILS-JI+1)=REAL(ZRCVBUFV((JGL-1)*ILS+JI,JSETV),JPRD)
```
samhatfield:
```suggestion
                  S%RPNMS_DGEMM(JGL,ILS-JI+1)=ZRCVBUFV((JGL-1)*ILS+JI,JSETV)
```

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -1029,13 +1080,19 @@ SUBROUTINE SULEG
...
+                  ENDIF
                 ENDDO
               ENDDO
             ENDIF
           ELSE
             DO JI=1,ILS
               DO JGL=1,IDGLU
                 S%FA(IMLOC)%RPNMS(JGL,JI) = F%RPNM(ISL+JGL-1,D%NPMS(IM)+IS+(JI-1)*2)
+                IF (LLCACHEDGEMM) THEN
+                  S%RPNMS_DGEMM(JGL,JI) = REAL(F%RPNM(ISL+JGL-1,D%NPMS(IM)+IS+(JI-1)*2),JPRD)
```
samhatfield:
```suggestion
                  S%RPNMS_DGEMM(JGL,JI) = F%RPNM(ISL+JGL-1,D%NPMS(IM)+IS+(JI-1)*2)
```

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -710,13 +720,19 @@ SUBROUTINE SULEG
               DO JGL=1,IDGLU
                 DO JI=1,ILA
                   S%FA(IMLOC)%RPNMA(JGL,ILA-JI+1)=ZRCVBUFV((JGL-1)*ILA+JI,JSETV)
+                  IF (LLCACHEDGEMM) THEN
+                    S%RPNMA_DGEMM(JGL,ILA-JI+1)=REAL(ZRCVBUFV((JGL-1)*ILA+JI,JSETV),JPRD)
+                  ENDIF
```
samhatfield: I don't believe this is needed. The `IF ( S%LUSEFLT .AND. ILA > ITHRESHOLD )` is for handling zonal wavenumbers lower than the threshold for activation of the fast Legendre transform. The FLT doesn't do a simple matrix multiplication so neither `RPNMA` nor `RPNMA_DGEMM` are used. There is logic for "keeping" `RPNMA` but only for unrelated functionality which allows one to dump the polynomials to disc. That's my understanding, anyway. It's quite complicated. I say, let's just delete these lines.

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -750,13 +774,19 @@ SUBROUTINE SULEG
               DO JI=1,ILA
                 DO JGL=1,IDGLU
                   S%FA(IMLOC)%RPNMA(JGL,JI) = F%RPNM(ISL+JGL-1,D%NPMS(IM)+IA+(JI-1)*2)
+                  IF (LLCACHEDGEMM) THEN
+                    S%RPNMA_DGEMM(JGL,JI) = REAL(F%RPNM(ISL+JGL-1,D%NPMS(IM)+IA+(JI-1)*2),JPRD)
+                  ENDIF
```
samhatfield:
```suggestion
```
Similar to above.

src/trans/cpu/internal/suleg_mod.F90
samhatfield:
```suggestion
```

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -989,13 +1027,19 @@ SUBROUTINE SULEG
               DO JGL=1,IDGLU
                 DO JI=1,ILS
                   S%FA(IMLOC)%RPNMS(JGL,ILS-JI+1)=ZRCVBUFV((JGL-1)*ILS+JI,JSETV)
+                  IF (LLCACHEDGEMM) THEN
+                    S%RPNMS_DGEMM(JGL,ILS-JI+1)=REAL(ZRCVBUFV((JGL-1)*ILS+JI,JSETV),JPRD)
+                  ENDIF
```
samhatfield:
```suggestion
```
Similar to above.

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -1029,13 +1080,19 @@ SUBROUTINE SULEG
               DO JI=1,ILS
                 DO JGL=1,IDGLU
                   S%FA(IMLOC)%RPNMS(JGL,JI) = F%RPNM(ISL+JGL-1,D%NPMS(IM)+IS+(JI-1)*2)
+                  IF (LLCACHEDGEMM) THEN
+                    S%RPNMS_DGEMM(JGL,JI) = REAL(F%RPNM(ISL+JGL-1,D%NPMS(IM)+IS+(JI-1)*2),JPRD)
+                  ENDIF
```
samhatfield:
```suggestion
```
Similar to above.

samhatfield:
Todo:
- [x] Rebase against develop
- [x] Approve https://github.com/CarlosPenaDePedro/ectrans/pull/1 which tries to mitigate repeated code and also some style fixes to `LEDIR`
- [x] Check all comments above
- [x] Verify that all checks pass
- [x] For @samhatfield, check potential clash with `READ_LEGPOL` code path

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -690,14 +692,22 @@ SUBROUTINE SULEG
...
         DO JSETV=1,IPRTRV
           IMLOC=JMLOC+JSETV-1
           IM = D%MYMS(IMLOC)
           ISL = MAX(R%NDGNH-G%NDGLU(IM)+1,1)
           IA  = 1+MOD(R%NSMAX-IM+2,2)
           ILA = (R%NSMAX-IM+2)/2
           IDGLU = MIN(R%NDGNH,G%NDGLU(IM))
+
+          LLCACHEDGEMM = (.NOT. LLDOUBLE) .AND. (IM == 0)
+
```
samhatfield:
```suggestion
```

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -730,14 +743,22 @@ SUBROUTINE SULEG
...
```
samhatfield:
```suggestion
```

src/trans/cpu/internal/suleg_mod.F90
```diff
@@ -969,14 +993,22 @@ SUBROUTINE SULEG
...
           IMLOC=JMLOC+JSETV-1
           IM = D%MYMS(IMLOC)
           ISL = MAX(R%NDGNH-G%NDGLU(IM)+1,1)
           IS  = 1+MOD(R%NSMAX-IM+1,2)
           ILS = (R%NSMAX-IM+3)/2
           IDGLU = MIN(R%NDGNH,G%NDGLU(IM))
+
+          LLCACHEDGEMM = (.NOT. LLDOUBLE) .AND. (IM == 0)
+
+
```
samhatfield:
```suggestion
```

samhatfield:
All tests passed. Well done @CarlosPenaDePedro, you are off the critical path and can now relax 😝

I still need to take a closer look at the codepath for reading / writing the Legendre polynomials to see if there's a clash.

Also I'd like to repeat my benchmarking from earlier as a sanity check. Load balance is one thing, but it would be good to check that there is no negative unexpected impact on overall wall times etc. I'm travelling at the moment but hopefully next week I'll find some time.

samhatfield:
The issue Copilot identifies above is the following:

ecTrans can read pre-computed Legendre polynomials from file or a memory buffer. In this case, the computation of the polynomials is skipped. That includes the allocation and initialisation of `S%RPNMA_DGEMM` and `S%RPNMS_DGEMM`. In that case, `LEDIR` will segfault because those arrays are still accessed.

The proper fix is to make sure the cached arrays are initialised even when polynomials are read from file or buffer. Rather than have the double-precision IM=0 arrays written to and read from disk, I think we should simply copy the single-precision polynomials which are read into those cached arrays. This will incur a single -> double upcast which is different to what we have now, where the double-precision source arrays for the polynomials are stored directly in the cache. Then, we only have to make a slight modification in `READ_LEGPOL_MOD`.

It's actually closer to what we do in the develop branch. There in `LEDIR` we copy the stored single-precision polynomials into a double-precision buffer (so again, a single -> double upcast). This works, and implies that the problem with the IM=0 mode is not that the polynomials are too low in precision but rather the GEMM itself is too low in precision.

CarlosPenaDePedro:
In that case we could to keep the behaviour consistent across setup paths, we could also populate the SULEG `S%RPNMA_DGEMM` and `S%RPNMS_DGEMM`  by upcasting the already stored single-precision RPNMA/RPNMS arrays, instead of filling it directly from the double-precision source values. This would make the SULEG and READ_LEGPOL paths behave the same way, and should also be closer to the previous implementation, where LEDIR promoted the single-precision matrix to double at runtime.


> The issue Copilot identifies above is the following:
> 
> ecTrans can read pre-computed Legendre polynomials from file or a memory buffer. In this case, the computation of the polynomials is skipped. That includes the allocation and initialisation of `S%RPNMA_DGEMM` and `S%RPNMS_DGEMM`. In that case, `LEDIR` will segfault because those arrays are still accessed.
> 
> The proper fix is to make sure the cached arrays are initialised even when polynomials are read from file or buffer. Rather than have the double-precision IM=0 arrays written to and read from disk, I think we should simply copy the single-precision polynomials which are read into those cached arrays. This will incur a single -> double upcast which is different to what we have now, where the double-precision source arrays for the polynomials are stored directly in the cache. Then, we only have to make a slight modification in `READ_LEGPOL_MOD`.
> 
> It's actually closer to what we do in the develop branch. There in `LEDIR` we copy the stored single-precision polynomials into a double-precision buffer (so again, a single -> double upcast). This works, and implies that the problem with the IM=0 mode is not that the polynomials are too low in precision but rather the GEMM itself is too low in precision.

samhatfield:
> In that case we could to keep the behaviour consistent across setup paths, we could also populate the SULEG `S%RPNMA_DGEMM` and `S%RPNMS_DGEMM` by upcasting the already stored single-precision RPNMA/RPNMS arrays, instead of filling it directly from the double-precision source values. This would make the SULEG and READ_LEGPOL paths behave the same way, and should also be closer to the previous implementation, where LEDIR promoted the single-precision matrix to double at runtime.

Yes, that's exactly what I was thinking.

src/trans/cpu/internal/ledir_mod.F90
```diff
@@ -188,68 +202,82 @@ SUBROUTINE LEDIR(KM,KMLOC,KFC,KIFC,KSL,KDGLU,KLED2,PAIA,PSIA,POA1,PW)
...
              ENDDO
-             DO I3=1,I1
-                DO I4=1,I2
-                   ZRPNMS(I3,I4) = S%FA(KMLOC)%RPNMS(I3,I4)
-                END DO
-             END DO
-             CALL GEMM('T','N',ILS,KIFC,KDGLU,1.0_JPRD,ZRPNMS,KDGLU,&
-                  &ZB_D,KDGLU,0._JPRD,ZCS_D,ILS)
+
+             CALL GEMM('T','N',ILS,KIFC,KDGLU,1.0_JPRD,S%RPNMS_DGEMM,KDGLU,&
```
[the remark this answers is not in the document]
samhatfield: Done.

src/trans/cpu/internal/ledir_mod.F90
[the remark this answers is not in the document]
samhatfield: Done.

src/trans/cpu/internal/ledir_mod.F90
```diff
@@ -268,4 +215,59 @@ SUBROUTINE LEDIR(KM,KMLOC,KFC,KIFC,KSL,KDGLU,PAIA,PSIA,POA1,PW)
...
+REAL(KIND=JPRB), INTENT(OUT) :: POUT(KDGLU, KIFC)
+
+INTEGER(KIND=JPIM) :: IFLD, JK, J
+
+IFLD = 0
+DO JK = 1, KFC, KSKIP
+  IFLD = IFLD + 1
+  DO J = 1, KDGLU
+    POUT(J, IFLD) = REAL(PIN(JK, KSL + J - 1) * PW(KSL + J - 1), JPRB)
+  ENDDO
```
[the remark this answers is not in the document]
samhatfield: You must be joking...

samhatfield:
I discovered a few slightly annoying quirks about develop: for some reason the packing of the inputs to the direct Legendre transform follows a different ordering of rounding for the antisymmetric and symmetric arrays.

Consider `ZB`. For the [antisymmetric case](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ledir_mod.F90#L117-L123), it is filled by `PAIA * PW`, where `PAIA` is `JPRB` and `PW` is `JPRD`. There is an implicit cast of `PAIA` to `JPRD`, a `JPRD` multiply, then the result is stored in `ZB` with a cast back to `JRPB`.

For the [symmetric case](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/cpu/internal/ledir_mod.F90#L191-L197), `PW` is first cast to `JPRB`, then the multiply is done at `JPRB` with no further casting as the result is stored in `ZB`.

For single-precision builds these two cases differ of course and so before commit [d7f5b14](https://github.com/ecmwf-ifs/ectrans/pull/409/commits/d7f5b1445872692f22f6d6d05943f0d0e3da308b) this branch did not give bit-identical norms compared with develop. I've now made this difference in logic as obvious as possible in the code and left a note that we should reconsider this in future. All checksums now match develop.

Looks like this difference was introduced by https://github.com/ecmwf-ifs/ectrans/commit/71a00c1707978e9ae17e7593fdd6f97fc43d6e95#diff-00f90301741845e1bcbd0eefb06115666eba9bb2f38d716e949af5d4deb2ba06 which I should have spotted when reviewing. Or was it deliberate @wdeconinck?

The really crazy one is the difference in behaviour for the IM=0 case. I have written out the chain of casts explicitly in [d7f5b14](https://github.com/ecmwf-ifs/ectrans/pull/409/commits/d7f5b1445872692f22f6d6d05943f0d0e3da308b):

```fortran
IF (CDSYMM_OR_ANTI == 'A') THEN
  POUT(J, IFLD) = REAL(REAL(REAL(PIN(JK, J), JPRD) * PW(KSL + J - 1), JPRB),JPRD)
ELSE
  POUT(J, IFLD) = REAL(PIN(JK, J) * REAL(PW(KSL + J - 1), JPRB), JPRD)
END IF
```
This should also be simplified in future because I don't think this special treatment has any scientific basis.

wdeconinck:
@samhatfield The change I did there was to force PW to be double precision, so it could be passed down in precision-independent setup_trans0. All the remaining changes were only to accommodate this.
It was not the intention that symmetric or antisymmetric has different behaviour.
I would suggest to make a series of PRs to tackle this, marked as non-bit-id (we have a label for this), and put it in the pipe-line for a next major release.

samhatfield:
Spotted a bug in `SUPOLF` which for reasons I don't fully understand is triggered by the changes in this PR. Essentially we were accessing elements in `DDPOL` which were uninitialised. This triggered a floating-point invalid exception in some large-scale runs. develop doesn't seem to suffer from the same problem.

[bdbe187](https://github.com/ecmwf-ifs/ectrans/pull/409/commits/bdbe187d99962911ddd90df6f6fbb27bdca6799c) fixes this, but my investigations also revealed a potentially related issue in `SULEG`, in which we are allocating `ZLPOL` outside of a threaded region and then accessing it within that region as a `PRIVATE` variable. According to the standard, allocatable arrays declared as `PRIVATE` are treated as unallocated. We are probably getting away with it because of compiler laziness. We should fix this in a separate PR.

samhatfield:
Hi @CarlosPenaDePedro, we are almost ready to merge the PR. I want to do some final benchmarking, but I'm not sure how to construct a benchmark that demonstrates the benefit of the new better-load balanced approach. Of course I could look at individual thread timings from Dr Hook from before. But ultimately the reason we are doing this is to reduce the overall wall time of the direct transform. Reducing the wall time of the thread handling `IM=0` by caching the DP polynomial arrays should reduce the overall wall time, and this should be measurable. So far I've tried running benchmarks on ECMWF's HPC with e.g. T1279 on 16 nodes, T2559 on 80 nodes and T5119 on 200 nodes, which is a very rough weak scaling set. However I can't really detect any notable differences in the median walltime of the direct transform. This is with `MKL_CBWR=AUTO`. In your experience, should this PR make a noticeable difference for this kind of benchmark, or are the benefits only in more niche situations?

CarlosPenaDePedro:
Great to hear. Thanks for all the work, including making it bit-identical.

When you say the median wall time of the direct transform, do you mean the median time reported by Dr. Hook, or the average time?

Looking at the median or average across all processes may not show any meaningful difference. For this reason, I was checking the maximum time across MPI processes reported by gstats. This is still relevant because, at some point in the IFS workflow, the processes must synchronise through an MPI collective or barrier. Therefore, even if only one process falls behind, it can affect the overall execution time.

In your setup, you can probably verify this by looking at the total ECTRANS time or the overall throughput. You should see a difference there, even if the median remains the same as in the previous version.

CarlosPenaDePedro:
Also note that the performance improvement was much more noticeable for larger decompositions, for instance on TCo2559 using 256 nodes. This is expected because the LEDIR computation decreases with the decomposition size, whereas the allocation time remains essentially constant. As a result, the allocation overhead becomes a larger fraction of the total runtime, making this fundamentally a scalability issue.

samhatfield:
Yes I see what you mean. It takes a bit of care to formulate the question properly. We have essentially three ways of profiling ecTrans: hand-coded timers in the benchmark program, GSTATS, and Dr Hook. There are barriers interspersed in various places and you need to know where these are in order to draw proper conclusions (unless you simply put a barrier at the beginning and end of every GSTATS or Dr Hook region, but would no longer be a "real world" benchmark).

Firstly we should look at GSTATS region 3. This records the overall time for an inverse->direct transform cycle, _after_ the calls to `SPECNORM`. The latter has an MPI barrier at the very end, so region 3 gives the synchronised wall time covering both the inverse and direct transform, so this will be the same for all ranks thanks to the barrier.

- For T1279 on 16 nodes this region is **306 ms** for develop, **312 ms** for this branch. **This branch is 3% slower.** (Unlikely to be statistically significant.)
- For T2559 on 80 nodes this region is **477 ms** for develop, **422 ms** for this branch. **This branch is 12% faster.**
- For T5119 on 200 nodes this region is **1267 ms** for develop, **1110 ms** for this branch. **This branch is 12% faster.**

I'm now also trying two more interesting setups: T1279 on 50 nodes ("operational like"), and T2559 on 250 nodes (Carlos' example).

CarlosPenaDePedro:
Unless I'm missing something, I agree that this approach would be the best way to evaluate the difference.

samhatfield:
Here are those two further benchmarks:
- T1279 on 50 nodes: **152 ms** (develop) vs. **147 ms** (this branch). **3% improvement.**
- T2559 on 250 nodes: **256 ms** (develop) vs. **208 ms** (this branch. **19% improvement.** Nice!!

samhatfield:
Okay, we've established that this is bit-ID with develop, has no impact on "operational-like" performance, and a clear positive impact on high-resolution, high-node-count runs. So, time to merge. Many thanks @CarlosPenaDePedro - in hindsight you discovered a very obvious scalability bug which we should have identified years ago! This should make a real impact on the performance of IFS at high resolutions. Good stuff.

samhatfield (approved):

CarlosPenaDePedro:
Thanks, @samhatfield ! And thank you for making the whole review and merge process so smooth.

PR #412, opened 2026-06-08
Field api interface ial integration fix

chesnelj:
### Description

Found some small issues using this product with field api interface for IAL.
optional arguments for inv_trans_field_api and dir_trans_field_api were not checked leading to crash.


samhatfield:
Looks reasonable to me. Could @dhaumont take a look?

dhaumont:
Yes, these fixes are necessary. Thank you Jeremy !
The get_layout code was introduced by Willem, so I will also let him check @wdeconinck :-)

src/field_api/dir_trans_field_api.F90
```diff
@@ -106,13 +106,13 @@ SUBROUTINE DIR_TRANS_FIELD_API(KRESOL,                &
...
-CALL GET_LAYOUT_S(YDFSPDIV, NSPEC2)
-CALL GET_LAYOUT_S(YDFSPSCALAR, NSPEC2)
-CALL GET_LAYOUT_S(YDFSPVOR, NSPEC2)
+IF (PRESENT(YDFU)) CALL GET_LAYOUT_G(YDFU, NBLK, NPROMA)
+IF (PRESENT(YDFV)) CALL GET_LAYOUT_G(YDFV, NBLK, NPROMA)
+IF (PRESENT(YDFSCALAR)) CALL GET_LAYOUT_G(YDFSCALAR,NBLK, NPROMA)
+IF (PRESENT(YDFSPVOR)) CALL GET_LAYOUT_S(YDFSPVOR, NSPEC2)
+IF (PRESENT(YDFSPDIV)) CALL GET_LAYOUT_S(YDFSPDIV, NSPEC2)
+IF (PRESENT(YDFSPSCALAR)) CALL GET_LAYOUT_S(YDFSPSCALAR, NSPEC2)
+IF (PRESENT(YDFSPVOR)) CALL GET_LAYOUT_S(YDFSPVOR, NSPEC2)
```
wdeconinck: Nice catch @chesnelj !

To be completely sure, in case none of these arguments are present, we need to assert that NSPEC2 > 0 .AND. NBLK > 0 .AND. NPROMA > 0.

(and also in inv_trans_field_api)

chesnelj: Ok I will manage this and propose a modification soon

chesnelj: I pushed the proposition of @wdeconinck, ty for the suggestion.
I intentionally kept the asserts separated for each variable to check.
This way user can know the bad one if any (even if I guess all of 3 are if no optional arguments are present ^^)

src/field_api/inv_trans_field_api.F90
```diff
@@ -145,6 +145,11 @@ SUBROUTINE INV_TRANS_FIELD_API(KRESOL,                                       &
 IF (PRESENT(YDFSPSCALAR)) CALL GET_LAYOUT_S(YDFSPSCALAR, NSPEC2)
 IF (PRESENT(YDFSPVOR)) CALL GET_LAYOUT_S(YDFSPVOR, NSPEC2)
 
+! Assert that NBLK, NPROMA and NSPEC2 are properly valued after getting the field(s) layout
+IF (.NOT.(NBLK > 0))    CALL ABOR1("[DIR_TRANS_FIELD_API] NBLK must be strictly positive")
+IF (.NOT.(NPROMA > 0))  CALL ABOR1("[DIR_TRANS_FIELD_API] NPROMA must be strictly positive")
+IF (.NOT.(NSPEC2 > 0))  CALL ABOR1("[DIR_TRANS_FIELD_API] NSPEC2 must be strictly positive")
```
wdeconinck: Thanks @chesnelj ! 
A copy-paste mistake: this should be INV_TRANS_FIELD_API.

In the abort, I think an extra hint that arguments are possibly missing is also good to add.

chesnelj: Typo fixed. Hint added. Should be good now

wdeconinck (approved):

PR #416, opened 2026-06-22
Make data device resident in `GPNORM_TRANS`

awnawab:
This PR makes data device resident throughout the execution of `GPNORM_TRANS`, including the addition of gpu aware comms.


src/trans/gpu/external/gpnorm_trans.F90
```diff
@@ -111,11 +122,24 @@ SUBROUTINE GPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL)
 INTEGER(KIND=JPIM) :: IPROC,ITAG,ILEN,ILENR,IBEG,IEND,IND
 TYPE(BUFFERED_ALLOCATOR) :: ALLOCATOR
 TYPE(TRGTOL_HANDLE) :: HTRGTOL
+LOGICAL :: LLGPU_ON_GPU = .FALSE.
```
samhatfield: Should this be called `LLPGP_ON_GPU`?

src/trans/gpu/external/gpnorm_trans.F90
```diff
@@ -196,22 +215,25 @@ SUBROUTINE GPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL)
 
 LGPNORM=.TRUE.
 CALL TRGTOL(ALLOCATOR,HTRGTOL,PREEL_REAL,IF_FS,IF_GP,0,IF_SCALARS_G,&
-  & KVSETSC=IVSET,PGP=PGP)
+  & KVSETSC=IVSET,PGP=PGP,LPGP_ON_GPU=LPGP_ON_GPU)
```
samhatfield:
```suggestion
  & KVSETSC=IVSET,PGP=PGP,LPGP_ON_GPU=LLGPU_ON_GPU)
```
But same comment as above applies to `LLGPU_ON_GPU`.

samhatfield:
I suggest we broaden the scope slightly and call it "tidy of GPNORM_TRANS" on GPU, and also delete gpnorm_trans_gpu.F90. The latter has never been functional and will be redundant once this is merged.

src/trans/gpu/external/gpnorm_trans.F90
samhatfield:
```suggestion
```
Might as well delete lines 207-211 while we're here.

samhatfield:
I added a test suite for `GPNORM_TRANS` which checks whether the average computed matches the (0, 0) mode of a corresponding spectral array. Currently it only tests `LPGP_ON_GPU=.FALSE.`.

samhatfield:
Failure on LUMI. Is this directive correct @awnawab?

gpnorm_trans.F90:224-228
```
#ifdef OMPGPU
!$OMP TARGET DATA MAP(PRESENT,ALLOC:ZAVE,ZMINGL,ZMAXGL,ZMINGPN,ZMAXGPN,ZAVEG,ZMING,ZMAXG) &
!$OMP& MAP(TO:IVSETS,IVSETG) &
!$OMP& MAP(PRESENT,ALLOC:PREEL_REAL,F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
#endif
```
Arrays `ZAVE` etc. -> those haven't yet been allocated on device right? Should they be `MAP(TO:...)`?

awnawab:
> Failure on LUMI. Is this directive correct @awnawab?
> 
> gpnorm_trans.F90:224-228
> 
> ```
> #ifdef OMPGPU
> !$OMP TARGET DATA MAP(PRESENT,ALLOC:ZAVE,ZMINGL,ZMAXGL,ZMINGPN,ZMAXGPN,ZAVEG,ZMING,ZMAXG) &
> !$OMP& MAP(TO:IVSETS,IVSETG) &
> !$OMP& MAP(PRESENT,ALLOC:PREEL_REAL,F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
> #endif
> ```
> 
> Arrays `ZAVE` etc. -> those haven't yet been allocated on device right? Should they be `MAP(TO:...)`?

I suspect it's the present map modifier. Looking back at it, the only function that serves is a safety check to abort if that symbol isn't already present, and of course it should throw an error for variables yet to be allocated on device. I think removing it is the right fix. Thanks a lot for adding the test! Otherwise this would have gone uncaught.

samhatfield:
We have now an error on LUMI only affecting mpi2 tests, which is not at all obvious. I'll have to reproduce it and interactively debug it on LUMI.

awnawab:
Considering the actual use case for this is in ifs coupled builds, and there's no pressing need to do that on lumi, we can exclude these 4 tests on lumi, leave an issue and then I'm happy to try to debug it in a non time pressing manner.

samhatfield:
Looks like you spotted a bug @awnawab ...? Tests passing on LUMI now.

awnawab:
Yeah, I'd misunderstood the difference between `use_device_ptr` and `use_device_addr`, all good now.

samhatfield:
Then I'll take a proper look at gpu/external/gpnorm_trans.F90 and give a final review.

src/trans/gpu/external/gpnorm_trans.F90
```diff
@@ -125,52 +146,47 @@ SUBROUTINE GPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL)
...
+  ENDIF
+  IF(IUBOUND(2) < KFIELDS) THEN
+    WRITE(NOUT,*)'GPNORM_TRANS:SEC. DIM. OF PGP TOO SMALL ',IUBOUND(2),KFIELDS
+    CALL ABORT_TRANS('GPNORM_TRANS:SECOND DIMENSION OF PGP TOO SMALL ')
+  ENDIF
+  IF(IUBOUND(3) < NGPBLKS) THEN
+    WRITE(NOUT,*)'GPNORM_TRANS:THIRD DIM. OF PGP TOO SMALL ',IUBOUND(3),NGPBLKS
+    CALL ABORT_TRANS('GPNORM_TRANS:THIRD DIMENSION OF PGP TOO SMALL ')
+  ENDIF
 ENDIF
```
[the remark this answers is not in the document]
samhatfield: Does this really matter @awnawab ?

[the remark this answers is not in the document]
awnawab: No I wouldn't think so. In the forecast timestep hotpath, running those consistency will incur an overhead, and it feels like overkill. However outside the hotpath, where we also call `GPNORM_TRANS`, the data in those cases is on host anyway, and there's little overhead for those checks.

samhatfield (approved):

PR #419, opened 2026-06-29
ectrans4py: add distributed-memory (MPI) Python interface

nwedi:
# ectrans4py: distributed-memory (MPI) Python interface + adjoint transforms

## What
Adds an MPI distributed-memory interface to `ectrans4py`, **additively** alongside
the existing serial LAM/Gaussian wrappers (no existing routine is removed or changed).
New names are suffixed to avoid clashes with the serial API.

### Commits

1. **Distributed-memory (MPI) forward transforms** — `inv_trans_scalar_dist4py`,
   `dir_trans_scalar_dist4py`, `inv_trans_uv_dist4py`, `dir_trans_uv_dist4py`
   (+ scalar-with-derivatives variant); SP single-precision build; de-duplicated
   serial/distributed interface.

2. **Adjoint transforms** — `inv_trans_scalar_dist4py_ad`, `dir_trans_scalar_dist4py_ad`,
   `inv_trans_uv_dist4py_ad`, `dir_trans_uv_dist4py_ad`. Four Fortran wrappers appended
   to `transform_dist4py.F90` + four Python bindings. IN/OUT directions reversed vs. forward.
   `ZSP`/`ZVOR`/`ZDIV` zeroed before `INV_TRANSAD` (PRFI1BAD appending convention,
   confirmed from `tests/trans/adjoint/test_invtrans_adjoint.F90`).

### Key fact (adjoint)

`adj(INV_TRANS) = Sᵀ ≠ Sᵀ W = DIR_TRANS` — they differ by the Gaussian weights W,
so dedicated adjoint routines are required. Identity validated to `20000·ε_float32`
(matching SCALPRODGP/SCALPRODSP from the ecTrans adjoint reference tests):

    <INV_TRANS x, y>_Euclid  =  <x, INV_TRANSAD y>_mfold
    <DIR_TRANS u, v>_mfold   =  <u, DIR_TRANSAD v>_Euclid

Measured max relative error ~1.4e-6 at T79/L137, 1 and 2 MPI tasks.

## Contributor Licence Agreement

By submitting this pull request I confirm that my contribution is made under the
terms of the Apache License 2.0.


src/ectrans4py/trans_inq_dist4py.F90
samhatfield, line 1: I think with some slight changes we could remove this subroutine. Firstly `PGW` can already be retrieved using `GET_LEGENDRE_ASSETS`. `PMU` can't at the moment but we could modify `GET_LEGENDRE_ASSETS` easily to also provide this.

Once this is done we can simply modify `TRANS_INQ4PY` to provide the other variables such as `KSPEC2` which aren't already provided.

samhatfield:
Regarding the new `SETUP_TRANS4PY`: at the moment we have a `SPEC_SETUP4PY` subroutine which doesn't have a Python interface. It's called automatically by `GP2SP_GAUSS4PY` etc., and it keeps track of initialisation state. It isn't parallel aware at all though. I wonder if it's possible to combine these two, or at least minimise repetition. Otherwise we will have two ways to initialise ecTrans.

src/ectrans4py/trans_inq_dist4py.F90
nwedi: Done — removed `trans_inq_dist4py` and folded it into `trans_inq4py`, which now takes a `KRESOL` argument: `KRESOL<=0` self-initialises serially from the grid parameters (unchanged behaviour), and `KRESOL>0` inquires the resolution already set up by `setup_trans_4py` (distributed). It now also returns `KSPEC2`, `KGPTOTG`, `KSPEC2G`, `KSMAX`, `PMU` and `PGW`. And `get_legendre_assets` now also returns `PMU`, as you suggested.

nwedi:
On the setup: I've factored the resolution-independent `SETUP_TRANS0` into a shared `SETUP_TRANS0_CORE` (new `setup_core4py`). Both the serial `SPEC_SETUP4PY` (`LDMPOFF=.TRUE.`) and the distributed `SETUP_TRANS0_4PY` (processor grid, `LDMPOFF=.FALSE.`) now call it, so the once-per-process init lives in one place.

I kept the two Python entry points (serial `SPEC_SETUP4PY` lazy self-init + resolution cache; distributed `setup_trans0_4py` → `setup_trans_4py`, mirroring `SUTRANS0`/`SUTRANS`) rather than a single one, because ecTrans fixes the parallelism at `SETUP_TRANS0` time — `LDMPOFF` and the processor grid `KPRGPNS×KPRGPEW×KPRTRW` — and `SETUP_TRANS0` is once-per-process, so a lazy self-initialising entry point has nowhere to receive the grid. Happy to collapse them further if you'd prefer a single parameterised entry point.

Two related changes while here:
- `trans_inq_dist4py` is gone; `trans_inq4py` now serves both via `KRESOL` (see the resolved thread), and `get_legendre_assets` returns `PMU`.
- The distributed transform/dist-gather/norm wrappers were passing float32 into the `ectrans4py_dp` (double) build and corrupting memory; they now pass float64 to match. Added a self-contained MPI round-trip test (`tests/test_ectrans4py/mpi_distributed.py`), green at 1 and 2 ranks (round trip machine-exact); the existing serial tests are updated for the new signatures.

nwedi:
Follow-up on the precision point: I've added a **single-precision build** so the distributed transforms can run in float32 (what forecast models want), keeping double as the default.

- `src/ectrans4py` now builds one library per enabled precision (`libectrans4py_sp` linking `trans_sp`/`etrans_sp`, `libectrans4py_dp` linking `trans_dp`/`etrans_dp`); the `ECTRANS4PY` feature condition is relaxed from `HAVE_ETRANS AND HAVE_DOUBLE_PRECISION` to `HAVE_ETRANS`.
- The Python layer loads `libectrans4py_sp`/`_dp` (selectable with `ECTRANS4PY_PRECISION={single,double}`, else double preferred) and sets the field-data dtype accordingly; the distributed transform/dist-gather/norm wrappers follow the build precision, the serial routines keep double (their dummy data is `REAL64`), and geometry stays double.
- Build with `-DENABLE_SINGLE_PRECISION=ON -DENABLE_DOUBLE_PRECISION=OFF` for the single-precision library.

Validated both on macOS: dp — serial suite 7/7 + distributed round trip machine-exact (9e-15); sp — distributed round trip at single-precision accuracy (1.2e-7), both at 1 and 2 ranks.

nwedi:
Pushed 2 more commits while building the pmap-ifs full-step TL/AD engine on top of this branch:

- **`ectrans4py: build single-precision by default`** -- the build was defaulting to double precision (`-DENABLE_SINGLE_PRECISION=OFF`), contradicting the adjacent comment's own stated guidance for distributed NWP use and breaking any downstream consumer (pmap-ifs) that hardcodes float32 across the interface.
- **`ectrans4py: add INV_TRANS_SCALAR_DERS_DIST4PY_AD`** -- exposes `INV_TRANSAD`'s existing `LDSCDERS` derivative-transform adjoint (no new transform maths, just a binding mirroring the existing `INV_TRANS_SCALAR_DIST4PY_AD`/`INV_TRANS_SCALAR_DERS_DIST4PY` pattern). Needed to linearise the horizontal-gradient half of the inverse transform. Validated natively (`tests/mpi_transform_adjoint.py` in pmap-ifs): relative error `4.4e-7` on TCo79 (tol `2.4e-3`).

samhatfield:
I've refactored the PR so it builds on two other PRs:
- https://github.com/ecmwf-ifs/ectrans/pull/424
- https://github.com/ecmwf-ifs/ectrans/pull/429
This has simplified it somewhat. Those PRs should be merged before this one.

samhatfield:
For reference here's a list of all the new ectrans4py routines introduced by this PR:
- `inv_trans_scalar_dist4py_ad`
- `inv_trans_scalar_ders_dist4py_ad`
- `dir_trans_scalar_dist4py_ad`
- `inv_trans_uv_dist4py_ad`
- `dir_trans_uv_dist4py_ad`
- `mpl_init4py`
- `mpl_end4py`
- `setup_trans0_4py`
- `setup_trans_4py`
- `dist_spec4py`
- `gath_spec4py`
- `dist_grid4py`
- `gath_grid4py`
- `inv_trans_scalar_dist4py`
- `inv_trans_scalar_ders_dist4py`
- `dir_trans_scalar_dist4py`
- `inv_trans_uv_dist4py`
- `dir_trans_uv_dist4py`
- `specnorm4py`
- `gpnorm_trans4py`

Existing routines:
- `ectrans_version`
- `get_legendre_assets`
- `etrans_inq4py`
- `trans_inq4py`
- `sp2gp_lam4py`
- `gp2sp_lam4py`
- `sp2gp_gauss4py`
- `gp2sp_gauss4py`
- `sp2gp_fft1d4py`

PR #420, opened 2026-06-29
Remove KLED2 parameter for LEDIR

samhatfield:
Since https://github.com/ecmwf-ifs/ectrans/commit/201c310a4259586ad3dbe2fbc874ed2b7ea378d9 this is not used.


wdeconinck (approved):
Nice tidy :)

PR #423, opened 2026-07-13
Simplify application of Gaussian weights

samhatfield:
As discovered in https://github.com/ecmwf-ifs/ectrans/pull/409, the application of the Gaussian weights (which are stored in double precision) is not consistent between the antisymmetric and symmetric code paths in the direct Legendre transform. This commit simplifies the logic by performing the multiplication in double precision in all cases (even though for `PACK_FOR_GEMM`, the output is cast to `JPRB`).

This change is NOT bit identical, but scientifically trivial.

@wdeconinck what's the best approach to merging this? Should we merge this immediately before making the next release?

Note that the packing routines will probably disappear in future when I refactor to remove `ZB`.


samhatfield:
We need to make sure we don't upset people with this merge so let's discuss offline before that happens.

wdeconinck (approved):
Looks good!

PR #425, opened 2026-07-14
Add action to check license header on new files

samhatfield:
This action checks that all new files in a PR contain the necessary license header. It checks Fortran, C/C++, Python, and CMake files.

This check does not look for a copyright notice. That's something we could try adding too.

This action was agentically written.


wdeconinck:
This looks like something that could be more centralised, e.g. in https://github.com/ecmwf-ifs/prtk @msleigh ?

samhatfield:
Yes I like the new bot for checking copyright statements. Could that be extended to check the license header too?

PR #427, opened 2026-07-15
Update required ecBuild version to 3.14.2

samhatfield:
In order for the compile flag mechanism introduced in #397 and fixed by https://github.com/ecmwf-ifs/ectrans/pull/407 to work properly, ecBuild >= 3.14.2 is required. This requirement should have been added in #407 but I overlooked it. So here it is.


wdeconinck (approved):
Thanks! Indeed an oversight.

PR #429, opened 2026-07-16
Allow single-precision ectrans4py installation

samhatfield:
ectrans4py is hard-coded to only work with the double-precision backend of ecTrans. This PR modifies it to also support single-precision.

At the moment, single-precision must be chosen when the pip package is installed by passing flags `-C cmake.define.ENABLE_SINGLE_PRECISION=ON -C cmake.define.ENABLE_DOUBLE_PRECISION=OFF`. It is only possible currently to build a pip package supporting one precision at a time. A further PR may allow one to build a single pip package which can accept inputs of either precision but I'm not sure how to do that at this stage.


samhatfield:
https://github.com/ecmwf-ifs/ectrans/pull/424 should be merged first.

wdeconinck (approved):
Looks good!

PR #433, opened 2026-07-22
Make F%RPNM a local array of SULEG

samhatfield:
This array does not need to be a module variable. It is only used locally within `SULEG`. Having it in `TPM_FIELDS` is confusing because it suggests that this is where the persistent Legendre polynomials are kept, when in fact they are kept in `TPM_FLT` (where they are split into symmetric and antisymmetric components).

Also a bit of tidying.


samhatfield:
Failure in etrans... why the heck does etrans even care about Legendre polynomials??!

wdeconinck:
> Failure in etrans... why the heck does etrans even care about Legendre polynomials??!

It is possible it is just posing as Legendre polynomials to be similar in API as "trans" but in fact is in fact another array related to bifourier transforms. @ddegrauwe you know?

ddegrauwe:
It seems RPNM is only used in etrans_inq, so I guess it's a remnant of blind translation of the global (trans_inq) to the lam (etrans_inq) case. suleg is also not called from esetup_trans.
I think it's ok to just remove it from etrans_inq.

wdeconinck (approved):
Thanks for confirming @ddegrauwe.
Nice find @samhatfield ! Many thanks!

samhatfield:
We could just remove the argument... depends how strictly you want to protect the API. I can almost guarantee that nobody is passing that argument to `ETRANS_INQ`...

wdeconinck:
I guess so yes.

PR #438, opened 2026-07-24
Device resident GPNORM_TRANS fix

MichaelSt98:
## Summary

`GPNORM_TRANS` (GPU): unguarded `PRESENT(PREEL_REAL, …)` aborts on ranks with `IF_FS == 0`

### Description

On MPI decompositions where the V-set parallelism `NPRTRV` exceeds
the number of fields `KFIELDS` passed to `GPNORM_TRANS`, some MPI ranks
end up with a local field count `IF_FS == 0`. On those ranks
`PREEL_REAL` is associated but has zero length, and the OpenACC /
OpenMP `PRESENT` clause on the outer `!$ACC DATA` / `!$OMP TARGET
DATA` region introduced by PR #416 (`c7f348d8`, "Make data device
resident in `GPNORM_TRANS`") aborts with the NVHPC runtime fatal:

```
FATAL ERROR: data in PRESENT clause was not found on device 1:
name=preel_real(:) host:0x...
```

All uses of `PREEL_REAL` inside `GPNORM_TRANS` are already guarded by
`IF (IF_FS > 0) THEN … ENDIF`, and the analogous DATA region inside
`TRGTOL` carries `IF (KF_FS > 0)` on its `PRESENT(PREEL_REAL)`. The
outer DATA region in `GPNORM_TRANS` is missing that same guard.

The fix moves `PREEL_REAL` out of the outer DATA region's PRESENT list
and places it in a nested DATA region wrapping the existing
`IF (IF_FS > 0) THEN … ENDIF` block, with `IF (IF_FS > 0)` on the
directive itself. Behaviour when `IF_FS > 0` is unchanged.

### Symptom

Large-resolution IFS-RAPS runs with NVHPC GPU builds abort during
model setup with:

```
FATAL ERROR: data in PRESENT clause was not found on device 1:
name=preel_real(:) host:0x401dfc3d1180
```

DrHook traceback ends at:

```
… CSTA → SUINIF → SUGRIDF → GPNORM3,#2 → GPNORM1,#2 → GPNORM_TRANS,#2 → TRGTOL,#2
```

i.e. the second surface-field norm called from `SUGRIDF`. The fatal is
raised on the `!$ACC DATA … PRESENT(PREEL_REAL, …)` at line ~225 of
`src/trans/gpu/external/gpnorm_trans.F90` (post PR #416), immediately
after `TRGTOL` returns.

The failure is only observed for "large" runs. It is deterministic on
those runs and always fires on ranks with `MYSETV > KFIELDS`.

### Root cause

Since PR #416 (`c7f348d8`), `GPNORM_TRANS` opens an outer DATA region
right after the call to `TRGTOL`:

```fortran
#ifdef OMPGPU
!$OMP TARGET DATA MAP(ALLOC:ZAVE,ZMINGL,ZMAXGL,ZMINGPN,ZMAXGPN,ZAVEG,ZMING,ZMAXG) &
!$OMP& MAP(TO:IVSETS,IVSETG) MAP(ALLOC:PREEL_REAL,F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
#endif
#ifdef ACCGPU
!$ACC DATA CREATE(ZAVE,ZMINGL,ZMAXGL,ZMINGPN,ZMAXGPN,ZAVEG,ZMING,ZMAXG) &
!$ACC& COPYIN(IVSETS,IVSETG) PRESENT(PREEL_REAL,F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
#endif

CALL GSTATS(1429,0)
IF (IF_FS > 0) THEN
  … all PREEL_REAL uses live here …
ENDIF
```

`F_RW`, `D_NSTAGTF`, `D_NPTRLS`, `G_NLOEN` are all mapped to the
device by `SETUP_TRANS` and are always present. `PREEL_REAL`, however,
is set up on the fly inside `TRGTOL`:

```fortran
! trgtol_mod.F90 (excerpt)
CALL ASSIGN_PTR(PREEL_REAL, GET_ALLOCATION(ALLOCATOR, HTRGTOL%HCOMBUFR_AND_REEL), &
                1_JPIB*KF_FS*D%NLENGTF*C_SIZEOF(PREEL_REAL(1))+1,                 &
                1_JPIB*KF_FS*D%NLENGTF*C_SIZEOF(PREEL_REAL(1)))
!$ACC DATA COPYIN(IRECV_BUFR_TO_OUT) PRESENT(PREEL_REAL) IF (KF_FS > 0) ASYNC(1)
```

Note the `IF (KF_FS > 0)` on `TRGTOL`'s own PRESENT check. When
`KF_FS == 0`, `ASSIGN_PTR` is called with `LENGTH_IN_BYTES == 0` and
`PREEL_REAL` is a zero-length pointer into the growing allocator's
buffer. The `TRGTOL` DATA region is not entered in that case, so no
PRESENT check runs.

The outer DATA region in `GPNORM_TRANS`, however, has no such guard.
When `IF_FS == 0` on entry to it, NVHPC evaluates
`PRESENT(PREEL_REAL)` against a zero-length descriptor whose host
base falls just past the mapped range of `GROWING_ALLOCATION%PTR`,
finds no match, and aborts.

#### When does `IF_FS == 0` happen?

`GPNORM_TRANS` computes:

```fortran
IF_FS=0
DO J=1,KFIELDS
  IVSET(J)=MOD(J-1,NPRTRV)+1
  IF (IVSET(J) == MYSETV) IF_FS = IF_FS + 1
ENDDO
```

so on ranks where `MYSETV > KFIELDS` and `NPRTRV > KFIELDS`, no field
is assigned and `IF_FS == 0`. This is exactly the situation for
`GPNORM_TRANS` called out of `SUGRIDF` (via `GPNORM3` → `GPNORM1`) for
surface field norms, where `KFIELDS = number of soil / snow levels`
(typically 2–8). On small runs `NPRTRV ≤ KFIELDS` so every V-set gets
at least one field and the bug is masked; on large runs where
`NPRTRV` exceeds the small `KFIELDS` used for these norms, ranks with
`MYSETV > KFIELDS` hit the fatal.

### Fix

Remove `PREEL_REAL` from the outer DATA region's PRESENT list and
wrap the existing `IF (IF_FS > 0) THEN … ENDIF` block in a nested
DATA region that carries the same guard:

```fortran
CALL GSTATS(1429,0)
IF (IF_FS > 0) THEN

! PREEL_REAL is only meaningfully device-mapped when IF_FS > 0.
#ifdef OMPGPU
  !$OMP TARGET DATA MAP(PRESENT,ALLOC:PREEL_REAL) IF (IF_FS > 0)
#endif
#ifdef ACCGPU
  !$ACC DATA PRESENT(PREEL_REAL) IF (IF_FS > 0)
#endif

  … existing kernels that reference PREEL_REAL …

#ifdef OMPGPU
  !$OMP END TARGET DATA
#endif
#ifdef ACCGPU
  !$ACC END DATA
#endif

ENDIF ! IF_FS > 0
```

Correspondingly, the outer directive loses `PREEL_REAL` from its
list:

```fortran
!$OMP TARGET DATA MAP(ALLOC:ZAVE,ZMINGL,ZMAXGL,ZMINGPN,ZMAXGPN,ZAVEG,ZMING,ZMAXG) &
!$OMP& MAP(TO:IVSETS,IVSETG) MAP(ALLOC:F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
…
!$ACC DATA CREATE(ZAVE,ZMINGL,ZMAXGL,ZMINGPN,ZMAXGPN,ZAVEG,ZMING,ZMAXG) &
!$ACC& COPYIN(IVSETS,IVSETG) PRESENT(F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
```

This is the minimal change and mirrors the guard already present in
`TRGTOL` for its own `PREEL_REAL` DATA region.

#### Why this is safe

- When `IF_FS > 0`: the nested DATA region is entered, `PREEL_REAL`
  is verified present (via `GROWING_ALLOCATION%PTR`'s device mapping)
  exactly as before. All kernels using `PREEL_REAL` run inside the
  nested region, so their PRESENT semantics are unchanged.
- When `IF_FS == 0`: the nested DATA region's `IF (IF_FS > 0)` clause
  causes it to be a no-op; no PRESENT check is performed. The
  containing `IF (IF_FS > 0) THEN … ENDIF` block already skips all
  kernels that would touch `PREEL_REAL`, so this matches the
  intended behaviour.
- `F_RW`, `D_NSTAGTF`, `D_NPTRLS`, `G_NLOEN` remain in the outer
  PRESENT list; they are populated once by `SETUP_TRANS` and are
  always present regardless of `IF_FS`.
- The OpenMP-offload branch is patched identically for consistency.

#### Verification

With the fix applied and the diagnostic left in place, the previously
aborting runs progress past `SUGRIDF`.

### Contributor Declaration


src/trans/gpu/external/gpnorm_trans.F90
```diff
@@ -219,16 +219,24 @@ SUBROUTINE GPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL,LPGP
...
 
 CALL GSTATS(1429,0)
 IF (IF_FS > 0) THEN
 
+! PREEL_REAL is only meaningfully device-mapped when IF_FS > 0
+#ifdef OMPGPU
+  !$OMP TARGET DATA MAP(PRESENT,ALLOC:PREEL_REAL) IF (IF_FS > 0)
+#endif
+#ifdef ACCGPU
+  !$ACC DATA PRESENT(PREEL_REAL) IF (IF_FS > 0)
```
marsdeno, line 237: Out of curiosity, is the conditional on the DATA statement even necessary, since we are actually inside a  non-directive IF (IF_FS>) at line 230?
Not that it matters much!

marsdeno (approved):

src/trans/gpu/external/gpnorm_trans.F90
MichaelSt98: No (probably) not necessary, but better safe than sorry.

src/trans/gpu/external/gpnorm_trans.F90
```diff
@@ -219,16 +219,24 @@ SUBROUTINE GPNORM_TRANS(PGP,KFIELDS,KPROMA,PAVE,PMIN,PMAX,LDAVE_ONLY,KRESOL,LPGP
...
-!$ACC& COPYIN(IVSETS,IVSETG) PRESENT(PREEL_REAL,F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
+!$ACC& COPYIN(IVSETS,IVSETG) PRESENT(F_RW,D_NSTAGTF,D_NPTRLS,G_NLOEN)
 #endif
 
 CALL GSTATS(1429,0)
 IF (IF_FS > 0) THEN
 
+! PREEL_REAL is only meaningfully device-mapped when IF_FS > 0
+#ifdef OMPGPU
+  !$OMP TARGET DATA MAP(PRESENT,ALLOC:PREEL_REAL) IF (IF_FS > 0)
```
awnawab: The `PRESENT` attribute here is just a mapper modifier, and here it serves effectively as an assert, which will throw an error if `PREEL_REAL` isn't already in the map. Since we are allocating it here for the first time, we should just remove it. The lumi tests are failing, and this could be the reason, although the logs don't show any information beyond a failed state.

awnawab (changes requested):
Thanks a lot @MichaelSt98 for mopping this up 🙏 I've left a small correction for the new openmp target clause, once that's fixed and the lumi test is green, we can merge this.

awnawab (approved):
Thanks @MichaelSt98 for the quick turnaround, this is now G2G ✅

wdeconinck (approved):
Very nice. Merging.

PR #439, opened 2026-07-29
Remove GPU version of PREPSNM

samhatfield:
Through progressive refactoring this subroutine gradually lost its usefulness. Now it is only called once, and all it does is copy from `F%REPSNM` to `FG%ZEPSNM`. This is a simple operation which doesn't justify a whole separate subroutine.

I also removed some unused symbols in `SETUP_TRANS`.


wdeconinck (approved):
Nice tidying. Similar approach for CPU path is possible?

samhatfield:
> Nice tidying. Similar approach for CPU path is possible?

The CPU subroutine also seems very minimal, but it is at least well used, so there isn't as strong a case to remove it, compared with the GPU version.

PR #441, opened 2026-07-29
Add a C pointer interface to ectrans

dhaumont:
### Description

This PR adds a C pointer interface to ecTrans, allowing in the future to interface ecTrans with different technology stacks: Python, atlas, field API interface, etc.

It's a continuation of https://github.com/ecmwf-ifs/ectrans/pull/314, which introduced a field API interface to ectrans. The new C pointert interface takes make the connectionto the legacy PGP/PSP interface, instead of field API interface. Field API interface becomes a thin layer, calling the C pointer interface.

This was implemented by Willem Deconink during the ectrans scientific stay in Reading, April 2026.

### Contributor Declaration


dhaumont:
@wdeconinck The merge with develop was not flawless. I'm fixing the compilation issues generated

wdeconinck:
@dhaumont see my last commit with a workaround for a wrong gfortran optimisation bug.

dhaumont:
> @dhaumont see my last commit with a workaround for a wrong gfortran optimisation bug.

 Indeed I also saw the bug happening when running the unit tests. Well done Willem !
What is the status now regarding the unit tests ?

wdeconinck:
> > @dhaumont see my last commit with a workaround for a wrong gfortran optimisation bug.
> 
> Indeed I also saw the bug happening when running the unit tests. Well done Willem ! What is the status now regarding the unit tests ?

Actually this only fixed it locally for gfortran-16; gfortran-14 is still failing the same way....

Debugging this still with Claude at the moment... and finding another workaround.

Root cause: create_fields_lists built the field lists as

    ylf%spvor = [make_field_spec(ywflds%spvor,'spvor',kvsetuv)]

passing the b-set (ivset) as an allocatable component of a function result
through an array constructor. gfortran miscompiles this at -O2: the
component's data is lost across the assignment, so ylf%spvor(1)%ivset was
already garbage before use. The value was then handed to inv_trans as
KVSETUV and tripped its bounds check. Any read of the value masked the bug,
which is why it only appeared in optimised builds.

dhaumont:
Willem, I already had to fix this bug before with the field API interface.

In the initial implementation, the arrays were allocated and returned by functions. And  I had the same crashes in optimisation mode.
 The bug fix was to replace all the functions by subroutines,  preallocate the arrays and pass them as arguments to the subroutines.

dhaumont:
> Willem, I already had to fix this bug before with the field API interface.
> 
> In the initial implementation, the arrays were allocated and returned by functions. And I had the same crashes in optimisation mode. The bug fix was to replace all the functions by subroutines, preallocate the arrays and pass them as arguments to the subroutines.

See here: https://github.com/dhaumont/ectrans/commit/3434f46777935e2c82d2c1a64aa01a54d1f304ce

src/field_api/inv_trans_field_api.F90
```diff
@@ -78,452 +75,135 @@ SUBROUTINE INV_TRANS_FIELD_API(KRESOL,                                       &
...
+ALLOCATE(YLFVGVOR(NFIELD_VORGP))
+ALLOCATE(YLFVGDIV(NFIELD_DIVGP))
+ALLOCATE(YLFVGSCALAR(NFIELD_SCALAR))
+ALLOCATE(YLFVGU_EW(NFIELD_UVDER))
+ALLOCATE(YLFVGV_EW(NFIELD_UVDER))
+ALLOCATE(YLFVGSCALAR_NS(NFIELD_SCDER))
+ALLOCATE(YLFVGSCALAR_EW(NFIELD_SCDER))
+
+DO IFIELD=1,NFIELD_UV
+  CALL MAKE_FIELD_VIEW(YLFVGU(IFIELD),   YDFU(IFIELD),     FIELD_TYPE, .FALSE., .TRUE.)
```
dhaumont: I hope the MAKE_FIELD_VIEW calls are not too costly. We have a lot of pointer conversions, from array to field_view to C, and then back from pointer to field_view, and finally a copy from field_view to array. We should probably try to measure the overhead of the pointer conversions (the last copy to an array will disappear so it's less of a concern)

src/trans/include/fspgl_intf.h
```diff
@@ -0,0 +1,32 @@
+! (C) Copyright 2001- ECMWF.
```
dhaumont: The file fspgl_intf.h is duplicated:

- /source/ectrans/src/trans/include/ectrans/fspgl_intf.h
- /source/ectrans/src/trans/include/fspgl_intf.h

Is it the intent ?

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -380,4 +380,37 @@ subroutine synchost_rdonly_wrapped_fields(ywflds)
 
 end subroutine synchost_rdonly_wrapped_fields
 
+
+subroutine nullify_wrapped_fields(ywflds)
+
+    ! Nullify all pointers in ywflds
```
dhaumont: indent

dhaumont:
@wdeconinck  
The CI/CD unit tests are green,  but when I run them manually  they failed in compare_checksums.

Do the CI/CD execute compare_checksums ? I don't find it in the logs of the github action:
https://github.com/ecmwf-ifs/ectrans/actions/runs/30618690988/job/91120948797

It's strange, since `-E compare_checksums` is only defined for LUMI:
https://github.com/ecmwf-ifs/ectrans/blob/0d41adefe7d7d9249009f5f0f730ede675d43f3f/.github/workflows/build-hpc.yml#L111

Here are the results of my manual testing
```
$ cd build/ectrans/tests
$ ctest

The following tests FAILED:
	  1 - ectrans_test_install (Failed)
	  2 - compare_checksums (Failed)
Errors while running CTest
```
And the log of compare_checksums is :
```
80% comparisons passed, 63 comparisons failed out of 318
The following checksum files do not match their mpi0_omp1 references:
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode1_nfld10_nlev20_npromatr20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld0_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode1_nfld10_nlev20_npromatr20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld0_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode1_nfld10_nlev20_derivatives_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_field_api_nfld10_nlev20_flt_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_field_api_nfld10_nlev20_flt_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld0_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld0_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_vordiv_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_vordiv_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_nproma16_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_nproma16_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_flt_inv_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode2_nfld10_nlev20_flt_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp8_callmode2_nfld10_nlev20_flt_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp1_callmode2_nfld10_nlev20_flt_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi1_omp8_callmode2_nfld10_nlev20_flt_dir_trans.checksums
./ectrans-benchmark-cpu-dp_T47_O48_mpi4_omp1_callmode2_nfld10_nlev20_flt_dir_trans.checksums

```

dhaumont:
UPDATE: compare_checksums: is executed successfuly by the CI/CD here:
https://github.com/ecmwf-ifs/ectrans/actions/runs/30620730891/job/91124448271?pr=441

wdeconinck:
The comparison of checksums can be quite sensitive to the used BLAS/LAPACK libraries even though results will be correct to some round-off errors. To be sure, use the reference BLAS/LAPACK (from Netlib). Other libraries may decide to use different algorithms to solve the GEMMs depending on matrix dimensions. Dimensions may differ sometimes with MPI values and NPROMATR values.

I have been working on a more robust checksum branch in parallel here: #442 which in my opinion is ready to merge, and then we can merge this together. This still requires a reproducible BLAS/LAPACK though.
And in the end the goal is #434 to compare with stored reference files for each CI runner.
I have locally merged this branch with #434 one to compare references and they all seem to match well now.

dhaumont:
> The comparison of checksums can be quite sensitive to the used BLAS/LAPACK libraries even though results will be correct to some round-off errors. To be sure, use the reference BLAS/LAPACK (from Netlib). Other libraries may decide to use different algorithms to solve the GEMMs depending on matrix dimensions. Dimensions may differ sometimes with MPI values and NPROMATR values.

Here I only run without MPI, and it's clearly the comparison field API/no field API which fails. At some point this was working, and there is no reason the code in this PR would have broken this. I will re-test in develop to see if field API/no field API is still bit-reproducible

> I have been working on a more robust checksum branch in parallel here: #442 

Thanks for pointing me to this branch. I will review it

CMakeLists.txt
```diff
@@ -119,6 +119,9 @@ if( (NOT HAVE_CPU) AND (NOT HAVE_GPU) )
 endif()
 
 if( HAVE_GPU )
+  if( HAVE_MPI)
+     ecbuild_find_package( NAME MPI REQUIRED COMPONENTS Fortran )
+  endif()
```
samhatfield: Looks like a merge bug? I think this can be removed.

dhaumont:
> Here I only run without MPI, and it's clearly the comparison field API/no field API which fails. At some point this was working, and there is no reason the code in this PR would have broken this. I will re-test in develop to see if field API/no field API is still bit-reproducible

The develop branch has the same behaviour as this PR (ie field API/ no field API are also not bit reproducible) with my test setup. Given the fact that this PR passes the CI/CD tests, I'm reassured that the PR doesn't break bit reproducibility

src/field_api/dir_trans_field_api.F90
```diff
@@ -39,305 +39,77 @@ SUBROUTINE DIR_TRANS_FIELD_API(KRESOL,                &
...
-    ENDDO
+ALLOCATE(YLFVGU(NFIELD_UV))
+ALLOCATE(YLFVGV(NFIELD_UV))
+ALLOCATE(YLFVGSCALAR(NFIELD_SCALAR))
+ALLOCATE(YLFVSVOR(NFIELD_UV))
+ALLOCATE(YLFVSDIV(NFIELD_UV))
+ALLOCATE(YLFVSSCALAR(NFIELD_SCALAR))
+
+DO IFIELD=1,NFIELD_UV
+  CALL MAKE_FIELD_VIEW(YLFVGU(IFIELD),   YDFU(IFIELD),     FIELD_TYPE, LDACC=.FALSE., LDRDONLY=.TRUE.)
```
dhaumont, line 90: @wdeconinck LDRDONLY attributes are not consistent.
It's equal to .TRUE .for UV fields and to .FALSE. for scalar fields.
But the distinction should be on GRID and SPECTRAL instead.

src/field_api/inv_trans_field_api.F90
```diff
@@ -78,452 +74,135 @@ SUBROUTINE INV_TRANS_FIELD_API(KRESOL,                                       &
...
+ALLOCATE(YLFVGVOR(NFIELD_VORGP))
+ALLOCATE(YLFVGDIV(NFIELD_DIVGP))
+ALLOCATE(YLFVGSCALAR(NFIELD_SCALAR))
+ALLOCATE(YLFVGU_EW(NFIELD_UVDER))
+ALLOCATE(YLFVGV_EW(NFIELD_UVDER))
+ALLOCATE(YLFVGSCALAR_NS(NFIELD_SCDER))
+ALLOCATE(YLFVGSCALAR_EW(NFIELD_SCDER))
+
+DO IFIELD=1,NFIELD_UV
+  CALL MAKE_FIELD_VIEW(YLFVGU(IFIELD),   YDFU(IFIELD),     FIELD_TYPE, LDACC=.FALSE., LDRDONLY=.TRUE.)
```
dhaumont: Same as in dir_trans_field_api: LDRDONLY attributes are not consistent.
The value should be set in with regards to the fact that it's a GRID or a SPECTRAL field

src/field_api/dir_trans_field_api.F90
wdeconinck: You are right, good spotting

CMakeLists.txt
wdeconinck: Removed.

src/field_api/dir_trans_field_api.F90
wdeconinck: I have fixed it here as well as in inv_trans_field_api. I was not careful.

src/programs/util/ectrans_field_api_helper.F90
wdeconinck: OK. 👍
I was wondering why we needed the `nullify_wrapped_fields` as they should be nullified upon construction of the type(wrapped_fields). Was this due to a compiler bug @dhaumont @samhatfield ? We did not have it before this PR and it was working OK.

dhaumont: > I was wondering why we needed the nullify_wrapped_fields as they should be nullified upon construction of > the type(wrapped_fields). Was this due to a compiler bug @dhaumont @samhatfield ? We did not have it > > before this PR and it was working OK.

This is not intentional. I implemented `nullify_wrapped_fields` in a previous version of ectrans with Field API, because I thought it was required (I didn't know that the pointers would be nullified upon construction). You probably removed it at some point, but I put it back during the merge conflict resolution.

If you are 100% sure it's fine without it, we can remove it.

wdeconinck: Good to know, I will remove it. Following pattern sets them to `null` upon construction:

```text
type wrapped_fields
  ...
  class(field_3rb), pointer :: spscalar3 => null()  ! spectral scalar fields
  class(field_2rb), pointer :: spscalar2 => null()  ! spectral surfacic scalar fields
  class(field_2rb), pointer :: spscalar  => null()  ! spectral scalar fields as a single field
  ...
  ```

File: src/trans/field_view/ectrans_field_view_internal_util_mod.F90 [hunk 7]
```diff
@@ -0,0 +1,581 @@
...
+! (C) Copyright 2001- Meteo-France.
+!
+! This software is licensed under the terms of the Apache Licence Version 2.0
+! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+! In applying this licence, ECMWF does not waive the privileges and immunities
+! granted to it by virtue of its status as an intergovernmental organisation
+! nor does it submit to any jurisdiction.
+
+
+MODULE ECTRANS_FIELD_VIEW_INTERNAL_UTIL_MOD
```
dhaumont, line 11: @wdeconinck This module should be made accessible from the routines `trans/cpu/internal`, in order to propagate lists of `GRID_VIEW` and `SPEC_VIEW` down the call tree.

Does it mean that the file src/trans/field_view/ectrans_field_view_internal_util_mod.F90 has to be moved somewhere else ?

wdeconinck: Yes it could be that the SPEC_VIEW and GRID_VIEW types need to go to deeper, if those are the types that are being used to pass down all the way to TRGTOL and TRLTOG. For the moment the idea was to make it clearly not in the field_api directory, and mark them clearly as internal so we can move them around as needed.

dhaumont: I've started a prototype to propagate down SPEC_VIEW and GRID_VIEW. I will show you when it's in better shape

wdeconinck: Thanks Denis, much appreciated!

src/programs/util/ectrans_field_api_helper.F90
wdeconinck: Removed.

src/trans/include/fspgl_intf.h
wdeconinck: The duplication has been fixed by @samhatfield now.

src/field_api/inv_trans_field_api.F90
wdeconinck: At least at this level the cost should be negligible because all levels and scalars are still combined into these single multidimensional views here, matching the field_api fields.

Further extraction of the 2d GRID_VIEW and SPEC_VIEW will be slightly more expensive because levels and individual variables will each become a view, but I believe still negligible compared to the cost of the entire transforms algorithm following. Still good to measure to remove doubt, but I would start there.
Currently there is no benefit in measuring this because the overhead is dominated quite substantially with all the copies to/from the "classic" dir_trans inv_trans input/output fields, until we move the 2d views down to TRGTOL/TRLTOG.

dhaumont: I agree with you, it might not be a problem if we keep converting only unfolded data. I was wondering if   the conversion from C pointer to Fortran was fast, or if it was as costly as a memory allocation or a system call. But as it's just some arithmetic manipulation and assignments, it should not be the case.

src/trans/field_view/include/ectrans/dir_trans_field_view.h
```diff
@@ -0,0 +1,47 @@
...
+!       YDSPSCALAR(:)  - List of spectral scalar fields
+!       YDSPVOR(:)     - List of spectral vector fields (vorticity)
+!       YDSPDIV(:)     - List of spectral vector fields (divergence)
+
+USE ISO_FORTRAN_ENV, ONLY : INT32
+USE ECTRANS_FIELD_VIEW_MOD, ONLY : FIELD_VIEW
+
+IMPLICIT NONE
+
+INTEGER(KIND=INT32), INTENT(IN)    :: KRESOL
```
samhatfield: I think @wdeconinck wrote this originally - any reason you chose `INT32` instead of e.g. `JPIM`?

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -0,0 +1,301 @@
...
+#include "abor1.intfb.h"
+
+IF (LHOOK) CALL DR_HOOK('DIR_TRANS_FIELD_VIEW',0,ZHOOK_HANDLE)
+
+NFIELD_UV           = SIZE(YDGPU)
+NFIELD_SCALAR       = SIZE(YDGPSCALAR) 
+NPROMA              = GET_NPROMA(YDGPU, YDGPV, YDGPSCALAR)
+NBLK                = GET_NBLK(YDGPU, YDGPV, YDGPSCALAR)
+NFIELD_TOTAL_UV     = GET_NFLD(YDGPU)
+NFIELD_TOTAL_SCALAR = GET_NFLD(YDGPSCALAR)
```
samhatfield:
```suggestion
```
Looks like these aren't needed anymore. Suggest you delete these and also `GET_NFLD`.

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -0,0 +1,301 @@
...
+  ALLOCATE(ZPGPUV(NPROMA,KFLEVG, IUVG * IUVDIM,NBLK))
+
+  ! For LG we need the ivset of each grid point field,
+  ! so we extract a matching list from the spectral fields.
+  ALLOCATE(IVSETUV_LIST(IUVG))
+  DO JFLD=1,IUVG
+    CALL FIELD_VIEW_GET_IVSET_PTR(YDSPVOR(JFLD), IVSETUV_LIST(JFLD)%PTR)
+  ENDDO
+
+  C = LG(YDGPU, YLGVU, IVSETUV_LIST)
```
samhatfield: I remember this variable `C` was used before `LG_COUNT` was introduced - is that right? `C` is not used anymore so you can just delete this. Also `LG` should probably be a subroutine.

src/trans/field_view/dir_trans_field_view.F90
dhaumont: I have  done a lot of cleaning here: https://github.com/dhaumont/ectrans/pull/11    

We should not spend too much time cleaning dir_trans_field_view.F90 and inv_trans_field_view.F90 here

src/trans/field_view/dir_trans_field_view.F90
dhaumont: `C` is also still present in https://github.com/dhaumont/ectrans/pull/11, and it should be better cleaned there.

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -0,0 +1,301 @@
...
+REAL(KIND=JPRB),POINTER :: ZPSPVOR(:,:),ZPSPDIV(:,:)  ! spectral vector fields (out)
+REAL(KIND=JPRB),POINTER :: ZPSPSC2(:,:)               ! spectral scalar fields(out)
+REAL(KIND=JPRB),POINTER :: ZPGPUV(:,:,:,:)            ! grid vector fields (in)
+REAL(KIND=JPRB),POINTER :: ZPGP2(:,:,:)               ! grid scalar fields (in)
+
+REAL(KIND=JPRB), POINTER :: ZZ1_1(:)
+REAL(KIND=JPRB), POINTER :: ZZ1_2(:)
+REAL(KIND=JPRB), POINTER :: ZZ2_1(:,:)
+REAL(KIND=JPRB), POINTER :: ZZ2_2(:,:)
+REAL(KIND=JPRB):: S
```
samhatfield:
```suggestion
```
I think this is also dead leftover code.

src/trans/field_view/include/ectrans/dir_trans_field_view.h
wdeconinck: This is to become a more PARKIND independent API hopefully, with C-interoperable types.

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -0,0 +1,301 @@
...
+IF (IUVG>0) THEN
+  C = LS(YDSPVOR, YLSPVVOR)
+  C = LS(YDSPDIV, YLSPVDIV)
+
+  DO JFLD=1,IFLDSPVOR
+    IF (ASSOCIATED(YLSPVVOR(JFLD)%P)) THEN
+      ZZ1_1=>YLSPVVOR(JFLD)%P
+      ZZ1_2=>YLSPVDIV(JFLD)%P
+      ZZ1_1(:) = ZPSPVOR(JFLD,:)
+      ZZ1_2(:) = ZPSPDIV(JFLD,:)
```
samhatfield, line 258:
```suggestion
      YLSPVVOR(JFLD)%P = ZPSPVOR(JFLD,:)
      YLSPVDIV(JFLD)%P = ZPSPDIV(JFLD,:)
```
Any reason you can't do something like this?

src/trans/field_view/dir_trans_field_view.F90
wdeconinck: I agree with @dhaumont, this whole story is very much to be made in multiple PRs, with more chance to revisit the cleanup there, also avoiding too many merge conflicts. But thanks @samhatfield for flagging this and I hope that the branch from @dhaumont will address this.

src/trans/field_view/dir_trans_field_view.F90
dhaumont: This has been cleaned already in  https://github.com/dhaumont/ectrans/pull/11.

src/field_api/dir_trans_field_api.F90
```diff
@@ -39,305 +39,77 @@ SUBROUTINE DIR_TRANS_FIELD_API(KRESOL,                &
...
-	             & PSPSC2 = ZPSPSC2,PGP2 = ZPGP2, KVSETSC2 = IVSETSC2, &
-	             & KPROMA = NPROMA, KRESOL = KRESOL)
-ELSE IF (ASSOCIATED(ZPGP2)) THEN
-	CALL DIR_TRANS(PSPSC2 = ZPSPSC2,PGP2 = ZPGP2, KVSETSC2 = IVSETSC2, &
-	             & KPROMA = NPROMA, KRESOL = KRESOL)
-ELSE IF (ASSOCIATED(ZPGPUV)) THEN
-	CALL DIR_TRANS(PSPVOR = ZPSPVOR,PSPDIV = ZPSPDIV,PGPUV = ZPGPUV,KVSETUV = IVSETUV, &
-	             & KPROMA = NPROMA, KRESOL = KRESOL)
+IF (PRESENT(YDFSCALAR)) THEN
+  NFIELD_SCALAR = SIZE(YDFSCALAR)
```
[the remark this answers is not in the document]
dhaumont: I believe this is fixed in in https://github.com/dhaumont/ectrans/pull/11, but this needs to be double checked.

src/field_api/inv_trans_field_api.F90
```diff
@@ -78,452 +74,135 @@ SUBROUTINE INV_TRANS_FIELD_API(KRESOL,                                       &
...
-        CALL INV_TRANS(PSPVOR=ZPSPVOR,PSPDIV=ZPSPDIV,PGPUV=ZPGPUV,KVSETUV=IVSETUV, &
-                     & LDSCDERS=LLSCDERS, LDVORGP=LLVORGP, LDDIVGP=LLDIVGP, LDUVDER=LLUVDER,  &
-                     & KPROMA=NPROMA, FSPGL_PROC=FSPGL_PROC, KRESOL=KRESOL)
-    ELSE
-        CALL INV_TRANS(PSPVOR=ZPSPVOR,PSPDIV=ZPSPDIV,PGPUV=ZPGPUV,KVSETUV=IVSETUV, &
-                     & LDSCDERS=LLSCDERS, LDVORGP=LLVORGP, LDDIVGP=LLDIVGP, LDUVDER=LLUVDER,  &
-                     & KPROMA=NPROMA, KRESOL=KRESOL)
-    ENDIF
+IF (PRESENT(YDFSCALAR)) THEN
+  NFIELD_SCALAR = SIZE(YDFSCALAR)
```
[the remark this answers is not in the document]
dhaumont: I believe this is fixed in in https://github.com/dhaumont/ectrans/pull/11, but this needs to be double checked.

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -0,0 +1,511 @@
...
+    LLDIVGP = .TRUE.
+    IUVDIM = 5
+    ALLOCATE(YLGVDIV(LG_COUNT(YDGPDIV)))
+  ENDIF
+
+  ! Output vorticity of vector fields
+  IF (SIZE(YDGPVOR) > 0) THEN
+    LLVORGP = .TRUE.
+    IUVDIM = 6
+    ALLOCATE(YLGVVOR(LG_COUNT(YDGPVOR)))
```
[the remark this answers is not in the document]
dhaumont: I believe this is fixed in in https://github.com/dhaumont/ectrans/pull/11, but this needs to be double checked.

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -0,0 +1,511 @@
...
+    LLUVDER = .TRUE.
+    IUVDIM = 5
+    ALLOCATE(YLGVU_EW(LG_COUNT(YDGPU_EW)))
+    ALLOCATE(YLGVV_EW(LG_COUNT(YDGPV_EW)))
+  ENDIF
+
+  ! Output divergence of vector fields
+  IF (SIZE(YDGPDIV)  > 0) THEN
+    LLDIVGP = .TRUE.
+    IUVDIM = 5
```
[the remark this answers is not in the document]
dhaumont: I believe this is fixed in in https://github.com/dhaumont/ectrans/pull/11
But this needs to be double checked.

src/field_api/inv_trans_field_api.F90
```diff
@@ -13,7 +13,6 @@ SUBROUTINE INV_TRANS_FIELD_API(KRESOL,                                       &
                              & YDFSCALAR, YDFU, YDFV,                        &
                              & YDFVOR,YDFDIV,                                &
                              & YDFSCALAR_NS, YDFSCALAR_EW, YDFU_EW, YDFV_EW, &
-                             & KGPTOT,                                       &
                              & FSPGL_PROC)
```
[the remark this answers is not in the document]
dhaumont: Fixed in in https://github.com/dhaumont/ectrans/pull/11

src/trans/field_view/dir_trans_field_view.F90
dhaumont: @samhatfield As this PR #441 is included in https://github.com/dhaumont/ectrans/pull/11, maybe it makes more sense to review that one for the dir_trans_field_view and inv_trans_field_view, instead of this one, which is an intermediate version ?

src/field_api/inv_trans_field_api.F90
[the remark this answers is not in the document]
wdeconinck: I don't think https://github.com/dhaumont/ectrans/pull/11 fixes this, but also I don't think it needs to.
This change was intentional, and this API was not ( or should not have been ) set in stone yet as it was not in a tagged release. I think this is a warranted change.

src/trans/field_view/dir_trans_field_view.F90
samhatfield: I definitely don't want to review code that will be deleted anyway. Is there any reason not to simply merge `propagate_ptr` into `ectrans_field_view_denis` before any reviewing? Reviewing PRs outside of ecmwf-ifs/ectrans is not ideal because we don't have access to all features, e.g. CI testing on external machines.

src/field_api/inv_trans_field_api.F90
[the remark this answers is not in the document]
dhaumont: You are right.

I had to fix a similar problem, and changed the order of arguments in src/trans/cpu/internal/trgtol_mod.F90, to have the optional `KPTRGP` in the last position.
It is a look alike problem, but it has nothing to do with this, my bad.

src/trans/field_view/dir_trans_field_view.F90
dhaumont: The main reson for keeping them separated was the separation of concerns, but we can reconsider this. 
If we want to integrate `propagate_ptr` into `ectrans_field_view_denis`, I think we should first agree on several points:
- the new naming conventions (ie duplicating routines and adding the `_view` suffix)
- the change of location for `ectrans_field_view_mod.F90`

@wdeconinck What do you think ?

src/trans/field_view/inv_trans_field_view.F90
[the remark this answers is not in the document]
wdeconinck: It would be good to have this fixed here in this PR as well.

src/trans/field_view/ectrans_field_view_mod.F90
```diff
@@ -0,0 +1,232 @@
...
+SUBROUTINE GET_DATA_PTR_2RM(YLFV, PTR)
+  IMPLICIT NONE
+  TYPE(FIELD_VIEW), INTENT(IN) :: YLFV
+  REAL(C_FLOAT), POINTER, INTENT(INOUT) :: PTR(:,:)
+  INTEGER(C_INT), PARAMETER :: RANK = 2
+  INTEGER(C_INT) :: EXTENTS(RANK), JDIM
+  CALL ASSERT_EXPECTED_RANK_AND_KIND("GET_DATA_PTR_2RM", YLFV, RANK, C_FLOAT)
+  DO JDIM=1,RANK
+    EXTENTS(JDIM) = YLFV%STRIDES(JDIM+1)/YLFV%STRIDES(JDIM)
+  ENDDO
```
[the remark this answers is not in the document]
wdeconinck: Needs to be fixed

src/field_api/ectrans_field_api_mod.F90
```diff
@@ -13,100 +13,83 @@ MODULE ECTRANS_FIELD_API_MOD
...
 END TYPE
 
 INTERFACE MAKE_FIELD_GRID
   MODULE PROCEDURE :: MAKE_FIELD_GRID_1RB, MAKE_FIELD_GRID_2RB, MAKE_FIELD_GRID_3RB, MAKE_FIELD_GRID_4RB
 END INTERFACE
 
 INTERFACE MAKE_FIELD_SPEC
-  MODULE PROCEDURE :: MAKE_FIELD_SPEC_1RB, MAKE_FIELD_SPEC_2RB, MAKE_FIELD_SPEC_3RB
+  MODULE PROCEDURE :: MAKE_FIELD_SPEC_2RB, MAKE_FIELD_SPEC_3RB
 END INTERFACE
```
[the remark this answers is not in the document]
wdeconinck: @dhaumont do we need support for 1RB spectral fields?

src/trans/field_view/dir_trans_field_view.F90
wdeconinck: Indeed it is the separation which is the reason. At this point we want to have a situation that introduces the field_view API, but keeping the situation mostly status quo compared to the field_api with respect to the actual implementation. i.e. we are not yet going deeper. This makes the field_api layer now very thin.
In the end. I don't consider https://github.com/dhaumont/ectrans/pull/11 ready for review at this point, until this is merged.
I do think that we can address most issues here if they are minor code changes. copilot showed a number of potential bugs and I think those should be fixed already here.
Also I'd prefer to have the PR #434 merged before this PR, and rebase this, so we can be confident of these changes.

dhaumont: I completely agree with you, Willem. Let's stick to the initial plan: first #434, then #441, and later on https://github.com/dhaumont/ectrans/pull/11 when it will be mature enough.
I can take care of the small code cleanings mentioned by Sam, in this PR and port them to https://github.com/dhaumont/ectrans/pull/11 as well.
However, I will skip the changes that have been superseeded in https://github.com/dhaumont/ectrans/pull/11 
Regarding the bugs mentionned by copilot, I prefer to them to you.
Is this okay with you ?

src/field_api/ectrans_field_api_mod.F90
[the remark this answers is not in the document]
dhaumont: no, but for me it doesn't hurt to keep it

src/trans/field_view/dir_trans_field_view.F90
dhaumont: fixed in 302b676470aaeaff9b98dccfbd6a08cdf10ceebe

src/trans/field_view/dir_trans_field_view.F90
dhaumont: Fixed in 302b676470aaeaff9b98dccfbd6a08cdf10ceebe

src/trans/field_view/dir_trans_field_view.F90
dhaumont: `NFIELD_TOTAL_SCALAR` removed in 302b676470aaeaff9b98dccfbd6a08cdf10ceebe

src/trans/field_view/ectrans_field_view_internal_util_mod.F90
dhaumont: For the record, the prototype is here https://github.com/dhaumont/ectrans/pull/11

src/trans/field_view/dir_trans_field_view.F90
dhaumont: I suggest to leave this suggestion behind as the code is gone in https://github.com/dhaumont/ectrans/pull/11

src/field_api/dir_trans_field_api.F90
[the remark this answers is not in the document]
dhaumont: I suggest to leave this as it is

src/field_api/inv_trans_field_api.F90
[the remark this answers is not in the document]
dhaumont: I suggest to leave this as it is

wdeconinck:
#434 has landed in develop and I merged it into this branch, which now proves bit reproducibility. Even though there are some issues reported by copilot; locally on my laptop field_api tests are reproducible.
I will continue working on the copilot reported issues now.

dhaumont:
@wdeconinck @samhatfield I would like to validate `dhaumont:propagate_ptr`  as well.
I have already merged develop into it. What's the proper way to execute the tests to prove bit-reproducibility ?

src/trans/field_view/inv_trans_field_view.F90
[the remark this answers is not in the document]
wdeconinck: Done now

src/trans/field_view/ectrans_field_view_mod.F90
```diff
@@ -0,0 +1,232 @@
...
+  EXTENTS(RANK) = YLFV%EXTENTS(RANK)
+  CALL C_F_POINTER(YLFV%DATA, PTR, EXTENTS)
+  PTR => PTR(1:YLFV%EXTENTS(1), 1:YLFV%EXTENTS(2))
```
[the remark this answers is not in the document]
wdeconinck: Fixed.

src/trans/field_view/ectrans_field_view_mod.F90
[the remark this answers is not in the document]
wdeconinck: Fixed.

src/field_api/ectrans_field_api_mod.F90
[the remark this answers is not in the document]
wdeconinck: Added.

src/trans/field_view/inv_trans_field_view.F90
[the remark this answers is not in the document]
wdeconinck: Done now.

src/trans/field_view/ectrans_field_view_mod.F90
```diff
@@ -139,7 +139,7 @@ SUBROUTINE GET_DATA_PTR_2RM(YLFV, PTR)
   INTEGER(C_INT), PARAMETER :: RANK = 2
   INTEGER(C_INT) :: EXTENTS(RANK), JDIM
   CALL ASSERT_EXPECTED_RANK_AND_KIND("GET_DATA_PTR_2RM", YLFV, RANK, C_FLOAT)
-  DO JDIM=1,RANK
+  DO JDIM=1,RANK-1
```
dhaumont, line 534: Indeed, can we explain why  YLFV%STRIDES(JDIM+1) was not crashing before, when JDIM = RANK ?

wdeconinck:
> What's the proper way to execute the tests to prove bit-reproducibility ?

@dhaumont  if following guidance in the README with the `ECTRANS_TEST_REFERENCE_DIR` options doesn't seem obvious, please let me know and we can adapt the README:

https://github.com/ecmwf-ifs/ectrans/#testing-ectrans

src/trans/field_view/ectrans_field_view_mod.F90
wdeconinck: The arrays `YLFV%STRIDES` and `YLFV%EXTENTS` are sized with `MAX_RANK=4`
After this DO loop the value `YLFV%EXTENTS(RANK)` gets a correct value, so no problem then.
Perhaps we were not exercising the `GET_DATA_PTR_4RM` function which would go out of bounds in the DO loop.

Fair to say that some unit-tests are still missing to really cover everything. Rather than in the ectrans-benchmark we should get some dedicated field_api and/or field_view tests.

dhaumont: > The arrays `YLFV%STRIDES` and `YLFV%EXTENTS` are sized with `MAX_RANK=4` After this DO loop the value `YLFV%EXTENTS(RANK)` gets a correct value, so no problem then. Perhaps we were not exercising the `GET_DATA_PTR_4RM` function which would go out of bounds in the DO loop.
> 
> Fair to say that some unit-tests are still missing to really cover everything. Rather than in the ectrans-benchmark we should get some dedicated field_api and/or field_view tests.

Thank you for the nice explanation. Indeed, it could only crash for `GET_DATA_PTR_4RM`

dhaumont:
> > What's the proper way to execute the tests to prove bit-reproducibility ?
> 
> @dhaumont if following guidance in the README with the `ECTRANS_TEST_REFERENCE_DIR` options doesn't seem obvious, please let me know and we can adapt the README:
> 
> https://github.com/ecmwf-ifs/ectrans/#testing-ectrans

For the time beeing I will rely on the lazy approach, letting the github action validates this for me...

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -0,0 +1,520 @@
...
+
+!     ------------------------------------------------------------------
+IF (LHOOK) CALL DR_HOOK('INV_TRANS_FIELD_VIEW',0,ZHOOK_HANDLE)
+
+
+NFIELD_UV           = SIZE(YDGPU)
+NFIELD_SCALAR       = SIZE(YDGPSCALAR)
+NPROMA              = GET_NPROMA(YDGPU, YDGPV, YDGPSCALAR)
+NBLK                = GET_NBLK(YDGPU, YDGPV, YDGPSCALAR)
+NFIELD_TOTAL_UV     = GET_NFLD(YDGPU)
```
samhatfield, line 143: `NFIELD_TOTAL_UV` isn't used, nor in `DIR_TRANS_FIELD_VIEW`. That also means `GET_NFLD` isn't used. Is this leftover from a previous version or needed for future use in https://github.com/dhaumont/ectrans/pull/11? Same thing for `NFIELD_UV` and `NFIELD_SCALAR`.

samhatfield: Oh I see they're deleted in https://github.com/dhaumont/ectrans/pull/11. Sorry I wasn't sure what the verdict is for this PR. How much code review do I need to do?

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -0,0 +1,520 @@
...
+ENDIF
+IF (SIZE(YDGPU_EW) > 0 .OR. SIZE(YDGPV_EW) > 0) THEN
+  IF (SIZE(YDGPU_EW) /= SIZE(YDGPV_EW) .OR. SIZE(YDGPU_EW) /= SIZE(YDGPU)) THEN
+    CALL ABORT_TRANS("[INV_TRANS_FIELD_VIEW] YDGPU_EW, YDGPV_EW and YDGPU must have the same size")
+  ENDIF
+ENDIF
+IF ((SIZE(YDSPSCALAR)/= SIZE(YDGPSCALAR))) THEN
+  CALL ABORT_TRANS("[INV_TRANS_FIELD_VIEW] Inconsistent size for YDSPSCALAR and YDGPSCALAR")
+ENDIF
+
```
samhatfield, line 188: Should we have some consistency checks also on `YDGPSCALAR_NS` and `YDGPSCALAR_EW`?

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -0,0 +1,520 @@
...
+
+
+! 1. Vector fields transformation to grid space
+
+! Do we have vector fields?
+IF (SIZE(YDGPU) > 0) THEN
+
+  ! Convert list of spectral vector fields into a list of 2d FIELD_VIEW
```
samhatfield, line 196:
```suggestion
  ! Convert list of spectral vector fields into a list of 2d SPEC_VIEW
```

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -0,0 +1,520 @@
...
+
+  IFLDSPVOR = LS_COUNT(YDSPVOR)
+  ALLOCATE(YLSPVVOR(IFLDSPVOR))
+  ALLOCATE(YLSPVDIV(IFLDSPVOR))
+  ! Convert list of grid-point vector fields into a list of 2d FIELD_VIEW
```
samhatfield, line 201:
```suggestion
  ! Convert list of grid-point vector fields into a list of 2d GRID_VIEW
```

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -0,0 +1,298 @@
...
+! Preliminary checks
+
+! Do we have vector fields?
+IF (SIZE(YDGPU) > 0) THEN
+
+  IF ((SIZE(YDGPU) /= SIZE(YDGPV)).OR.(SIZE(YDGPU) /= SIZE(YDSPDIV)).OR.(SIZE(YDGPU)/= SIZE(YDSPVOR))) THEN
+    CALL ABORT_TRANS("[DIR_TRANS_FIELD_VIEW] The vector arrays have inconsistent sizes: YDGPU, YDGPV, YDSPDIV, YDSPVOR")
+  ENDIF
+
+  ! Convert list of spectral vector fields into a list of 2d FIELD_VIEW
```
samhatfield:
```suggestion
  ! Convert list of spectral vector fields into a list of 2d GRID_VIEW
```

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -0,0 +1,298 @@
...
+  IFLDSPVOR = LS_COUNT(YDSPVOR)
+
+  ALLOCATE(YLSPVVOR(IFLDSPVOR))
+  ALLOCATE(YLSPVDIV(IFLDSPVOR))
+
+  ! Convert list of grid-point vector fields into a list of 2d FIELD_VIEW
```
samhatfield:
```suggestion
  ! Convert list of grid-point vector fields into a list of 2d SPEC_VIEW
```

src/programs/util/ectrans_field_api_helper.F90
```diff
@@ -234,7 +234,7 @@ subroutine create_fields_lists(ywflds,ylf, kvsetuv, kvsetsc,kvsetsc2)
   ! Create field lists in ylf from field API objects in ywflds
 
   type(wrapped_fields), intent(in) :: ywflds       !input fields api objects
-  type(fields_lists), intent(inout) :: ylf ! output field lists
+  type(fields_lists), intent(inout), target :: ylf ! output field lists
```
samhatfield, line 237: Very minor comment, but why do we need `target` here? It's also used in a few places in `ECTRANS_FIELD_VIEW_INTERNAL_UTIL_MOD` which I didn't understand.

wdeconinck: `TARGET` is required when pointers are being created, which we do with calls like `C_LOC` when creating `FIELD_VIEW`

src/trans/field_view/dir_trans_field_view.F90
wdeconinck: Done

src/trans/field_view/dir_trans_field_view.F90
wdeconinck: Done

wdeconinck:
Hi @dhaumont 
I have further refined the ectrans_field_view_mod with some helper functions inside moved from the ectrans_field_api_mod.
Changes: 
  - unused FIELD_TYPE removed
  - int ACCESS :  contains bits describing accessibility: READONLY , DEVICE, HOST (For unified memory both DEVICE and HOST must be set). Currently the access is not verified but that should be done, perhaps in your propagate_ptr follow-up PR.
  - int INDEXING: contains bits describing the role of each dimension. This variable can be used to distinguish in the future if we have to distinguish e.g. fields where dimensions are swapped. (NPROMA,NFIELD,NLEV,NBLK) instead of (NPROMA,NLEV,NFIELD,NBLK)
  
I added a dir_trans_field_view testsuite: 
It follows exactly the existing dir_trans testsuite with all callmodes.
 
There is some duplication now between the dir_trans and dir_trans_field_view test suite that we could clean up in the future with a util library.
We could also merge the two test suites in one which will bloat a bit but also allow for testing bit-identiality.

dhaumont:
> * unused FIELD_TYPE removed

Indeed, it was supposed to be used if we had a single list of fields propagated, some sorting beeing done in the routine using the FIELD_TYPE. But in the end it's cleaner to keep propagating a set of list already filtered by type (YDGVU, YDGVV, etc)

> * int ACCESS :  contains bits describing accessibility: READONLY , DEVICE, HOST (For unified memory both DEVICE and HOST must be set). Currently the access is not verified but that should be done, perhaps in your propagate_ptr follow-up PR.
> * int INDEXING: contains bits describing the role of each dimension. This variable can be used to distinguish in the future if we have to distinguish e.g. fields where dimensions are swapped. (NPROMA,NFIELD,NLEV,NBLK) instead of (NPROMA,NLEV,NFIELD,NBLK)

> I added a dir_trans_field_view testsuite: It follows exactly the existing dir_trans testsuite with all callmodes.

Great !

> There is some duplication now between the dir_trans and dir_trans_field_view test suite that we could clean up in the future with a util library. We could also merge the two test suites in one which will bloat a bit but also allow for testing bit-identiality.

Yes, we can also do that in a future PR

src/trans/field_view/inv_trans_field_view.F90
dhaumont: The verdict was: the changes in the code that will survive https://github.com/dhaumont/ectrans/pull/11 are made here, the changes in the code that will be deleted in 11 are  discarded, but the one that are  are very small and localized.
For the code review, you should not worry to much about ectrans_field_view and dir_trans_field_view, that will change a lot in propagate_ptr, but you can focus on all the rest that will remain the same.

src/field_api/inv_trans_field_api.F90
```diff
@@ -78,452 +74,186 @@ SUBROUTINE INV_TRANS_FIELD_API(KRESOL,                                       &
...
- ENDIF
-
-  ! Output divergence of vector fields
-  IF (PRESENT(YDFDIV)) THEN
-    LLDIVGP = .TRUE.
-    IUVDIM = IUVDIM + 1
-    ALLOCATE(YLGVDIV(LG_COUNT(YDFDIV)))
+IF (PRESENT(YDFVOR)) THEN
+  IF (.NOT. PRESENT(YDFSPVOR)) THEN
+    CALL ABOR1('INV_TRANS_FIELD_API: YDFSPVOR BE PRESENT IF YDFVOR IS PRESENT')
```
samhatfield, line 119:
```suggestion
    CALL ABOR1('INV_TRANS_FIELD_API: YDFSPVOR MUST BE PRESENT IF YDFVOR IS PRESENT')
```

src/field_api/ectrans_field_api_mod.F90
```diff
@@ -154,451 +160,186 @@ FUNCTION MAKE_FIELD_GRID_4RB(YLF,NAME)
...
-
-SUBROUTINE LG4RB(LG4RBV, YLF,NAME,IVSET,LDACC,LDRDONLY)
-  ! Creation of a list of GRID_VIEW encapsulating the layers of an input FIELD_4RB field
-  ! Given YLF dimensioned(NPROMA, NLEVS, NFIELDS, NBLKS) as input,
-  ! the output list size will contain  (NLEVS*NFIELDS) elementa, their view beeing dimensioned(NPROMA, NBLKS)
-
-  CLASS (FIELD_4RB), INTENT(IN), POINTER :: YLF
-  CHARACTER(LEN=*), INTENT(IN) :: NAME
-  INTEGER(KIND=JPIM), POINTER, INTENT(IN) :: IVSET(:)
+SUBROUTINE SET_DATA_1RB(YLFV, FIELD, LDACC, LDRDONLY)
```
samhatfield, line 243: Could `SET_DATA_1RB`, `SET_DATA_2RB` etc. be combined into a generic subroutine operating on `TYPE(FIELD_BASIC)` (parent of `FIELD_1RB` etc.) with `PTR` declared as assumed size (`(*)`)?

samhatfield: Seems that the answer is "no" because `FIELD_BASIC` is a very lightweight base class with none of the getters.

dhaumont:
@wdeconinck I'm strugging to retrieve `SPEC_VIEW` from the device in an efficient way.

Here are different pseudo-code versions of `prfi1b_mod.F90`, where  SPEC data are copied to the PIA array. I have tested  different variants to access `SPEC_VIEW%P` from the device, but none of them is satisfying.
Any help regarding this topic is welcome, I believe some GPU implementation should also be tested in `dir_trans_field_view_test_suite`.


----
1. Initial reference version,  based on PSPEC (legacy interface, no SPEC_VIEW used here)
```
REAL(KIND=JPRB)   ,INTENT(IN)   :: PSPEC(:,:)
!$ACC PARALLEL LOOP COLLAPSE(2) PRESENT(PSPEC)
DO JFLD=1,IFIELDS  
  DO JN=0,R_NSMAX+3
    PIA(2*JFLD-1,JN+1) = PSPEC(JFLD,INM)
  END DO
ENDDO
!$ACC END DATA
```
- Description: data stored in PSPEC which is on device, the loop on GPU is a as fast as possible array-to-array copy on GPU.

- Observation: the reference version, fast.

2. Direct adaptation of  the reference version for GPU.
```
TYPE(SPEC_VIEW), INTENT(IN) :: YDSP(:)
!$ACC DATA COPYIN(YDSP) 
!$ACC PARALLEL LOOP COLLAPSE(2) PRESENT(YDSP)
DO JFLD=1,IFIELDS  
   DO JN=0,R_NSMAX+3
        PIA(2*JFLD-1,JN+1) = YDSP(JFLD)%P(INM)
   END DO
ENDDO
!$ACC END DATA
```

- Description: Copy YDSP to the GPU beforehand, than access the data with YDSP(JFLD)%P
- Observation:  works, but 5x slower than the reference version. Not sure why it's slow, maybe the memory access are not coalescent anymore, or  the COPYIN(YDSP) requires a deep copy or an `attach` clause

3. Assign ZZ pointer on CPU and use a PRESENT clause for the GPU cause : FAILURE
```
TYPE(SPEC_VIEW), INTENT(IN) :: YDSP(:)
REAL(KIND=JPRB), POINTER :: ZZ(:)

DO JFLD=1,IFIELDS  
   ZZ=>YDSP(JFLD)%P
   !$ACC PARALLEL LOOP  PRESENT(ZZ, PIA)
   DO JN=0,R_NSMAX+3
       PIA(2*JFLD-1,JN+1,KMLOC) = ZZ(INM)
   END DO
ENDDO
```

- Description: a variant where we loop over the fields on the CPU.
- Observation:  the present clause will fail, ZZ is not present. Suprising to me, as this pattern is used in IAL, whenever we use GET_DEVICE_PTR to retieve the GPU adress of a field API object.

4. assign ZZ pointer on CPU and use a DEVICEPTR clause for the GPU: works, but slow.
```
TYPE(SPEC_VIEW), INTENT(IN) :: YDSP(:)
REAL(KIND=JPRB), POINTER :: ZZ(:)

DO JFLD=1,IFIELDS  
   ZZ=>YDSP(JFLD)%P
   !$ACC PARALLEL LOOP  PRESENT(ZZ) DEVICEPTR(ZZ)
   DO JN=0,R_NSMAX+3
       PIA(2*JFLD-1,JN+1,KMLOC) = ZZ(INM)
   END DO
ENDDO
```
 
- Description: the only change with the previous version is that we use the DEVICEPTR instead of the PRESENT clause for ZZ
- Observation: the code works, so ZZ is a genuine device memory address ? In that case, where is the code converting the GET_DEVICE_PTR from field API (which retrieves a CPU adress) to a GPU pointer ?
Unfortunately, the code is still very slow, 10x slower than the reference version.

samhatfield:
It seems that we're all satisfied with this PR - is there anything else to add? If not, I'm happy to approve.

dhaumont:
> It seems that we're all satisfied with this PR - is there anything else to add? If not, I'm happy to approve.

Indeed, I would be happy to have this merged, to move forward with the other PR's coming on top of this one.
I  still think we need more GPU code in the unit tests, but this can come in some future PR's.

wdeconinck:
One thing which I was considering now, which could be a duplication in metadata.
We have the `INDEXING` integer which has the information which dimension is 'LEVELS'.
If we have the convention that IVSET is on levels, then the IVSET_DIM integer is not needed.
If you (@dhaumont @samhatfield) agree I can remove IVSET_DIM, and rely on INDEXING for this information.

dhaumont:
> One thing which I was considering now, which could be a duplication in metadata. We have the `INDEXING` integer which has the information which dimension is 'LEVELS'. If we have the convention that IVSET is on levels, then the IVSET_DIM integer is not needed. If you (@dhaumont @samhatfield) agree I can remove IVSET_DIM, and rely on INDEXING for this information.

No problem for me.

samhatfield:
> One thing which I was considering now, which could be a duplication in metadata.
> We have the `INDEXING` integer which has the information which dimension is 'LEVELS'.
> If we have the convention that IVSET is on levels, then the IVSET_DIM integer is not needed.
> If you (@dhaumont @samhatfield) agree I can remove IVSET_DIM, and rely on INDEXING for this information.

Sounds good.

wdeconinck:
> @wdeconinck I'm strugging to retrieve `SPEC_VIEW` from the device in an efficient way.

I would have thought option 3 would be the one to use. That would be so if field_api would "acc-map" the host-pointer to the device pointer.

There are probably ACC environment variables to debug (I think NVCOMPILER_ACC_NOTIFY=3) and see what's going on, as well as nvhpc profilers

dhaumont:
> > @wdeconinck I'm strugging to retrieve `SPEC_VIEW` from the device in an efficient way.
> 
> I would have thought option 3 would be the one to use. That would be so if field_api would "acc-map" the host-pointer to the device pointer.
> 
> There are probably ACC environment variables to debug (I think NVCOMPILER_ACC_NOTIFY=3) and see what's going on, as well as nvhpc profilers

We are on the same page. I'm working out a small reproducer to better understand what is going on

wdeconinck:
We can anyway merge this PR @dhaumont ?

dhaumont:
> We can anyway merge this PR @dhaumont ?

yes, please go ahead !

dhaumont:
> > > @wdeconinck I'm strugging to retrieve `SPEC_VIEW` from the device in an efficient way.
> > 
> > 
> > I would have thought option 3 would be the one to use. That would be so if field_api would "acc-map" the host-pointer to the device pointer.
> > There are probably ACC environment variables to debug (I think NVCOMPILER_ACC_NOTIFY=3) and see what's going on, as well as nvhpc profilers
> 
> We are on the same page. I'm working out a small reproducer to better understand what is going on

Reproducer of the problem here: https://github.com/destination-earth-digital-twins/DE330_GPU/blob/main/support/ATOS/copy_list_of_ptrs/


[copy_list_of_ptrs.tar.gz](https://github.com/user-attachments/files/31948522/copy_list_of_ptrs.tar.gz)

dhaumont:
> [copy_list_of_ptrs.tar.gz](https://github.com/user-attachments/files/31948522/copy_list_of_ptrs.tar.gz)

@wdeconinck My understanding is that the data layout of YDSP and YDGP is not compatible with the current GPU implementation. I hope we can find a solution

samhatfield (approved):
I had to add a slight tweak so that the `TYPE` of `ectrans_field_view_gpu_{prec}` matches `trans_gpu_{prec}`. The latter is, by default `STATIC`. For various reasons, `ectrans_field_view_gpu_{prec}` must also be `STATIC` in this case.

PR #440, opened 2026-07-29
Update CI workflow

wdeconinck:
### Description

While working on PR #434 I made quite a number of improvements to the workflows that should really be a PR in its own right, also to make the review more manageable. This PR focuses on modernising the CI workflows in preparation.

This pull request introduces several improvements to the GitHub Actions CI workflows, focusing on smarter job execution to avoid duplicate builds, improved dependency caching, and more consistent and reproducible build environments. The main changes are the addition of a reusable action to skip redundant CI jobs, refactoring of caching logic, and standardization of compile flags for cross-runner consistency.

**Workflow execution improvements:**

* Added a reusable composite action (`.github/actions/should-run`) that determines whether a workflow job should run, preventing duplicate builds for the same commit when both `push` and `pull_request` events are triggered. This logic is now used in both `build.yml` and `build-hpc.yml` workflows to conditionally run CI jobs. 

**Dependency caching and environment setup:**

* Refactored dependency caching logic to use explicit cache key generation, separate restore and save steps, and enabled caching for more build matrix entries. The dependency cache is now only saved for Debug builds and when not already hit. 
* Moved environment setup and dependency installation steps to after repository checkout, and updated the Python setup and checkout actions to latest versions (`actions/setup-python@v7`, `actions/checkout@v7`).

**Build reproducibility and flags:**

* Standardized compile flags across runners and compilers to ensure reproducible builds, including setting `-march=x86-64-v3` for Linux (AVX2 instructions but not AVX512), `-march=native` for macOS, and appropriate math flags for different compilers, disabling FMA and strict IEEE compliance.
These flags are used for general compilation of the ECMWF packages, but also for FFTW, as that is important for bit-reproducibility across all AMD/Intel hardware used.

**Workflow matrix and configuration changes:**

* Simplified and standardized CMake options for matrix entries
* Enabled caching for more builds.
* Added explicit LAPACK vendor configuration and improved handling of BLAS/LAPACK paths on macOS. It seemed that a combination of Netlib and Accelerate was still used on macOS.

**HPC workflow improvements:**

* Updated the `build-hpc.yml` workflow to use the new conditional execution logic and standardized output/work directories using job and attempt IDs for better isolation.

### Contributor Declaration

By opening this pull request, I affirm the following:

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf-ifs/ectrans/blob/develop/contributor_license_agreement.md).
* The code follows the project's coding standards.
* I have performed self-review and added comments where needed.
[...]

PR #422, opened 2026-07-08
Create internal ectrans_util library for reuse in benchmark and test programs

wdeconinck:
### Description

This pull request introduces a new internal utility library (`ectrans_util`) for the repository, refactors the build system to use this library, and improves the checksum logic in benchmark programs. The checksum algorithm is now fiat's `fletcher16` algorithm.
The changes are primarily focused on improving code organization, maintainability, and internal tooling for test and benchmark executables. Below are the most important changes grouped by theme.

**Build System and Internal Utility Library**

* Added a new internal utility library `ectrans_util` (with CPU and GPU variants) under `src/util`, which centralizes common modules for use by test code and benchmark programs but is not installed as part of the public package. The build system now adds this directory and uses the appropriate variant depending on the target (CPU/GPU). (`src/CMakeLists.txt`, `src/util/CMakeLists.txt`, `src/programs/CMakeLists.txt`)
* Most routines are extracted from `ectrans-benchmark.F90`, making `ectrans-benchmark.F90` much more readable and maintainable.
* The api tests and ectrans-benchmark.F90 are now linking to this library. The `ectrans-benchmark-lam` and `ectrans-benchmark-ifs` still need to be adapted! Perhaps after a first round of review.
* Added a detailed `README.md` in `src/util` describing the purpose, main entry points, and scope of the new utility library. _**Please read the README!**_

**Checksum Calculation Improvements**

* Refactored the checksum logic from `dump_checksums` to use a new `fletcher16_hex` function instead of the previous `crc64`, improving output clarity and consistency. This is done for both `ectrans-benchmark-lam.F90` and  for `ectrans-benchmark` this is done in a new `ectrans_checksum_mod.F90` file.

### Contributor Declaration

By opening this pull request, I affirm the following:

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf-ifs/ectrans/blob/develop/contributor_license_agreement.md).
* The code follows the project's coding standards.
* I have performed self-review and added comments where needed.
* I have added or updated tests to verify that my changes are effective and functional.
* I have run all existing tests and confirmed they pass.

PR #442, opened 2026-07-29
Update checksums in ectrans-benchmark and ectrans-lam-benchmark

wdeconinck:
### Description

This PR is in preparation for #434, and adapts the benchmark checksums generation. This is stacked on top of #440 which should be merged first. It took only relevant changes from #434 related to checksums.

This pull request makes several important updates to the `ectrans_benchmark` program, primarily focused on improving the consistency and clarity of checksum outputs, enhancing decomposition invariance, and modernizing the checksum logic. The changes ensure that benchmark results are stable across different MPI decompositions and that checksum files are easier to interpret and less dependent on execution details. Additionally, some technical and usability improvements are included.

### Checksum calculation and output modernization

* Replaced the legacy CRC64-based checksums with a new, more portable `fletcher16_hex` checksum, and updated all relevant subroutines (`dump_checksums_pgp`, `dump_checksums_pgp_uv_3a_2`, `dump_checksums_psp`, etc.) to use this new approach. This includes the addition of a helper function `discontiguous_fletcher16_hex` to ensure compatibility with compilers that do not respect the `CONTIGUOUS` attribute. 

* Updated the checksum file writing logic to include clearer headers, use append mode correctly across iterations, and consistently format output for easier parsing and comparison. The command-line help and variable names were also clarified to remove references to CRC.

* Only one checksum file instead of separate ones for dir_trans and inv_trans

* Checksum format is now as follows for e.g. `ectrans-benchmark-cpu-dp_T47_O48_mpi0_omp1_callmode1_nfld10_nlev20_derivatives.checksums`:
   ```
    # --------------------------------------------
    # Iteration 1
    # --------------------------------------------
    6a60   20 # zgp (1-20)               u
    ef60   20 # zgp (21-40)              v
    d1b0  200 # zgp (41-240)             scalar-3d
    d1b0    1 # zgp (241)                scalar-2d
    f9d3  200 # zgp (242-441)            scalar-3d-ns
    f9d3    1 # zgp (442)                scalar-2d-ns
    31cf   20 # zgp (443-462)            u-ew
    5aaf   20 # zgp (463-482)            v-ew
    edad  200 # zgp (483-682)            scalar-3d-ew
    edad    1 # zgp (683)                scalar-2d-ew
    d5af   20 # zspvor (1-20)            vorticity
    d629   20 # zspdiv (1-20)            divergence
    2d5f   10 # zspscalar (1-10)         scalar-3d
    675e    1 # zspscalar (11)           scalar-2d
    # --------------------------------------------
    # Iteration 2
    # --------------------------------------------
    3bc9   20 # zgp (1-20)               u
    4fc9   20 # zgp (21-40)              v
    1103  200 # zgp (41-240)             scalar-3d
    1103    1 # zgp (241)                scalar-2d
    0a97  200 # zgp (242-441)            scalar-3d-ns
    0a97    1 # zgp (442)                scalar-2d-ns
    d603   20 # zgp (443-462)            u-ew
    dbe2   20 # zgp (463-482)            v-ew
    35e5  200 # zgp (483-682)            scalar-3d-ew
    35e5    1 # zgp (683)                scalar-2d-ew
    c29f   20 # zspvor (1-20)            vorticity
    c12a   20 # zspdiv (1-20)            divergence
    f0af   10 # zspscalar (1-10)         scalar-3d
    4962    1 # zspscalar (11)           scalar-2d
   ```
   The output is reduced a lot, merging identical checksums into a single line. The second column shows how many duplications are merged. Especially for this case with 10 fields and 20 levels and since there are 5 iterations, this reduces the number of lines significantly. 

### Decomposition invariance and field ordering

* Fixed V-set assignment for 2D scalar fields in both callmode 1 and callmode 2 to ensure that the ordering and ownership of fields are stable and independent of MPI decomposition. This eliminates spurious checksum differences between serial and parallel runs.

* Also checksums should now be identical between callmode 1, callmode 2 and field_api version. Variations in files are only within comment section (see format above, where `#` starts a comment), so that a smarter diff will be able to compare various modes 

### Interface and argument improvements

* Increased the maximum length for checksum file paths and string command-line arguments from 128 to 1024 characters, improving robustness for long filenames or paths.

### Usability and documentation

* Updated help messages and variable names to reflect the new checksum logic, removing references to CRC and clarifying descriptions for users.

### Minor technical improvements

* Passed additional flags (`lscders`, `luvder`, `append_checksums`) to checksum dumping routines to support more flexible and correct output handling. This is to have different callmode output be compatible.

### Contributor Declaration

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -65,6 +65,7 @@ integer(kind=jpim) :: iters   = 10  ! Number of iterations for transform test
 integer(kind=jpim) :: nfld    = 1   ! Number of 3D scalar fields
 integer(kind=jpim) :: nlev    = 1   ! Number of vertical levels
 integer(kind=jpim) :: iters_warmup = 3 ! Number of warm up steps (for which timing statistics should be ignored)
+integer(kind=jpim) :: iters_checksums = -1 ! Number of iterations for which checksum output is written
 
 integer(kind=jpim) :: nflevg  ! Total number of vertical levels
 
@@ -239,7 +240,9 @@ endif
 call get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, nlev, lvordiv, lscders, &
   &                             luvder, luseflt, nopt_mem_tr, nproma, npromatr, verbosity, &
   &                             ldump_values, lprint_norms, lmeminfo, nprtrv, nprtrw, ncheck, &
-  &                             lpinning, lfield_api, icall_mode, ldump_checksums, cchecksums_path, lalloperm)
+  &                             lpinning, lfield_api, icall_mode, ldump_checksums, iters_checksums, &
+  &                             cchecksums_path, lalloperm)
+if (iters_checksums < 0) iters_checksums = iters_warmup
 if (cgrid == '') cgrid = cubic_octahedral_gaussian_grid(nsmax)
 call parse_grid(cgrid, ndgl, nloen)
 nflevg = nlev
@@ -728,7 +731,7 @@ do jstep = 1, iters+iters_warmup
       &            ldscders=lscders, ldvorgp=lvordiv, lddivgp=lvordiv, lduvder=luvder, kproma=nproma)
   endif
 
-  if (ldump_checksums) then
+  if (ldump_checksums .and. jstep <= iters_checksums) then
     ! Remove trash at end of last block
     iend = ngptot - nproma * (ngpblks - 1)
     if (icall_mode == 1) then
@@ -808,7 +811,7 @@ do jstep = 1, iters+iters_warmup
       &            kvsetuv=ivset, kvsetsc2=ivsetsc2, kvsetsc3a=ivset, kproma=nproma)
   endif
 
-  if (ldump_checksums) then
+  if (ldump_checksums .and. jstep <= iters_checksums) then
 
     if (icall_mode == 1) then
       call dump_checksums_psp(filename=cchecksums_path, noutdump=noutdump_checksum, &
@@ -1264,6 +1267,8 @@ subroutine print_help(unit)
     & iterations (default = 10)"
   write(nout, "(a)") "    --niter-warmup      Number of warm up iterations,&
     & for which timing statistics should be ignored (default = 3)"
+  write(nout, "(a)") "    --niter-checksums  Number of iterations for which checksum output is&
+    & written, counting warmup iterations too (default = --niter-warmup)"
   write(nout, "(a)") "    -f, --nfld NFLD     Number of scalar fields (default = 1)"
   write(nout, "(a)") "    -l, --nlev NLEV     Number of vertical levels (default = 1)"
   write(nout, "(a)") "    --vordiv            Also transform vorticity-divergence to wind"
@@ -1325,6 +1330,7 @@ subroutine get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, n
   &                                   lscders, luvder, luseflt, nopt_mem_tr, nproma, npromatr, &
   &                                   verbosity, ldump_values, lprint_norms, lmeminfo, nprtrv, &
   &                                   nprtrw, ncheck, lpinning, lfield_api, icall_mode, ldump_checksums, &
+  &                                   iters_checksums, &
   &                                   cchecksums_path, lalloperm)
 
 #ifdef _OPENACC
@@ -1335,6 +1341,7 @@ subroutine get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, n
   character(len=16), intent(inout) :: cgrid ! Grid
   integer, intent(inout) :: iters           ! Number of iterations for transform test
   integer, intent(inout) :: iters_warmup    ! Number of iterations for transform test
+  integer, intent(inout) :: iters_checksums ! Number of iterations for which checksum output is written
   integer, intent(inout) :: nfld            ! Number of scalar fields
   integer, intent(inout) :: nlev            ! Number of vertical levels
   logical, intent(inout) :: lvordiv         ! Also transform vorticity/divergence
@@ -1393,6 +1400,11 @@ subroutine get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, n
         iters_warmup = get_int_value('--niter-warmup', iarg)
         if (iters_warmup < 0) then
           call parsing_failed("Invalid argument for --niter-warmup: must be >= 0")
+        end if
+      case('--niter-checksums')
+        iters_checksums = get_int_value('--niter-checksums', iarg)
+        if (iters_checksums < 0) then
+          call parsing_failed("Invalid argument for --niter-checksums: must be >= 0")
         end if
       ! Parse spectral truncation argument
       case('-t', '--truncation')
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -183,7 +183,6 @@ type(fields_lists) :: ylf
 logical :: ldump_values = .false.
 logical :: lpinning = .false.
 logical :: ldump_checksums = .false.
-character(len=1024) :: checksums_filename
 
 integer, external :: ec_mpirank
 logical :: luse_mpi = .true.
@@ -722,11 +721,10 @@ do jstep = 1, iters+iters_warmup
   if (ldump_checksums) then
     ! Remove trash at end of last block
     iend = ngptot - nproma * (ngpblks - 1)
-    write (checksums_filename,'(A)') trim(cchecksums_path)
     if (icall_mode == 1) then
       ! Remove trash at end of last block
       zgp (iend+1:, :, ngpblks) = 0
-      call dump_checksums_pgp(filename=checksums_filename, noutdump=noutdump_checksum, &
+      call dump_checksums_pgp(filename=cchecksums_path, noutdump=noutdump_checksum, &
                             & jstep=jstep, myproc=myproc, nproma=nproma, ngptotg=ngptotg, &
                             & zgp=zgp)
     else
@@ -734,7 +732,7 @@ do jstep = 1, iters+iters_warmup
       zgpuv (iend+1:, :, :, ngpblks) = 0
       zgp3a (iend+1:, :, :, ngpblks) = 0
       zgp2 (iend+1:, :, ngpblks) = 0
-      call dump_checksums_pgp_uv_3a_2(filename=checksums_filename, noutdump=noutdump_checksum, &
+      call dump_checksums_pgp_uv_3a_2(filename=cchecksums_path, noutdump=noutdump_checksum, &
                                     & jstep=jstep, myproc=myproc, nproma=nproma, ngptotg=ngptotg, &
                                     & lscders=lscders, luvder=luvder, zgpuv=zgpuv, zgp3a=zgp3a, zgp2=zgp2)
     endif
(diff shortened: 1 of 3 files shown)


src/programs/ectrans-benchmark.F90
```diff
@@ -691,7 +722,7 @@ program ectrans_benchmark
   if (ldump_checksums) then
     ! Remove trash at end of last block
     iend = ngptot - nproma * (ngpblks - 1)
-    write (checksums_filename,'(A)') trim(cchecksums_path)//'_inv_trans.checksums'
+    write (checksums_filename,'(A)') trim(cchecksums_path)
```
samhatfield: I think you can just delete this and pass `cchecksums_path` directly to `dump_checksums_*`. It is trimmed eventually so there's no need to use another filename variable.

src/programs/ectrans-benchmark.F90
```diff
@@ -768,17 +799,18 @@ program ectrans_benchmark
   endif
 
   if (ldump_checksums) then
-    write (checksums_filename,'(A)') trim(cchecksums_path)//'_dir_trans.checksums'
+    write (checksums_filename,'(A)') trim(cchecksums_path)
```
samhatfield: Same as above.

src/programs/ectrans-lam-benchmark.F90
```diff
@@ -640,7 +640,7 @@ program ectrans_lam_benchmark
     ! Remove trash at end of last block
     iend = ngptot - nproma * (ngpblks - 1)
     zgp2 (iend+1:, :, ngpblks) = 0
-    write (checksums_filename,'(A)') trim(cchecksums_path)//'_inv_trans.checksums'
+    write (checksums_filename,'(A)') trim(cchecksums_path)
```
samhatfield: Same here.

src/programs/ectrans-lam-benchmark.F90
```diff
@@ -727,11 +727,11 @@ program ectrans_lam_benchmark
   endif
 
   if (ldump_checksums) then
-    write (checksums_filename,'(A)') trim(cchecksums_path)//'_dir_trans.checksums'
+    write (checksums_filename,'(A)') trim(cchecksums_path)
```
samhatfield: And here.

src/programs/ectrans-lam-benchmark.F90
wdeconinck: Done.

src/programs/ectrans-benchmark.F90
wdeconinck: Done.

src/programs/ectrans-benchmark.F90
wdeconinck: Done.

src/programs/ectrans-lam-benchmark.F90
wdeconinck: Done.

samhatfield:
Benchmark of the ecTrans test suite at `develop`:
```
Label Time Summary:
ectrans       = 327.47 sec*proc (416 tests)
executable    =  18.20 sec*proc (36 tests)
mpi           = 245.13 sec*proc (256 tests)
openmp        = 324.71 sec*proc (405 tests)
script        = 309.27 sec*proc (380 tests)

Total Test time (real) = 329.96 sec
```

Benchmark of this branch:
```
Label Time Summary:
ectrans       = 273.26 sec*proc (416 tests)
executable    =  19.78 sec*proc (36 tests)
mpi           = 206.82 sec*proc (256 tests)
openmp        = 270.63 sec*proc (405 tests)
script        = 253.48 sec*proc (380 tests)

Total Test time (real) = 275.57 sec
```

About 17% improvement, presumably because of the reduced number of I/O operations when writing checksums?

wdeconinck:
> About 17% improvement, presumably because of the reduced number of I/O operations when writing checksums?

Ah good news for once!
Could be related to
- the checksum algorithm: fletcher16 could be faster than crc64. In fact even more checksums are actually done compared to the develop branch, so puzzling...
- the checksum writing to file. The maximum number of lines for only some of the most compute heavy tests was on the order of ~4000 lines before to ~80. Which file system is used for this test? Could also be randomly different file system load on shared file system.

wdeconinck:
The reason to have the more compressed checksum format is evident when storing reference files for all runners. The compression is evident from this change: https://github.com/ecmwf-ifs/ectrans/pull/437/changes/5a13fffbe0932a8392247b7e3289d1274e13fb5f

dhaumont:
Nice improvement, Willem !
Regarding the code itself, I believe we should consider to move all the new checksum code into an helper file
`src/util/checksum_helper.F90`, as you added a lot of code to ectrans_benchmark.F90.
This would also allow to remove duplicated code between ectrans_benchmark and ectrans_lam_benchmark.

src/programs/ectrans-lam-benchmark.F90
```diff
@@ -1561,12 +1567,126 @@ subroutine dump_spectral_field(jstep, myproc, nspec2, nsmax, nmsmax, fld, kvset,
 end subroutine dump_spectral_field
 
 
+!===================================================================================================
+
+function discontiguous_fletcher16_hex(field) result(checksum_hex)
```
dhaumont, line 1592: duplicated code

wdeconinck:
Thanks @dhaumont for reviewing! Regarding code duplication, I completely 100% agree, and already had started PR #422 with the ectrans_util library to avoid this; I think you commented on it already.
Also there I did such checksumming changes. But I changed my mind then thinking it is better to split the checksum changes out first and have the more rigorous tests available, and then make sure that the refactoring with the util library would be mandated to stay bit-identical with reference tests across all CI available with #434.
The PR #422 will then be rebased and make sure that the checksum format stays identical for the added confidence of refactoring.

src/programs/ectrans-lam-benchmark.F90
wdeconinck: Agreed, but only temporary for now, and will be removed with #422 still to be rebased and adapted after this is merged.
This unfortunate routine is a workaround for nvhpc-26.3 compiler not respecting the CONTIGUOUS attribute.

dhaumont:
> But I changed my mind then thinking it is better to split the checksum changes out first and have the more rigorous tests available, and then make sure that the refactoring with the util library would be mandated to stay bit-identical with reference tests across all CI available with #434.

This makes a lot of sense, thanks for the explanation.

wdeconinck:
@lukasm91 
I have created a reproducer for the NVHPC bug requiring the `discontiguous_fletcher16_hex` function: 
https://github.com/wdeconinck/compiler_bug_reproducers/tree/main/003

wdeconinck:
@samhatfield I reduced the test time even more now, making sure we only have maximum 2 iterations for which checksums are created at all time in the CI. Because checksums show deviations immediately in principle only 1 iteration should be OK, but let's keep 2 for now.

samhatfield:
Here's an updated benchmark: (Not comparable with the previous - I can't remember which machine I ran that on. This is running on my M1 MacBook Pro.)

`develop`:
```
Label Time Summary:
ectrans       = 234.06 sec*proc (414 tests)
executable    =  14.41 sec*proc (34 tests)
mpi           = 151.68 sec*proc (256 tests)
openmp        = 230.54 sec*proc (405 tests)
script        = 219.65 sec*proc (380 tests)

Total Test time (real) = 241.66 sec
```

`checksum_updates`:
```
Label Time Summary:
ectrans       = 149.02 sec*proc (414 tests)
executable    =  12.50 sec*proc (34 tests)
mpi           = 102.34 sec*proc (256 tests)
openmp        = 147.24 sec*proc (405 tests)
script        = 136.52 sec*proc (380 tests)

Total Test time (real) = 156.65 sec
```
That's a 35% reduction.

src/programs/ectrans-benchmark.F90
```diff
@@ -357,8 +370,20 @@ program ectrans_benchmark
...
+! and on the spectral decomposition. That made the synthetic benchmark data layout change between serial
+! and MPI runs even though the logical test case was intended to be identical.
+! This showed up as checksum differences for zspsc2 between mpi0 and mpi4: the actual field being
+! transformed was the same benchmark field, but it was attached to a different V-set depending on the
+! decomposition. Since zspsc2 is a single 2D scalar field and has no vertical level distribution to mirror,
+! assigning it consistently to V-set 1 gives it a canonical owner. That makes the checksum independent of
+! execution decomposition, which is one of the invariants the checksum tests are meant to enforce.
+! In other words: this change removes decomposition metadata from the benchmark result. The checksum
+! should reflect transform output for the logical field, not where the benchmark happened to place that
+! single 2D scalar in a particular MPI layout.
```
samhatfield: I don't agree with this. The standalone 2D scalar field is gathered to task 1 with a call to `gath_spec` before the checksum calculation to ensure the same result regardless of the decomposition. Assigning this field to the first V-set rather than the last should be harmless, but also not necessary. We worked hard to ensure checksum differences between mpi0 and mpi4 were eliminated - have they reappeared?

src/programs/ectrans-benchmark.F90
```diff
@@ -484,7 +509,25 @@ program ectrans_benchmark
...
+  ! a V-set assignment that could disturb the canonical field order under NPROMATR blocking.
+
+  ! Assigning the extra 2D scalar to ivset(nflevg) places it on the same V-set as the final 3D scalar level,
+  ! preserving the intended combined scalar ordering across serial and MPI decompositions.
+  ! This keeps callmode 1’s packed scalar array stable when field blocking is enabled, so checksum files remain
+  ! independent of execution decomposition and NPROMATR.
+
+  ! The important distinction from the callmode 2 fix is that callmode 1 needs order preservation inside a single combined
+  ! scalar array. The extra scalar is not independent in the memory/interface layout; it follows the 3D scalar levels,
+  ! so its V-set assignment must be compatible with that sequence.
```
samhatfield: Again, I wasn't aware that we have a reproducibility issue between mpi0 and mpi4... Is this something you've observed @wdeconinck?

src/programs/ectrans-lam-benchmark.F90
```diff
@@ -58,12 +58,23 @@ program ectrans_lam_benchmark
...
+  integer(kind=jpim) :: count = 0
+end type checksum_run_type
+
 ! Default parameters
 integer(kind=jpim) :: nlon    = 128   ! Zonal dimension
 integer(kind=jpim) :: nlat    = 128   ! Meridional dimension
 integer(kind=jpim) :: nsmax   = 0   ! Spectral meridional truncation
 integer(kind=jpim) :: nmsmax  = 0   ! Spectral zonal truncation
 integer(kind=jpim) :: iters   = 10  ! Number of iterations for transform test
+integer(kind=jpim) :: iters_warmup = 0 ! Number of warm up steps (for which timing statistics should be ignored)
```
samhatfield: A warmup phase in the main loop hasn't actually been implemented.

samhatfield (changes requested):
All approved pending resolution of the discussion points above.

src/programs/ectrans-benchmark.F90
wdeconinck: Hmm to be honest, this was a fix for when I was using OpenBLAS, instead of Netlib's BLAS by accident on macOS, thinking that was the reference BLAS. This fixed the discrepancy and I thought it was nice to have the decomposition identical.
As you said it should not be necessary. I can look into it again to see if it is actually required.

src/programs/ectrans-lam-benchmark.F90
wdeconinck: True, this is a TODO. I thought to already make a placeholder here to at least have the following logic be the same with the global benchmark: when --iters-checksum is not passed, it defaults to --iters-warmup
```Fortran
if (iters_checksums < 0) then
  if (iters_warmup > 0) then
    iters_checksums = iters_warmup
  else 
    iters_checksums = iters
  endif
endif
```

src/programs/ectrans-benchmark.F90
wdeconinck: I have reverted this and above change in latest commit. The discrepancy I saw earlier had to do with me using OpenBLAS instead of Netlib (reference) BLAS. With Netlib BLAS the comparisons are all fine.

src/programs/ectrans-lam-benchmark.F90
wdeconinck: @samhatfield I have now implemented the iters_warmup handling like in ectrans-benchmark in latest commit.

wdeconinck:
@samhatfield I believe all discussion points are now resolved.

src/programs/ectrans-benchmark.F90
```diff
@@ -229,7 +240,9 @@ program ectrans_benchmark
 call get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, nlev, lvordiv, lscders, &
   &                             luvder, luseflt, nopt_mem_tr, nproma, npromatr, verbosity, &
   &                             ldump_values, lprint_norms, lmeminfo, nprtrv, nprtrw, ncheck, &
-  &                             lpinning, lfield_api, icall_mode, ldump_checksums, cchecksums_path, lalloperm)
+  &                             lpinning, lfield_api, icall_mode, ldump_checksums, iters_checksums, &
+  &                             cchecksums_path, lalloperm)
+if (iters_checksums < 0) iters_checksums = iters_warmup
```
samhatfield:
```suggestion
if (iters_checksums < 0) then
  if (iters_warmup > 0) then
    iters_checksums = iters_warmup
  else 
    iters_checksums = iters
  endif
endif
```
To match the equivalent in `ectrans-lam-benchmark`.

samhatfield (changes requested):
Nice job @wdeconinck 👍🏻 Tidy, smart, and efficient! Approved subject to one minor final comment.

src/programs/ectrans-benchmark.F90
wdeconinck: Done now, and updated the help/usage message too.

samhatfield (approved):

PR #443, opened 2026-07-29
Remove more dead GPU code

samhatfield:
More unreachable code to be deleted.


wdeconinck (approved):
Thanks Sam.
I suppose this may need to implemented some day, but then the cpu example is present still.

samhatfield:
I still don't know what `CDMAP` actually does...

Issue #428, opened 2026-07-16
Cleaner directive selection


samhatfield:
Currently in the GPU code path of ecTrans we provide both OpenMP and OpenACC offloading directives. These are guarded by `#ifdef` blocks to ensure only one is compiled at a time. I am not sure of the original reason for this as I think it was made by @anmrde and adopted by @lukasm91. Presumably compilers get (or used to get) confused when both are present.

Ideally compilers would have two switches for controlling directives, e.g. `-acc` or `-omp`, or equivalent. Activating one would ensure that the other is treated as a comment, and we wouldn't need any secondary mechanism. This would be the cleanest solution. Do you know if nvfortran nowadays provides a clean way to choose between these two directives @lukasm91, when both are present in source unguarded?

Meanwhile I've been exploring alternatives that could improve readability. So far the best solution I've come up with is to define a header file `acc_or_omp.h` like so:

```c
#ifdef ACCGPU
#define _ACC !$ACC
#define _OMP !
#else
#define _ACC !
#define _OMP !$OMP
#endif
```

Then all directives would be written like the following:

```
_OMP TARGET TEAMS DISTRIBUTE PARALLEL DO ...
_ACC PARALLEL LOOP ...
```

This improves readability over the current solution. E.g., compare
```fortran
#ifdef ACCGPU
    !$ACC WAIT(1)
#endif
    IF (PRESENT(PGP)) THEN
#ifdef OMPGPU
        !$OMP TARGET UPDATE TO(PGP) IF (.NOT. LLPGP_ON_GPU)
#endif
#ifdef ACCGPU
        !$ACC UPDATE DEVICE(PGP) IF (.NOT. LLPGP_ON_GPU)
#endif
    ENDIF
    IF (PRESENT(PGPUV)) THEN
#ifdef OMPGPU
        !$OMP TARGET UPDATE TO(PGPUV) IF (.NOT. LLPGP_ON_GPU)
#endif
#ifdef ACCGPU
        !$ACC UPDATE DEVICE(PGPUV) IF (.NOT. LLPGP_ON_GPU)
#endif
    ENDIF
    IF (PRESENT(PGP2)) THEN
#ifdef OMPGPU
        !$OMP TARGET UPDATE TO(PGP2) IF (.NOT. LLPGP_ON_GPU)
#endif
#ifdef ACCGPU
        !$ACC UPDATE DEVICE(PGP2) IF (.NOT. LLPGP_ON_GPU)
#endif
[...]

PR #445, opened 2026-07-30
Introduce a metadirective language

samhatfield:
This PR is for testing ideas for a tidier directive style in the GPU subtree, addressing #428.


dhaumont:
@pmarguinaud @ddegrauwe

dhaumont:
I'm not sure that sed is the best choice for this task, it's usage for the generation of dp and sp is already not very easy for maintenance.
Why don't you use Loki technology , which is already mature and used by ECMWF for the rest of the IFS codebase ? I'm a bit afraid that you will reinvent the wheel with this sed approach.

awnawab:
For this kind of simple templating, adding loki as a build dependency would be completely overkill imo. That would add significant time and complexity to the build system.

I also think there is a genuine need for a solution like this. Not only do we need to support both the openacc and openmp target offload models, but also each compiler toolchain will deviate from the standard of these programming models in different subtle ways. Without some kind of templating engine we would quickly descend into ifdef hell as we add support for more platforms and compiler toolchains.

That's also why we introduced a python based templating engine in field_api. The benefit of python is greater composability of the templates, but it does require fypping every file, which perhaps wouldn't be suitable for ectrans (whereas f-api was already fypp'd). So on balance I agree with @samhatfield that this is quite a good approach for ectrans.

awnawab:
Ultimately, the unavoidable source of complexity here is that you have to map the template to the eventual preprocessed pragma in your head whilst programming. That would be true whether we insert loki directives and ask loki to translate them for the target toolchain, or whether we use a simple bespoke solution like the one herein. The latter has the benefit that it is contained entirely within ectrans, whereas if we were to use loki, a developer would then need to look within loki to try and understand how the template maps to actual pragmas.

dhaumont:
@awnawab Thanks for the explanation of your motivations and the balance you try to find,  I think I got your point. My fear is that you will have more needs than just a string replacement tool to achieve the portability goals, but maybe I'm too pessimistic and sed will just be enough.

samhatfield:
I think a sed-based approach will be enough if we continue to use only a very limited subset of the directives' "grammar". In ecTrans we essentially use four constructs:

- `DATA`
- `PARALLEL`
- `CREATE DATA / EXIT DATA`
- `USE_DEVICE_PTR`

All of these are expressed very similarly between OpenMP and OpenACC, with the major difference being the meaning of the `SHARED` construct. As long as we stick to these, I don't think we need a sophisticated approach.

The directive aspects of ecTrans have not changed for over 2 years now so I think it's safe to say that they are "finished" for the foreseeable future.

samhatfield:
I'm going to park this PR as it not getting much traction and we've got other things to focus on.

PR #449, opened 2026-08-13
Add support for Fujitsu compiler

samhatfield:
https://github.com/ecmwf-ifs/ectrans/commit/c63ba22bdd8af367183ddb51daa6331fc2f530d9 neglected to add the Fujitsu compiler's "don't add a `main` program" flag for Fujitsu, so here it is. This allows us to build the API test suite on Fugaku.


wdeconinck (approved):
Nice!

PR #451, opened 2026-08-18
Fix multithreaded building with crayftn

samhatfield:
crayftn generates "*.acc.o" intermediate files when compiling accelerated code. The intermediate file name is based on the input source file name and seems to be written into the same directory from where the compile command is invoked. This means that when building SP and DP libraries in parallel (normal when both are enabled and `make` is invoked with more than one thread), intermediates for SP and DP source files have the same name and clash. This causes a race condition.

This commit appends _sp or _dp to the source file name when it is put through the generator to ensure the intermediates don't clash.

Up until now (for three years...) we have always had to build the ecTrans backend in serial on LUMI without really understanding why. I have finally found the reason.

Of course I should submit a bug report to Cray, but we've no idea how long they will take to release a fix.

### DIFF
--- a/.github/workflows/build-hpc.yml
+++ b/.github/workflows/build-hpc.yml
@@ -107,7 +107,6 @@ jobs:
               - ROCFFT_RTC_CACHE_PATH=$PWD/../../rocfft_kernel_cache.db
               - MPICH_GPU_SUPPORT_ENABLED=1
               - MPICH_SMP_SINGLE_COPY_MODE=NONE
-              - CMAKE_BUILD_PARALLEL_LEVEL=1
             ctest_options: -E compare_checksums # We don't bother with reproducibility checks on LUMI
             unsupported_dependency: ecmwf-ifs/field_api # We don't build Field API on LUMI for now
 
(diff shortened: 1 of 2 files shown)


samhatfield:
I tested this fix by executing the following in the build directory of ecTrans on LUMI:
```bash
for i in {0..10}; do
    echo "Iteration $i"; echo ""
    find ectrans/src/trans/gpu/CMakeFiles/ectrans_gpu_??.dir -name "*.o" -exec rm {} \;
    make -j128 > /dev/null
done
```
This was run in an interactive job on LUMI-C. Every build passed. On the develop branch, every build fails with errors such as `spnsdead_mod_1.acc.o: No such file or directory`.

wdeconinck:
Looks good! Thanks for figuring that one out! All CI still seems to work.

wdeconinck (approved):

samhatfield:
It occurred to me that we might not actually be doing any building in parallel. Let's see.

samhatfield:
Previous build + configure time: 15 minutes.

New build + configure time, with 4 threads: 4 minutes.

samhatfield:
We're now down to 2 minutes build + configure time, using Ninja. Ninja automatically builds in parallel so must be using more than 4 threads.

PR #452, opened 2026-08-20
Tidy unused variables

samhatfield:
`TPM_DISTR` is already enormous without these variables which are never used.


wdeconinck:
What is the history of these variables? What were they (intended to be) used for?

samhatfield:
> What is the history of these variables? What were they (intended to be) used for?

It looks like they're the start of a feature that was never finished, to somehow treat "tendency" fields different to "full" fields. This is probably also where `D%NTMAX` comes from (different do `D%NSMAX`). `NTMAX` seems to have a meaningful use that I don't fully understand, so I've left that in. But these other variables are just allocated and then ignored. They were added to `TPM_DISTR` between 11 and 22 years ago and either the functionality to use them was never written or it was written and removed at some point.

PR #454, opened 2026-08-20
Remove LEDIR work array

samhatfield:
This PR removes a number of copies by restructuring the main Legendre transform work arrays. These arrays are transposed and the fields dimension is halved for IM = 0, to match the work array `ZB` in `LEDIR`. These means we can eliminate `ZB` and operate directly on `ZAIA` / `ZSIA`.

It depends on https://github.com/ecmwf-ifs/ectrans/pull/423.


wdeconinck:
Hi @samhatfield I wonder why this was not done like this in the first place; one thing that comes to mind is that this was required for efficient vectorisation down the line?
Can you perhaps look at some vectorisation reports, and ultimately timings?

wdeconinck:
Oh I see GitHub bot added some timings. NEAT!

samhatfield:
> Oh I see GitHub bot added some timings. NEAT!

I tested out the new [benchmarking workflow](https://github.com/ecmwf-ifs/ectrans/pull/474) on this PR!

> Hi @samhatfield I wonder why this was not done like this in the first place; one thing that comes to mind is that this was required for efficient vectorisation down the line?
Can you perhaps look at some vectorisation reports, and ultimately timings?

Do you mean e.g., the loops which access `PAIA`/`PSIA` could possibly be more effectively vectorised when those arrays are arranged as (fields, latitudes) instead of (latitudes, fields)?

wdeconinck:
Yes, but perhaps it doesn't seem to matter, and it's actually faster now. Perhaps you can do a few more benchmarks to convince ourselves?

PR #456, opened 2026-08-24
Make trans_mpi_set_comm and trans_specnorm public; otherwise gfortran…

amckinstry:
…-16 hides them breaking atlas

### Contributor Declaration

### DIFF
--- a/src/transi/transi_module.F90
+++ b/src/transi/transi_module.F90
@@ -83,6 +83,7 @@ public :: GathSpec_t
 public :: DistSpec_t
 public :: VorDivToUV_t
 public :: trans_use_mpi
+public :: trans_mpi_set_comm
 public :: trans_set_handles_limit
 public :: trans_set_radius
 public :: trans_set_leq_regions
@@ -90,6 +91,7 @@ public :: trans_set_nprtrv
 public :: trans_set_nprgpew
 public :: trans_init
 public :: trans_setup
+public :: trans_specnorm
 public :: trans_inquire
 public :: trans_dirtrans
 public :: trans_dirtrans_adj


samhatfield:
Thanks @amckinstry! This looks reasonable. I'll leave the final say to @wdeconinck.

src/transi/transi_module.F90
```diff
@@ -83,13 +83,15 @@ module trans_module
 public :: DistSpec_t
 public :: VorDivToUV_t
 public :: trans_use_mpi
+public :: trans_mpi_set_comm
```
wdeconinck, line 86:
```suggestion
public :: trans_set_mpi_comm
```

wdeconinck: This is a typo and does not compile.

wdeconinck (changes requested):
Thanks @amckinstry for this change. You're right it should be public as are the other routines.
But I'm not quite understanding why making these routines public in the transi_module would affect the C-interface (which is the only thing used in Atlas).
What is the error in atlas that you're seeing. Could you post this?

Currently this change doesn't compile as it should be trans_set_mpi_comm instead of trans_mpi_set_comm.

wdeconinck:
Since I don't have permissions to edit this PR, I opened a new PR (#464) containing this with the suggested compilation fix.

Still @amckinstry I would like to know the error message. Many thanks!

amckinstry:
The error seen was that the symbol was hidden in the library - so the linking failed. It appears that GNU gcc/gfortran < 16 made all symbols visible, public or not, in the library, leaving it to the compiler to report any error. Making the symbol non-visible adds an extra layer of checking and in theory speeds up linking.

wdeconinck:
Thanks for the clarification, seems like a compiler bug. Closing this as #464 superseded this. Thanks @amckinstry !

PR #457, opened 2026-08-26
SEAM biperiodic extension method

dhaumont:
### Description

New biperiodic extension method by based upon Symetric+Antisymetric Extension
and a Mixing (SEAM). It is activated by the NAMFPC namelist key LFPSEAM=.true.

Original PR from Fabrice Voitus: https://github.com/ACCORD-NWP/ectrans/pull/13 

### Contributor Declaration


dhaumont:
@fabvoit: can you test this PR and verify it works in combination with your SEAM development ?

Regarding testing,  we are supposed to add a automated test for the new features. For now, it basically mean adding the option to ectrans-lam-benchmark, and add a new unit test launching ectrans-lam-benchmark with the option activated ( https://github.com/ecmwf-ifs/ectrans/blob/develop/src/programs/ectrans-lam-benchmark.F90)
@ddegrauwe  Do you think it's necessary to do this for this feature ?

samhatfield:
@dhaumont could you also add the three new arguments to https://github.com/ecmwf-ifs/ectrans/blob/develop/src/etrans/gpu/biper/external/fpbipere.F90 to make sure the signatures match between CPU and GPU? You don't need to implement the new behaviour itself, you could just make it print a warning that those options are not supported on GPU yet.

dhaumont:
> @dhaumont could you also add the three new arguments to https://github.com/ecmwf-ifs/ectrans/blob/develop/src/etrans/gpu/biper/external/fpbipere.F90 to make sure the signatures match between CPU and GPU? You don't need to implement the new behaviour itself, you could just make it print a warning that those options are not supported on GPU yet.

It's commited but I just saw it won't compile with nvidia compiler. Sorry, I'm fixing it tomorrow morning.

dhaumont:
> > @dhaumont could you also add the three new arguments to https://github.com/ecmwf-ifs/ectrans/blob/develop/src/etrans/gpu/biper/external/fpbipere.F90 to make sure the signatures match between CPU and GPU? You don't need to implement the new behaviour itself, you could just make it print a warning that those options are not supported on GPU yet.
> 
> It's commited but I just saw it won't compile with nvidia compiler. Sorry, I'm fixing it tomorrow morning.

Fixed 7136284438271464947a9ac35dec0d9275a102ca

dhaumont:
@ddegrauwe Will you have time to review this ?

src/etrans/cpu/internal/eseam_mod.F90
samhatfield, line 1: This file should be added to `src/etrans/sedrenames.txt`. Otherwise if you try to link a program against the single- and double-precision versions of etrans simultaneously there will be duplicate `eseam` symbols.

samhatfield:
```diff
diff --git a/src/etrans/sedrenames.txt b/src/etrans/sedrenames.txt
index a554fd1f..935ecf1b 100644
--- a/src/etrans/sedrenames.txt
+++ b/src/etrans/sedrenames.txt
@@ -15,6 +15,7 @@ s/EDIST_GRID( *($|\(| |\*))/EDIST_GRID_VARIANTDESIGNATOR\1/g
 s/DIST_SPEC_CONTROL_MOD/DIST_SPEC_CONTROL_MOD_VARIANTDESIGNATOR/g
 s/edist_spec( *($|\(| |\*))/edist_spec_VARIANTDESIGNATOR\1/g
 s/EDIST_SPEC( *($|\(| |\*))/EDIST_SPEC_VARIANTDESIGNATOR\1/g
+s/ESEAM_MOD/ESEAM_MOD_VARIANTDESIGNATOR/g
 s/FOURIER_IN_MOD/FOURIER_IN_MOD_VARIANTDESIGNATOR/g
 s/FOURIER_INAD_MOD/FOURIER_INAD_MOD_VARIANTDESIGNATOR/g
 s/FOURIER_OUT_MOD/FOURIER_OUT_MOD_VARIANTDESIGNATOR/g
```

dhaumont: Indeed, I  understand now why the build server was failing. 
https://github.com/ecmwf-ifs/ectrans/actions/runs/33078828455

Nice catch Sam, thanks for the hint !
Solved in daac55f2ade6de8303cf7b02e29fb574436283a6

src/etrans/include/etrans/fpbipere.h
```diff
@@ -11,7 +11,7 @@
 
 INTERFACE
 SUBROUTINE FPBIPERE(KDLUX,KDGUX,KDLON,KDGL,KNUBI,KD1,PGPBI,KDADD,LDZON,&
-& LDBOYD,KDBOYD,PLBOYD,PBIPOUT)
+& LDBOYD,KDBOYD,PLBOYD,LDSEAM,KDSEAM,PLSEAM,PBIPOUT)
```
ddegrauwe: The PBIPOUT argument doesn't appear in the F90 file(s).

dhaumont: Indeed. I'm fixing it. Note that this mistake is already present in develop.

samhatfield (changes requested):
Approved subject to passing test suite and resolution of @ddegrauwe's comment above, about inconsistent header subroutine signatures.

src/etrans/include/etrans/fpbipere.h
dhaumont: Fixed in b0d3952992419e4ce69b50d4af6b4b22a4ec1ef7

samhatfield:
Strictly speaking, modification of interface headers such as in commit [b0d3952](https://github.com/ecmwf-ifs/ectrans/pull/457/commits/b0d3952992419e4ce69b50d4af6b4b22a4ec1ef7) is an API change. Can you confirm @dhaumont that removal of `PBIOUT` doesn't break compilation for some user of `FPBIPERE`?

dhaumont:
@samhatfield 
I think it's safe, as I did not  change the interface in the F90 file, but only in the header file `fpbipere.h`.

In ectrans, as far as I can see, this header was not in use.

In IFS/IAL, it is in use, but always without the `PBIOUT` argument:
```
$ cd IAL
$ grep -R "CALL FPBIPERE" -A1  | grep -v ENDIF | grep -v CASE | grep -v IPROC | grep -v "!" 

./arpifs/fullpos/sufpgeometry.F90:  CALL FPBIPERE(YDFPGEOMETRY%YFPUSERGEO(1)%NFPLUX,YDFPGEOMETRY%YFPUSERGEO(1)%NFPGUX,YDFPGEOMETRY%YFPUSERGEO(1)%NLON, &
./arpifs/fullpos/sufpgeometry.F90-   & YDFPGEOMETRY%YFPUSERGEO(1)%NLAT,1,NFPRGPG_DEP,ZFPGM_DEP,0,LDZON=.TRUE.,LDBOYD=.TRUE.,KDBOYD=IBOYD,PLBOYD=RFPBSCAL)
--
./arpifs/fullpos/sufpg2.F90:      CALL FPBIPERE(NFPLUX(J),NFPGUX(J),IFPLSUR,IFPGSUR,1,ILEN,PFPGM(IPTR:IPTR2),0,LDZON=.TRUE.)
--
./aladin/fullpos/fpezone.F90:        CALL FPBIPERE(NFPLUX(1),NFPGUX(1),NLON(1),NLAT(1),IFIELD,NFPRGPG,&
./aladin/fullpos/fpezone.F90-         & ZREALIN(1,JFIELD),0,LDZON=LLCIE,LDBOYD=.FALSE.)
--
./aladin/fullpos/ebipos.F90:    CALL FPBIPERE(NDLUXG,NDGUXG,NDLON,NDGLG,INFG,NGPTOTG,ZGPL,0,LDZON=.TRUE.)
--
./aladin/fullpos/ebipos.F90:      CALL FPBIPERE(NDLUXG,NDGUXG,NDLON,NDGLG,INFG,NGPTOTG,ZGPG,0,LDZON=.TRUE.)
--
./aladin/fullpos/fpboyd.F90:    CALL FPBIPERE(NFPLUX(1),NFPGUX(1),NLON(1),NLAT(1),INFL,NFPRGPG_DEP,ZREALIN,0,LDZON=LLCIE, &
./aladin/fullpos/fpboyd.F90-     & LDBOYD=.TRUE.,KDBOYD=JBOYD,PLBOYD=PFPBSCAL)
./biper/programs/test_TwoTrunc.F90:CALL FPBIPERE(NDLUX1,NDGUX1,NDLON1,NDGL1,1,NDLON1*NDGL1,GGPBI1,0,.FALSE.)
--
./biper/programs/test_TwoTrunc.F90:CALL FPBIPERE(NDLUX2,NDGUX2,NDLON2,NDGL2,1,NDLON2*NDGL2,GGPBI2,0,.FALSE.)
--
./biper/programs/test_Ezones.F90:CALL FPBIPERE(NDLUX1,NDGUX1,NDLON1,NDGL1,1,NDLON1*NDGL1,GGPBI1,0,.FALSE.)
--
./biper/programs/test_Ezones.F90:CALL FPBIPERE(NDLUX2,NDGUX2,NDLON2,NDGL2,1,NDLON2*NDGL2,GGPBI2,0,.FALSE.)
--
./biper/programs/test_Biper.F90:CALL FPBIPERE(NDLUX,NDGUX,NDLON,NDGL,NNUBI,ND1,GGPBI,IADD,LNZON)
```

To be 100% sure, I'm checking now what happens during compilation of IAL/IFS if I add PBIOUT to one of these calls.

dhaumont:
> To be 100% sure, I'm checking now what happens during compilation of IAL/IFS if I add PBIOUT to one of these calls.

As a stress test, I have tried to compile IAL with this change (intel compiler):
```
diff --git a/aladin/fullpos/fpezone.F90 b/aladin/fullpos/fpezone.F90
index 3056bd61b7..eca71e1e46 100644
--- a/aladin/fullpos/fpezone.F90
+++ b/aladin/fullpos/fpezone.F90
@@ -150,7 +150,7 @@ DO JCH=1,ICHUNKS
       IFLDA=IBFAL-1+JFIELD
       IF (LDBIP(IFLDA)) THEN
         CALL FPBIPERE(NFPLUX(1),NFPGUX(1),NLON(1),NLAT(1),IFIELD,NFPRGPG,&
-         & ZREALIN(1,JFIELD),0,LDZON=LLCIE,LDBOYD=.FALSE.)
+         & ZREALIN(1,JFIELD),0,LDZON=LLCIE,LDBOYD=.FALSE.,PBIPOUT=ZREALIN)
       ENDIF
     ENDDO
   ENDIF
```

With the old `fpbipere.h`,  it compiled and linked (even if it shouldn't). I didn't run it but I expect a crash at runtime,
With the new `fpbipere.h`, compilation failed, as expected:
```
ial_dp/aladin/fullpos/fpezone.F90(153): error #6627: This is an actual argument keyword name, and not a dummy argument name.   [PBIPOUT]
         & ZREALIN(1,JFIELD),0,LDZON=LLCIE,LDBOYD=.FALSE.,PBIPOUT=ZREALIN)
```

On top of this, I removed the last argument so no side-effect can happen.

Conclusion: Given all that facts, I consider this change to be really safe.

samhatfield:
Thanks for investigating that @dhaumont - always good to make sure!

We just made a [2.0.0 release](https://github.com/ecmwf-ifs/ectrans/releases/tag/2.0.0) of ecTrans. This contained some really minor but non-bit-ID changes on top of 1.9.0, which we also released yesterday.

Right now this PR targets develop. That means it will appear in the next official release, 2.1.0. This release will NOT be bit identical with 1.9.0, nor earlier versions. They are scientifically equivalent though. We are talking about minor rounding error changes.

Question: do you need the feature in this PR in a version of ecTrans that is bit identical with 1.9.0? If so then we can create a 1.10.0 release with these changes.

dhaumont:
> Question: do you need the feature in this PR in a version of ecTrans that is bit identical with 1.9.0? If so then we can create a 1.10.0 release with these changes.

Thank you for the update, Sam. 
As you know, we are stucked on an old version of ecTrans, before  FFT992 was removed.
Is FFT992 already put back in the 2.0.0 release ? Otherwise I'm afraid there is no point to deliver any special version for us, as we can't use it.
@AlexandreMary Do you agree ?

samhatfield:
> Thank you for the update, Sam. As you know, we are stucked on an old version of ecTrans, before FFT992 was removed. Is FFT992 already put back in the 2.0.0 release ? Otherwise I'm afraid there is no point to deliver any special version for us, as we can't use it. @AlexandreMary Do you agree ?

FFT992 is still not put back in and I believe @marsdeno is still working on a pull request which does that. We had some pressure from elsewhere to make an ecTrans release so I didn't want to wait, but once Olivier is ready to open that pull request we can make sure to prioritise it and if necessary make a release just containing that change. Maybe we could create a 1.10.0 release which includes that and this PR rebased on 1.9.0.

dhaumont:
>we could create a 1.10.0 release which includes that and this PR rebased on 1.9.0.

I will let Alexandre discuss this with you. For the time beeing, we are not blocked as this PR will also be merged in our CY50t2_develop (https://github.com/ACCORD-NWP/ectrans/pull/13)

samhatfield:
> I will let Alexandre discuss this with you. For the time beeing, we are not blocked as this PR will also be merged in our CY50t2_develop ([ACCORD-NWP#13](https://github.com/ACCORD-NWP/ectrans/pull/13))

If you could upgrade all the way to 2.1.0 (assuming that's the release including this PR) you would be doing us a favour, but equally I understand how hard it can be to push non-bit-ID changes through even when they're trivial.

I'm about to merge this PR, just waiting for the build-hpc tests to pass (except for LUMI which is under maintenance today).

samhatfield (approved):

AlexandreMary:
Yes I agree, no need for a 1.10.0.

PR #459, opened 2026-08-26
Add test for multi-precision support

samhatfield:
In principle programs can be linked against both `libectrans_sp` and `libectrans_dp` in order to call `INV_TRANS` with multiple precisions. This capability is provided through the `*_{precision}.h` header files. The test in this PR gives an example of usage that should be supported.

Currently it's not possible, because there is a clash between backends over the "precision-dependent" structure modules, `TPM_FFTW` and `TPM_FLT`.

For example, on the first call to `SETUP_TRANS` (e.g. `SETUP_TRANS_SP`), the array of structs responsible for holding all of the FFT-related arrays for all resolutions, `TPM_FFTW_SP:FFTW_RESOL` is allocated. This array is accessed later on.

If the second call to `SETUP_TRANS` is `SETUP_TRANS_DP`, then `TPM_FFTW_DP:FFTW_RESOL` is not allocated because this logic is skipped when `TPM_DIM:DIM_RESOL` is allocated, which is the case. `TPM_DIM` is common so it's shared between the single- and double-precision backends.

Nevertheless, `TPM_FFTW_DP:FFTW_RESOL` is accessed further down and this triggers a segfault, because it is just not allocated.

In this PR I'll suggest solution(s) to resolve this problem.


samhatfield:
Change of role of this PR: this PR will only introduce the test for multi-precision support. This PR will build on other PRs which will gradually add support to allow this test to pass.

tests/trans/test_multiple_precisions.F90
```diff
@@ -0,0 +1,77 @@
...
+ALLOCATE(GRID_POINT_FIELD_SP(NGPTOT,1,1))
+ALLOCATE(SPECTRAL_FIELD_DP(1,NSPEC2))
+ALLOCATE(GRID_POINT_FIELD_DP(NGPTOT,1,1))
+
+! Initialise our spectral field arrays
+SPECTRAL_FIELD_SP(:,:) = 1.0_JPRM
+SPECTRAL_FIELD_DP(:,:) = 1.0_JPRD
+
+! Perform an inverse transform
+CALL INV_TRANS_SP(PSPSCALAR=SPECTRAL_FIELD_SP, PGP=GRID_POINT_FIELD_SP)
```
wdeconinck: Does each backend (in this case CPU_SP) keep track of its own NCUR_RESOL?
Otherwise we have to pass KRESOL as argument here to be explicit.

tests/CMakeLists.txt
```diff
@@ -591,6 +591,17 @@ if( HAVE_CPU AND HAVE_MPI )
...
 endif()
 
+# --------------------------------------------------------------------------------------------------
+# Add test for ability to call INV_TRANS at multiple precisions
+# --------------------------------------------------------------------------------------------------
+
+if( HAVE_SINGLE_PRECISION AND HAVE_DOUBLE_PRECISION )
+  ecbuild_add_test( TARGET ectrans_test_multiple_precisions
+    SOURCES trans/test_multiple_precisions.F90
+    LIBS ectrans_sp ectrans_dp fiat parkind_sp
```
wdeconinck: I think you should not need to link to parkind_sp here. ec_parkind is part of fiat.

tests/trans/test_multiple_precisions.F90
```diff
@@ -0,0 +1,77 @@
...
+!
+! This software is licensed under the terms of the Apache Licence Version 2.0
+! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+! In applying this licence, ECMWF does not waive the privileges and immunities
+! granted to it by virtue of its status as an intergovernmental organisation
+! nor does it submit to any jurisdiction.
+
+PROGRAM TEST_MULTIPLE_PRECISIONS
+
+USE PARKIND1, ONLY: JPRM, JPRD, JPIM
```
wdeconinck:
```suggestion
USE EC_PARKIND, ONLY: JPRM, JPRD, JPIM
```

tests/trans/test_multiple_precisions.F90
samhatfield: Good point.

wdeconinck (approved):
Great work @samhatfield , it was nice to discuss the entire process with you in multiple PRs.

PR #462, opened 2026-08-27
Make TRANS_END common

samhatfield:
`TRANS_END` is currently repeated for each backend and built as part of each backend's library. However it doesn't actually contain any precision-dependent code, so it is a candidate for moving to the common library. This will also allow us to call `TRANS_END` just once when using multiple backends within one program, like in #459.

This PR does some plumbing to make this possible. `TRANS_END` is solely responsible for deallocating common array structures which are backend independent. It also calls a "resol deallocator" for each allocated resol. When a backend calls `SETUP_TRANS`, it adds a pointer to its resol deallocator for `TRANS_END` to call. On top of that, each backend has its own "backend teardown" subroutine, and again, each backend registers its teardown subroutine in a common list.

With these changes, the test in #459 passes, so ecTrans finally supports multi-precision transforms. It also should support mixed CPU / GPU transforms, but I haven't tested that yet.


src/trans/gpu/internal/backend_teardown_mod.F90
```diff
@@ -0,0 +1,29 @@
...
+! which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+! In applying this licence, ECMWF does not waive the privileges and immunities
+! granted to it by virtue of its status as an intergovernmental organisation
+! nor does it submit to any jurisdiction.
+
+MODULE BACKEND_TEARDOWN_MOD
+
+USE GROWING_ALLOCATOR_MOD, ONLY: DESTROY_GROWING_ALLOCATOR
+USE TPM_FLT,               ONLY: S, FLT_RESOL
+USE TPM_FIELDS_GPU,        ONLY: FIELDS_GPU_RESOL, FG
```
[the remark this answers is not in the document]
samhatfield: Good spot.

src/trans/cpu/internal/set_resol_mod.F90
```diff
@@ -54,16 +55,25 @@ SUBROUTINE SET_RESOL(KRESOL,LDSETUP)
...
-  R => DIM_RESOL(NCUR_RESOL)
-  F => FIELDS_RESOL(NCUR_RESOL)
-  G => GEOM_RESOL(NCUR_RESOL)
-  D => DISTR_RESOL(NCUR_RESOL)
-  TW => FFTW_RESOL(NCUR_RESOL)
-  S => FLT_RESOL(NCUR_RESOL)
-  C => CTL_RESOL(NCUR_RESOL)
 ENDIF
 
+R => DIM_RESOL(NCUR_RESOL)
```
wdeconinck, line 70: Make a comment here that the re association is needed even when (IRESOL == NCUR_RESOL) for the copilot-reason mentioned for the `gpu/internal/set_resol_mod.F90` of released resolutions that could reuse the same RESOL (or something like this?). Also put this comment in that file.

src/trans/common/external/trans_end.F90
```diff
@@ -41,58 +41,58 @@ SUBROUTINE TRANS_END(CDMODE)
...
 
 IMPLICIT NONE
+
 CHARACTER(LEN=5), OPTIONAL,  INTENT(IN) :: CDMODE
 ! Local variables
-INTEGER(KIND=JPIM) :: JRES
+INTEGER(KIND=JPIM) :: JRES, JBACKEND
 CHARACTER(LEN=5) :: CLMODE
+
+IF (MSETUP0 == 0) CALL ABORT_TRANS('TRANS_END: TRANS NOT SETUP')
```
wdeconinck, line 66: Rather than abort, we could make this just return early; I guess that is the current behaviour implicitly.

wdeconinck (commented):
Great work @samhatfield this is exactly what I had in mind during our offline discussions. Of  course the fine details of wiring this together is the big effort here, thanks!
Just some minor comments.

wdeconinck:
Since this builds on #467 , and if because of this we can ONLY use ETRANS_INQ then this becomes an API change.
We need to verify that no user-code exists that would call TRANS_INQ instead of ETRANS_INQ in these cases.

Rather it would be good if there were a fix for ETRANS_INQ as well together with this PR, e.g. by having ETRANS and TRANS sharing the same backend registry? so that #467 is not absolutely required.
Perhaps we need an etrans unit test that actually makes sure that it is OK to use "TRANS_INQ" for ETRANS resols, so that we catch the API change.

wdeconinck (approved):
Thanks for addressing my comments.
Very good to have added the ectrans_trans_inq test!

src/trans/gpu/internal/set_resol_mod.F90
```diff
@@ -54,16 +55,37 @@ SUBROUTINE SET_RESOL(KRESOL,LDSETUP)
      ENDIF
    ENDIF
 ENDIF
+
+! Verify backend compatibility
+! Verify backend compatibility
```
samhatfield:
```suggestion
! Verify backend compatibility
```

PR #467, opened 2026-09-01
Make transi use ETRANS_INQ when running in LAM mode

samhatfield:
`trans_inquire_fstr` in transi uses both `trans_inq` and `etrans_inq` interchangeably, depending on the arguments requested. This was previously safe usage but will become dangerous once later changes such as https://github.com/ecmwf-ifs/ectrans/pull/462 are made. We are gradually building the notion of a "backend" in ecTrans, and this means we need to be strict about obscure usage of external subroutines, such as using `TRANS_INQ` to enquire about properties in a LAM context.

This PR changes the transi test so it always called `etrans_inq` when running in LAM mode, preemptively fixing potential failures once https://github.com/ecmwf-ifs/ectrans/pull/462 is merged.


src/transi/transi_module.F90
```diff
@@ -1290,156 +1290,229 @@ function trans_inquire_fstr(trans,vars_fstr) result(iret)
...
-
-    elseif( var == "rlapin" ) then
-      call allocate_ptr( trans%rlapin, trans%nsmax+4, double1 )
-      call TRANS_INQ( KRESOL=trans%handle,  PLAPIN=double1 )
-
-    elseif( var == "ndglu" ) then
-      call allocate_ptr( trans%ndglu, trans%nsmax+1, int1 )
-      call TRANS_INQ( KRESOL=trans%handle,  KDGLU=int1 )
-
-    elseif(       var /= "ndgl"         &
```
wdeconinck, line 1415: All the values in this `elseif` should be allowed as argument to trans_inquire.
It looks like we are a comprehensive unit-test to test all possible values. Sounds like a good job for copilot.

wdeconinck (changes requested):
Thanks this is certainly not a bad idea, regardless of #462 .
I have some worries that perhaps not all keys are accepted anymore. Please add a unit-test.

src/transi/transi_module.F90
samhatfield: Do you mean, we should implement those args? That seems best for a separate PR no?

samhatfield:
I have added two new transi tests, to check `trans_inquire` for both local and global transforms.

Along the way I spotted three arguments available for local transform inquiry which are not relevant for local transforms, so I deleted them. They are `rmu` (sine of latitudes, this is never initialised for local transforms), `rgw` (Gaussian weights, which is an argument of `ETRANS_INQ` but is just ignored), and `rpnm` (Legendre polynomials, not computed for local transforms).

src/transi/transi_module.F90
wdeconinck: No, they are not needed to be implemented as they are already stored in the Trans_t during the setup.
This was just making sure that the TRANS_INQ was not failing for those keys.

samhatfield: Oh I see - I deleted something I shouldn't have, and should put it back.

wdeconinck: Yes hence we need comprehensive unit-tests for this now.

samhatfield: Okay, I've added those back, and written tests for both global and local which check whether all args are accepted.

tests/transi/transi_test_trans_inquire_lam.c
```diff
@@ -0,0 +1,55 @@
...
+  const int nx = 20;
+  const int ny = 18;
+  const double dx = 2500.0;
+  const double dy = 2500.0;
+  const int tx = (nx - 1) / 2;
+  const int ty = (ny - 1) / 2;
+
+  TRANS_CHECK(trans_new(&trans));
+  TRANS_CHECK(trans_set_resol_lam(&trans, nx, ny, dx, dy));
+  TRANS_CHECK(trans_set_trunc_lam(&trans, ty, tx));
```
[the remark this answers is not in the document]
samhatfield: If this is wrong so is `ectrans_test_transi_lam`...

src/transi/transi_module.F90
```diff
@@ -1276,7 +1276,7 @@ function trans_inquire_fstr(trans,vars_fstr) result(iret)
   integer(c_int) :: iret
   type(Trans_t), intent(inout) :: trans
   character(len=*), intent(in) :: vars_fstr
-  character(20) :: var_arr(30), var
+  character(20) :: var_arr(48), var
```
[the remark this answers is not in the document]
wdeconinck: Indeed this goes out of bounds

src/transi/transi_module.F90
wdeconinck, line 1286: I would add here an assert that
`nvars < size(var_arr)`, even if we fix the above out-of-bounds

src/transi/transi_module.F90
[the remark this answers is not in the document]
samhatfield: Fixed by allocating `var_arr` exactly as many elements as it needs.

src/transi/transi_module.F90
samhatfield: Do I still need to do that? `var_arr` is now allocated to be exactly the right size.

wdeconinck: not necessary.

wdeconinck (approved):
Thanks @samhatfield all good!

PR #470, opened 2026-09-02
Make GEMM calls in LEDIRAD consistent

samhatfield:
The second GEMM call uses `JPRD` for alpha and beta, but all the other arguments are still `JPRB`. This is unlike the first GEMM call for each all arguments are `JPRB`.

For double-precision builds this PR obviously makes no difference. For single-precision builds it also makes no difference, because the GEMM method matching the generic interface with double-precision alpha/beta but single-precision arrays first converts alpha/beta to single-precision before calling a SGEMM. So this change is just for consistency and should be bit identical.


wdeconinck (approved):
Nice for consistency

PR #472, opened 2026-09-03
Move deprecated _32 routines to common

samhatfield:
These are obviously identical between CPU and GPU so don't need to go through the backend preprocessor.


wdeconinck:
What is the status with the IFS cycles still linking this symbol now?
If not, seeing this will be for a future release, we can finally think of removing.

samhatfield:
> What is the status with the IFS cycles still linking this symbol now? If not, seeing this will be for a future release, we can finally think of removing.

CY49R1, CY49R2: contains calls to `DIST_GRID_32` and `GATH_GRID_32`.
CY49R3, CY50R1, later cycles: contains no calls to the above.

CY49R1 doesn't have ecTrans in its bundle: it uses a hard-coded version in ifs-source/contrib/ectrans. So we don't have to worry about that as long as it won't be modified to take ecTrans from the bundle.

CY49R2 is the only one I think we have to worry about. Currently it uses ifs-source/contrib/ectrans, unless `--with-gpu` is passed. In that case, it uses `amdgpu` (ancient branch of ecTrans...). If they ever wanted to upgrade CY49R2 to the latest version of ecTrans, it wouldn't build if we removed these external routines entirely.

On the other hand, how likely is that?

wdeconinck:
I suppose OK to remove then.

samhatfield:
> I suppose OK to remove then.

To be very careful, can we merge this and defer complete removal until a later release? I'm just thinking about CY49R2.

wdeconinck (approved):

PR #473, opened 2026-09-03
Add benchmark command line argument for controlling LPGP_ON_GPU

samhatfield:
We now have an argument to `INV_TRANS` and `DIR_TRANS` that controls whether PGP arrays are copied back / to device, but there's no way to control it through the benchmark program. This PR adds this functionality. It also makes sure that any functionality that requires the arrays to be resident on host cannot be enabled at the same time.

@PaulMullowney

### DIFF
--- a/src/programs/ectrans-benchmark.F90
+++ b/src/programs/ectrans-benchmark.F90
@@ -194,6 +194,7 @@ type(fields_lists) :: ylf
 logical :: ldump_values = .false.
 logical :: lpinning = .false.
 logical :: ldump_checksums = .false.
+logical :: lpgp_on_gpu = .false.
 
 integer, external :: ec_mpirank
 logical :: luse_mpi = .true.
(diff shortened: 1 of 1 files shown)


src/programs/ectrans-benchmark.F90
```diff
@@ -1416,13 +1423,19 @@ subroutine get_command_line_arguments(nsmax, cgrid, iters, iters_warmup, nfld, n
...
       case default
         call parsing_failed("Unrecognised argument: " // trim(carg))
 
     end select
     iarg = iarg + 1
   end do
 
+  if (lpgp_on_gpu .and. (ldump_values .or. ldump_checksums)) then
+    call parsing_failed("Invalid combination of arguments: --keep-pgp-arrays-on-device cannot be&
+      &used with --dump-values or --dump-checksums")
```
wdeconinck, line 1436: and what about --norms?

samhatfield: Norms are based on spectral arrays which are always available on host currently. And yes, I have thought about adding a `LPSP_ON_GPU` argument so all device<->host copies can be controlled, not just those related to grid point arrays.

wdeconinck: Thanks, indeed you're right. I'm hoping in the end the field_view-based can be more explicit in this, what is given (input) or expected (output), as metadata in the field_view itself.
Doesn't mean we should not also have the PSP_ON_GPU option in the old API possible, when that is desired.

wdeconinck (approved):
Nice!

PR #475, opened 2026-09-08
Propagation of pointers

dhaumont:
### Description

In this PR `GRID_VIEW` and `SPEC_VIEW` pointers a propagated to TRLTOG for `INV_TRANS_FIELD_VIEW` and `DIR_TRANS_FIELD_VIEW`.

This PR supseeds https://github.com/dhaumont/ectrans/pull/11, that already contains some discussions and code reviewing:
    - [discussion - removal of TMP_TRANS%NPROMA and  TPM_TRANS%NGPBLKS](https://github.com/dhaumont/ectrans/pull/11#pullrequestreview-5089586434)
    - [discussion - removal of NPROMATR](https://github.com/dhaumont/ectrans/pull/11#discussion_r3903883012)
    - [discussion - code structure and field_view files introduction](https://github.com/dhaumont/ectrans/pull/11#issuecomment-5493850200)     

### Code modifications: 

1. duplication of the following internal routines to create the _view based version:

- CPU

```
meld ./src/trans/cpu/internal/dir_trans_ctl_mod.F90 src/trans/field_view/cpu/internal/dir_trans_view_ctl_mod.F90
meld ./src/trans/cpu/internal/fsc_mod.F90 src/trans/field_view/cpu/internal/fsc_view_mod.F90
meld ./src/trans/cpu/internal/ftdir_ctl_mod.F90 src/trans/field_view/cpu/internal/ftdir_view_ctl_mod.F90
meld ./src/trans/cpu/internal/ftinv_ctl_mod.F90 src/trans/field_view/cpu/internal/ftinv_view_ctl_mod.F90
meld ./src/trans/cpu/internal/inv_trans_ctl_mod.F90 src/trans/field_view/cpu/internal/inv_trans_view_ctl_mod.F90
meld ./src/trans/cpu/internal/ltdir_ctl_mod.F90 src/trans/field_view/cpu/internal/ltdir_view_ctl_mod.F90
meld ./src/trans/cpu/internal/ltdir_mod.F90 src/trans/field_view/cpu/internal/ltdir_view_mod.F90
meld ./src/trans/cpu/internal/ltinv_ctl_mod.F90 src/trans/field_view/cpu/internal/ltinv_view_ctl_mod.F90
meld ./src/trans/cpu/internal/ltinv_mod.F90 src/trans/field_view/cpu/internal/ltinv_view_mod.F90
meld ./src/trans/cpu/internal/prfi1b_mod.F90 src/trans/field_view/cpu/internal/prfi1b_view_mod.F90
meld ./src/trans/cpu/internal/trgl_mod.F90 src/trans/field_view/cpu/internal/trgl_view_mod.F90
meld ./src/trans/cpu/internal/trgtol_mod.F90 src/trans/field_view/cpu/internal/trgtol_view_mod.F90
meld ./src/trans/cpu/internal/trltog_mod.F90 src/trans/field_view/cpu/internal/trltog_view_mod.F90
meld ./src/trans/cpu/internal/updspb_mod.F90 src/trans/field_view/cpu/internal/updspb_view_mod.F90
meld ./src/trans/cpu/internal/updsp_mod.F90 src/trans/field_view/cpu/internal/updsp_view_mod.F90
```
- GPU

```
meld ./src/trans/gpu/internal/dir_trans_ctl_mod.F90 src/trans/field_view/gpu/internal/dir_trans_view_ctl_mod.F90
meld ./src/trans/gpu/internal/inv_trans_ctl_mod.F90 src/trans/field_view/gpu/internal/inv_trans_view_ctl_mod.F90
meld ./src/trans/gpu/internal/ltdir_mod.F90 src/trans/field_view/gpu/internal/ltdir_view_mod.F90
meld ./src/trans/gpu/internal/ltinv_mod.F90 src/trans/field_view/gpu/internal/ltinv_view_mod.F90
meld ./src/trans/gpu/internal/prfi1b_mod.F90 src/trans/field_view/gpu/internal/prfi1b_view_mod.F90
meld ./src/trans/gpu/internal/trgtol_mod.F90 src/trans/field_view/gpu/internal/trgtol_view_mod.F90
meld ./src/trans/gpu/internal/trltog_mod.F90 src/trans/field_view/gpu/internal/trltog_view_mod.F90
meld ./src/trans/gpu/internal/updspb_mod.F90 src/trans/field_view/gpu/internal/updspb_view_mod.F90
meld ./src/trans/gpu/internal/updsp_mod.F90 src/trans/field_view/gpu/internal/updsp_view_mod.F90
```
2. refactoring of `dir_trans_field_view` and `inv_trans_field_view` to remove the copy to temporary buffers (not needed anymore)

3. remove the `KFLDPTR` functionality for view mode

Note that the code can certainly be simplified further, my goal was to have a fiirst minimal version.

4. removal of dead code in `src/etrans/cpu/CMakeLists.txt`

### Test and validation

1. The new implementation has been tested manually on GNU (laptop), ATOS- CPU, ATOS - GPU.

2. The automated test launched on the historical PR are bit-reproducible

https://github.com/dhaumont/ectrans/actions/runs/34222665500

However, a performance issue was detected on GPU

### Performance issue

The following figures illustrate the 5x performance issue encountered when running on ATOS on GPU.

- reference run:
```
$build-nvidia/bin/ectrans-benchmark-gpu-dp -f 40 -l 40  --uvders --scders --vordiv 
ectrans_benchmark initialisation, on 1 tasks, took      0.38 sec
 
Running for 10 iterations with 3 extra warm-up iterations
 
Time step      1 took  2.7944
Time step      2 took  0.2076
Time step      3 took  0.2076
Time step      4 took  0.2111
Time step      5 took  0.2024
Time step      6 took  0.2088
Time step      7 took  0.2040
Time step      8 took  0.2043
Time step      9 took  0.2100
Time step     10 took  0.2077
Time step     11 took  0.2094
Time step     12 took  0.2100
Time step     13 took  0.2087
```

- field-api version:

```
$ build-nvidia/bin/ectrans-benchmark-gpu-dp -f 40 -l 40  --uvders --scders --vordiv --field-api 
ectrans_benchmark initialisation, on 1 tasks, took      0.38 sec
 
Running for 10 iterations with 3 extra warm-up iterations
 
Time step      1 took 10.8836
Time step      2 took  1.0803
Time step      3 took  1.0800
Time step      4 took  1.0781
Time step      5 took  1.0783
Time step      6 took  1.0773
Time step      7 took  1.0789
Time step      8 took  1.0798
Time step      9 took  1.0761
Time step     10 took  1.0771
Time step     11 took  1.0790
Time step     12 took  1.0801
Time step     13 took  1.0800
```

This performance issue is under investigation : https://github.com/ecmwf-ifs/ectrans/pull/441#issuecomment-5582304972

### History

- Initial idea by @pmarguinaud: https://github.com/pmarguinaud/ectrans/tree/inv_trans-split
- Preliminary PR : https://github.com/dhaumont/ectrans/pull/11
### Contributor Declaration


dhaumont:
Remaining todo's from https://github.com/dhaumont/ectrans/pull/11: 

[] add the missing INTENT in all the view routines
[] test that FSPGL_PROC calls are working as expected
[x] clean src/trans/field_view/CMakeLists.txt (https://github.com/dhaumont/ectrans/pull/11#issuecomment-5541914738)

Future PR:
- NPROMATR
- supression of TPM_TRANS%NPROMA and TPM_TRANS%NGPBLKS

samhatfield:
> * clean src/trans/field_view/CMakeLists.txt ([Propagate list of pointers dhaumont/ectrans#11 (comment)](https://github.com/dhaumont/ectrans/pull/11#issuecomment-5541914738))

Just did that :)

samhatfield:
I can reproduce the performance bugs with this branch and the slowdown (at least for a single MPI task) seems to be almost entirely caused by GSTATS region 1604 in `TRLTOG_VIEW_MOD`.

dhaumont:
> > * clean src/trans/field_view/CMakeLists.txt ([Propagate list of pointers dhaumont/ectrans#11 (comment)](https://github.com/dhaumont/ectrans/pull/11#issuecomment-5541914738))
> 
> Just did that :)

Thank you so much !

dhaumont:
> I can reproduce the performance bugs with this branch and the slowdown (at least for a single MPI task) seems to be almost entirely caused by GSTATS region 1604 in `TRLTOG_VIEW_MOD`.

I guess it's on the CPU  ? On the GPU I have seen many other issues.

Could you run your benchmark with the CPU and GPU builds and share your results, such that we have a baseline to compare with when we are optimizing ?

src/trans/field_view/inv_trans_field_view.F90
```diff
@@ -331,174 +196,46 @@ SUBROUTINE INV_TRANS_FIELD_VIEW(KRESOL, &
...
+IF (.NOT. ALLOCATED(YLGVSCALAR_EW))  ALLOCATE(YLGVSCALAR_EW(0))
+
+CALL INV_TRANS_VIEW_CTL(NPROMA, NBLK, &
+  & YLSPVVOR, YLSPVDIV, YLSPVSCALAR,&
+  & YLGVU,YLGVV,&
+  & YLGVVOR,YLGVDIV,&
+  & YLGVSCALAR,&
+  & YLGVU_EW,YLGVV_EW,&
+  & YLGVSCALAR_NS, YLGVSCALAR_EW,&
+  & FSPGL_PROC)
```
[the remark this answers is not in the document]
dhaumont: todo

src/etrans/cpu/CMakeLists.txt
```diff
@@ -32,11 +32,7 @@ function(generate_backend_sources)
   )
 
   set(outfiles)
-  foreach(file_i ${files})
```
dhaumont, line 35: Cleaning (nothing to do with this PR)

src/trans/gpu/internal/ltinv_mod.F90
```diff
@@ -344,31 +344,31 @@ SUBROUTINE LTINV(ALLOCATOR,HLTINV,KF_UV,KF_SCALARS,&
     CALL GSTATS(422,1)
 
     IF (KF_UV > 0) THEN
-      CALL PRFI1B(PVOR,PSPVOR,KF_UV,UBOUND(PSPVOR,2))
```
dhaumont, line 347: cleaning (nothing to do with this PR)

src/trans/gpu/internal/prfi1b_mod.F90
```diff
@@ -97,11 +96,11 @@ SUBROUTINE PRFI1B(PIA,PSPEC,KFIELDS,KDIM,KFLDPTR)
 
 #ifdef OMPGPU
     !$OMP TARGET TEAMS DISTRIBUTE PARALLEL DO COLLAPSE(3) DEFAULT(NONE) &
-    !$OMP& PRIVATE(KM,IASM0,INM) SHARED(KFIELDS,KDIM,D,R,PIA,PSPEC) MAP(TO:KFIELDS)
```
dhaumont, line 100: cleaning (nothing to do with this PR)

src/trans/field_view/ectrans_field_view_mod.F90
```diff
@@ -111,6 +111,7 @@ MODULE ECTRANS_FIELD_VIEW_MOD
   ! IVSET
   TYPE(C_PTR)    :: IVSET ! Value indexing is specified by IVSET_BASE
   INTEGER(C_INT) :: IVSET_SIZE
+  INTEGER(C_INT) :: IVSET_DIM ! C ORDERING; -1 when IVSET does not index distribution array dimension
```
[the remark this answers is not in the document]
dhaumont: probably a left overt from a merge that should be deleted, right ?

src/trans/field_view/ectrans_field_view_internal_util_mod.F90
```diff
@@ -103,7 +103,7 @@ FUNCTION GET_NPROMA_2(LIST1, LIST2) RESULT(NPROMA)
   NPROMA = 0
   CALL PROCESS_NPROMA(LIST1, NPROMA)
   CALL PROCESS_NPROMA(LIST2, NPROMA)
-END FUNCTION 
+END FUNCTION
```
dhaumont, line 106: add function name here

src/trans/field_view/ectrans_field_view_mod.F90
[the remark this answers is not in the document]
samhatfield: Probably. It looks like it can be safely deleted, but I didn't write this so I'm not 100% confident...

src/trans/field_view/dir_trans_field_view.F90
```diff
@@ -213,78 +129,25 @@ SUBROUTINE DIR_TRANS_FIELD_VIEW(KRESOL,                       &
...
-  ID = 1
-  DO JFLD = 1, IFLDSPSC
-    IF (ASSOCIATED(YLSPVSCALAR(JFLD)%P)) THEN
-      ZZ1_1=>YLSPVSCALAR(JFLD)%P
-      ZZ1_1(:) = ZPSPSC2(ID,:)
-      ID = ID + 1
-    ENDIF
-  ENDDO
-ENDIF
+CALL DIR_TRANS_VIEW_CTL(NPROMA,NBLK,YLSPVVOR, YLSPVDIV, YLSPVSCALAR,YLGVU, YLGVV, YLGVSCALAR)
```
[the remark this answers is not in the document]
samhatfield: This is urgent!

src/trans/field_view/ectrans_field_view_internal_util_mod.F90
samhatfield, line 46: Do we still need these?

samhatfield, line 58: And these?

samhatfield:
> I guess it's on the CPU ? On the GPU I have seen many other issues.
> 
> Could you run your benchmark with the CPU and GPU builds and share your results, such that we have a baseline to compare with when we are optimizing ?

No that was on GPU. I haven't looked at CPU performance yet. Our HPC is partly down this morning so when it's back I'll do some comprehensive benchmarks.

The problematic region I mentioned above is the following:
```fortran
      IRECV_WSET_OFFSET_V = IRECV_WSET_OFFSET(MYSETW)
      IRECV_WSET_SIZE_V = IRECV_WSET_SIZE(MYSETW)
      IIN_TO_SEND_BUFR_V = IIN_TO_SEND_BUFR_OFFSET(MYPROC)
      
      !$ACC PARALLEL LOOP COLLAPSE(2) DEFAULT(NONE) PRIVATE(JK,JBLK,IFLD,IPOS) &
      !$ACC&         FIRSTPRIVATE(KF_FS,IRECV_WSET_SIZE_V,IRECV_WSET_OFFSET_V, &
      !$ACC&         IIN_TO_SEND_BUFR_V,NPROMA) ASYNC(1)
      DO JFLD=1,KF_FS
        DO JL=1,IRECV_WSET_SIZE_V
          JK = MOD(IRECV_WSET_OFFSET_V+JL-1,NPROMA)+1
          JBLK = (IRECV_WSET_OFFSET_V+JL-1)/NPROMA+1
          IFLD = IFLDA(JFLD,1)
          IPOS = IIN_TO_SEND_BUFR(IIN_TO_SEND_BUFR_V+JL,1)+ &
              & (JFLD-1)*IIN_TO_SEND_BUFR(IIN_TO_SEND_BUFR_V+JL,2)+1
          YDGP(IFLD)%P(JK,JBLK) = PREEL_REAL(IPOS)
        ENDDO
      ENDDO
```
I'm no expert, but that looks like a bad access pattern on `YDGP` to me, because the `IFLD`s of `YDGP` aren't necessarily contiguous in memory?

dhaumont:
I believe we need to take these performance issues very seriously, as the `field_view` approach will never be adopted if performance can't match that of the current baseline solution.

The main issues lie the GPU to GPU copies transferring data from the fields themselves to internal  working arrays (within PRF1B, DPDP, TRLTOG and TRGTOL).
At this stage, I don't see how to perform these copies efficiently using a list of pointers.
The baseline version performs copies from PGP/PSP arrays, which is hightly efficient. However, the compiler can't optimize the copy when fields are identified by pointers, as no assumption can be made regarding the memory layout. This reproducer can be used to investigate this issue (https://github.com/destination-earth-digital-twins/DE330_GPU/tree/main/support/ATOS/copy_list_of_ptrs) or in this tar ball 
[copy_list_of_ptrs.tar.gz](https://github.com/user-attachments/files/32001491/copy_list_of_ptrs.tar.gz)


I'm also concerned about the `YDSP(JFLD)%{INDEX)` layout , and it's compatibility with the current GPU version, which relies on on a `PSPEC(JFLD, INDEX)` layout. The fastest-varying index differs between the two representations, and further adaptation  of the GPU code will likely be required to ensure code efficiency

dhaumont:
> I'm no expert, but that looks like a bad access pattern on `YDGP` to me, because the `IFLD`s of `YDGP` aren't necessarily contiguous in memory?

Exactly. This problem is present in every `YDGP(IFLDS)` loops (see my comment above), so I'm quite surprised you only see this in this one apperaing in your profiling.

samhatfield:
> This reproducer can be used to investigate this issue (https://github.com/destination-earth-digital-twins/DE330_GPU/tree/main/support/ATOS/copy_list_of_ptrs)

This is a private repo by the way, I can't access it.

> Exactly. This problem is present in every YDGP(IFLDS) loops (see my comment above), so I'm quite surprised you only see this in this one apperaing in your profiling.

Because all the kernels are non-blocking (`ASYNC(1)`) it makes interpreting GSTATS regions a bit tricky. I think I added barriers where necessary but I might have missed a few. I'll check more carefully later.

dhaumont:
> This is a private repo by the way, I can't access it.

I'm attaching a tarball containing the reproducer

> Because all the kernels are non-blocking (`ASYNC(1)`) it makes interpreting GSTATS regions a bit tricky. I think I added barriers where necessary but I might have missed a few. I'll check more carefully later.

To facilitate the analysis of profiling traces, maybe it makes sense to run without ASYNC, using `export PGI_ACC_SYNCHRONOUS=1`

src/trans/field_view/gpu/internal/prfi1b_view_mod.F90
```diff
@@ -0,0 +1,139 @@
...
+    !$ACC& ASYNC(1)
+#else
+    !$ACC&
+#endif
+#endif
+  DO KMLOC=1,D_NUMP
+    DO JN=0,R_NSMAX+3
+      DO JFLD=1,IFIELDS
+        KM = D_MYMS(KMLOC)
+        IF (JN+1 <= UBOUND(PIA,2)) THEN
```
dhaumont: The ` IF (JN+1 <= UBOUND(PIA,2)) THEN` test can be removed, as the out-of-bound access is solved in develop now

dhaumont:
For gpu performance investigation, I have created a branch propagate_ptr_gpu2 on top of this one, with some additional optimizations for GPU
https://github.com/dhaumont/ectrans/tree/propagate_ptr_gpu2/src/trans/field_view/gpu/internal

See for instance the CACHE implementation of prfi1b_view_mod: https://github.com/dhaumont/ectrans/blob/propagate_ptr_gpu2/src/trans/field_view/gpu/internal/prfi1b_view_mod.F90

src/trans/field_view/gpu/internal/prfi1b_view_mod.F90
dhaumont: done

dhaumont:
> I believe we need to take these performance issues very seriously, as the `field_view` approach will never be adopted if performance can't match that of the current baseline solution.

A possible solution to our performance issue would be to extend:
- SPEC_VIEW to store `P (JFLD, JLEV, :)` instead of `P (:)`
- GRID_VIEW to store `P(J:,FLD,JLEV,)`instead of  `P(:,:)`

This would allow us to repack the fields that are coming from the same matrix together, and process them with efficient array notation on GPU, as before.
 This is  equivalent to re-create the PGP2, PSP2, PGPUV arrays  on the fly and propagate them as list of (multidimensional) FIELD_VIEW and SPEC_VIEW.

For this to work, we would have to detect the pointers that can be grouped, and repack them together.  
I can imagine some recombination algorithm based on pointer arithmetic, grouping the pointers that have a regular indexing pattern,
But if this is too complicated to implement, we could also change the ectrans field API interface such that it takes "list of sublist of pointers": the pointers in the  sublist would grouped together in a single SPEC_VIEW (or GRID_VIEW) object.

msleigh:
A number of new NVIDIA copyright statements are included. Do we have a licence from NVIDIA to include / distribute this IP in ecTrans (and is it compatible with our Apache licence)?

samhatfield:
> A number of new NVIDIA copyright statements are included. Do we have a licence from NVIDIA to include / distribute this IP in ecTrans (and is it compatible with our Apache licence)?

The new files are actually entirely written by @dhaumont I think, so on that basis they probably shouldn't have any Nvidia license statements. I think they were added by mistake, is that right @dhaumont?

On the other hand, those files take a very heavy inspiration from files actually written by Nvidia. Compare [`PRFI1B` (Nvidia)](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/internal/prfi1b_mod.F90) and [`PRFI1B` (us)](https://github.com/dhaumont/ectrans/blob/propagate_ptr/src/trans/field_view/gpu/internal/prfi1b_view_mod.F90) for example. I'm not sure what the license implications are there, it's a gray area.

samhatfield:
> A possible solution to our performance issue would be to extend:
> 
> * SPEC_VIEW to store `P (JFLD, JLEV, :)` instead of `P (:)`
> * GRID_VIEW to store `P(J:,FLD,JLEV,)`instead of  `P(:,:)`
> 
> This would allow us to repack the fields that are coming from the same matrix together, and process them with efficient array notation on GPU, as before. This is equivalent to re-create the PGP2, PSP2, PGPUV arrays on the fly and propagate them as list of (multidimensional) FIELD_VIEW and SPEC_VIEW.
> 
> For this to work, we would have to detect the pointers that can be grouped, and repack them together. I can imagine some recombination algorithm based on pointer arithmetic, grouping the pointers that have a regular indexing pattern, But if this is too complicated to implement, we could also change the ectrans field API interface such that it takes "list of sublist of pointers": the pointers in the sublist would grouped together in a single SPEC_VIEW (or GRID_VIEW) object.

Yeah I guess the issue is that we can't do a coalesced write on `YDGP` because it is scattered throughout device memory rather than being contiguous.

Let's take a step back: what was the reason for converting everything into a "list of 2D arrays" in the first place?

dhaumont:
> Let's take a step back: what was the reason for converting everything into a "list of 2D arrays" in the first place?

To be as flexible as possible. By converting everything into list of 2d arrays, you can group different fields into a single list, regardless of their dimension or source array, and transform all of them in one go..
Furthermore, flattening everything simplifies the code, as there is no longer a neeed for separate code paths to handle PGP or PGP2, PGP3A, etc. 
But I think we can find a compromise, that preserves sufficient flexibility without sacrificing performance.

dhaumont:
> > A number of new NVIDIA copyright statements are included. Do we have a licence from NVIDIA to include / distribute this IP in ecTrans (and is it compatible with our Apache licence)?
> 
> The new files are actually entirely written by @dhaumont I think, so on that basis they probably shouldn't have any Nvidia license statements. I think they were added by mistake, is that right @dhaumont?
> 
> On the other hand, those files take a very heavy inspiration from files actually written by Nvidia. Compare [`PRFI1B` (Nvidia)](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/internal/prfi1b_mod.F90) and [`PRFI1B` (us)](https://github.com/dhaumont/ectrans/blob/propagate_ptr/src/trans/field_view/gpu/internal/prfi1b_view_mod.F90) for example. I'm not sure what the license implications are there, it's a gray area.

Indeed, most of the code in FIELD_VIEW is a copy-pasted and adpated version of the NVIDIA GPU code, and Meteo-France designed this list of pointer approach (@pmarguinaud).  So for me, it would be fair to leave the same copyright as in the original files.

wdeconinck:
I don't mind if you add NFLD and NLEV as part of the GRIDVIEW and SPECVIEW. This is all an internal detail behind the FIELD_VIEW API.  This is exactly why I was insisting on the separation of concerns. GRIDVIEW and SPECVIEW are and should always remain **private**.
```
GRIDVIEW:
    REAL, POINTER :: DATA(NPROMA,NLEV,NFLD,NBLK)
    INTEGER, POINTER :: IVSET(NLEV)
  
 SPECVIEW:
    REAL, POINTER :: DATA(NLEV,NSPEC2,NFLD)
    INTEGER :: MYVSET
 ```
 So a 2D gridpoint field_view (NPROMA,NBLK) should map to a GRIDVIEW with NLEV=1 and NFLD=1,
 And a 3D gridpoint field_view (NPROMA,NLEV,NBLK) should map to a GRIDVIEW with NFLD=1,
 Also we should assert that when we have the lists of GRIDVIEW and SPECVIEW created that they are matching in expected dimensions of NFLD and NLEV.
 Then all deeper implementations are still uniformly expecting lists of GRIDVIEW and lists of SPECVIEW, but now with  GRIDVIEW%DATA and SPECVIEW%DATA to have 4 and 3 dimensions respectively.
Note, most of the times the %DATA will now also be contiguous, but that should not be mandated. It would be up to how the  FIELD_API-FIELD DATA is strided that would set if the data is contiguous.

**Comment by msleigh** [comment]
> > > A number of new NVIDIA copyright statements are included. Do we have a licence from NVIDIA to include / distribute this IP in ecTrans (and is it compatible with our Apache licence)?
> > 
> > 
> > The new files are actually entirely written by @dhaumont I think, so on that basis they probably shouldn't have any Nvidia license statements. I think they were added by mistake, is that right @dhaumont?
> > On the other hand, those files take a very heavy inspiration from files actually written by Nvidia. Compare [`PRFI1B` (Nvidia)](https://github.com/ecmwf-ifs/ectrans/blob/develop/src/trans/gpu/internal/prfi1b_mod.F90) and [`PRFI1B` (us)](https://github.com/dhaumont/ectrans/blob/propagate_ptr/src/trans/field_view/gpu/internal/prfi1b_view_mod.F90) for example. I'm not sure what the license implications are there, it's a gray area.
> 
> Indeed, most of the code in FIELD_VIEW is a copy-pasted and adpated version of the NVIDIA GPU code, and Meteo-France designed this list of pointer approach (@pmarguinaud). So for me, it would be fair to leave the same copyright as in the original files.

That's fine as long as we have a licence to use that NVIDIA-copyrighted material. However from Sam's links in the comment above, it looks like the code that you have adapted _from_ is already in this repository, is that generally true? If so then I don't think there's anything to worry about and we can go ahead. Can you confirm?

**Comment by dhaumont** [comment]
The performance hit is due to the misuse of `get_device_ptr` from field API : we retrieve a cpu adress, but use it as a gpu adress in our kernel This is because the pointer retrieved by `get_device_ptr` is stored in the data member of a struct, and it won't be converted to a gpu pointer by openacc (no "deep conversion" of the pointers in the type). The kernel doesn't crash (don't ask me why, Philippe doesn't know either) but this results in a significant performance penalty. Performances improve when we explicitely convert pointers from CPU to GPU, and then rely on the openacc `USE_DEVICE` directive. I'm implementing a bug fix.

Many thanks to @pmarguinaud and Nicolas for identifying the error 

Note that I adressed this issue in the reproducer by using the `ATTACH` directive, but I didn't realize that it was also needed when we retrieve fields from field API.
Nevertheless, I still expect a slow down due to pointers introductions, but order of magnitude smaller. We will time it when the conversion is in place.

**Comment by dhaumont** [comment]
@wdeconinck @samhatfield @pmarguinaud 
I just pushed a bug fix for the wrong usage of get_device_ptr on GPU: e744558cdb4021605e687ec4e0fcf1325c5e6b7e
The bottom lines are :
 - conversion of the %P pointers from CPU adresses to GPU pointer adresses ( see src/trans/field_view/ectrans_field_view_internal_util_mod.F90)
 - add back the llacc to use get_device_ptr instead of get_host_device_ptr, when retrieving the field api pointers

This removes a bit bottleneck, but we still observe a performance penatly, the code is still about 2x slower then the reference GPU code.

@wdeconinck 
1. I know you are not keen to add an ldacc argument in inv_trans_field_api and dir_trans_field_api. Do you  see how we can remove it ?
2. Where do you want the CPU to GPU conversion ? The thing is that get_device_ptr doesn't return a GPU adress directly: it returns the CPU adress of the GPU buffer, and rely on openacc to retrieve the GPU adress of the buffer. I was not sure wether all your C pointer management code would be able to cope with GPU pointers, so I decided to put the   openacc conversion in field_view, right after the reconstruction of the Fortran pointer from the C pointer. But maybe a better place would be to move this code sooner, when we convert the get_device_ptr pointer to a C pointer?

**Comment by dhaumont** [comment]
I just profiled using nsys. It seems that the performance issue in the kernels themselves are gone, and an iteration is executed at the same speed at the reference version (or even faster). But we have some synchronisation device to host at the end of the iteration stalling completely the pipeline: 
image

In the CUDA API summary, there is this additional line when using field API that does not appear with the reference run: 
Time	Total Time	Num Calls	Avg	Med	Min	Max	StdDev	Name
29.8%	1.343 s	208	6.456 ms	873.109 μs	4.240 μs	35.258 ms	12.534 ms	cuMemcpyDtoHAsync_v2

**Comment by dhaumont** [comment]
I just removed the useless calls to synchost_dronly from the benchmark : 77c52191cb1c735d044522316c7faf63332e4f69

The benchmark is now running 10x faster with field API than withtout !
This can be explained by the fact that data stay now in GPU memory during the iterations. 
One could argue that we are now comparing apples to oranges, but it's also fair to say that we also don't have to transfer the fields back to the host during a forecast.


- Benchmark timings : with field API
```
$ build-nvidia/bin/ectrans-benchmark-gpu-dp -f 40 -l 40  --uvders --scders --vordiv --field-api
...
Time step      2 took  0.0255
...
Time step     12 took  0.0222
Time step     13 took  0.0222
```
- Benchmark timings :  reference version (no field API)

```
build-nvidia/bin/ectrans-benchmark-gpu-dp -f 40 -l 40  --uvders --scders --vordiv  
...
Time step      2 took  0.1957
...
Time step     12 took  0.1908
Time step     13 took  0.1917
...
```
Here are the full logs:
- [field_api.txt](https://github.com/user-attachments/files/32567153/field_api.txt)
- [reference.txt](https://github.com/user-attachments/files/32567154/reference.txt)

Any idea on how to improve the benchmark such that the field api and the regular version are comparable ?

**Comment by dhaumont** [comment]
> Any idea on how to improve the benchmark such that the field api and the regular version remains comparable ?

1. I added back some of the device-to-host copies, to better match the behaviour of the reference benchmark 953d56ab1e3e9df09f6483e4d299b1c3f1452bb1

2. I noticed that the  problematic synchronization was the one just after `inv_trans_field_api`.
When you force the copy of all the pgp fields,  performance drops: the time step duration goes from 0.02s to 0.48s.
This can be reproduced by changing line L698 from `ectrans-benchmark.F90 `, from
```
call synchost_rdonly_wrapped_fields(ywflds, .TRUE., .FALSE.,.FALSE.)
```
into:
```
call synchost_rdonly_wrapped_fields(ywflds, .TRUE., .FALSE.,.TRUE.)
```

3. I have tried to use the feature [asynchrous sync_host_rdonly](https://github.com/ecmwf-ifs/field_api#asynchronism) from field api, to measure its impact on the runtime, but unfortunately it didn't work. It requires CUDA, and I was not able to run any benchmark compiled with  `ENABLE_CUDA=ON`

PR #217, opened 2025-02-20
Flag more versions of gfortran as having issue with CONTIGUOUS

fmahebert:
Improves the CMake logic that identifies `gfortran` versions that have trouble compiling the `CONTIGUOUS` keyword. 

Please see https://github.com/ecmwf-ifs/ectrans/issues/216 for details and discussion.

PR #476, opened 2026-09-09
[transi] Fix for zero-sized local spectra which may happen with large MPI_size

wdeconinck:
### Description

This pull request makes several improvements to the handling of optional arguments and zero-sized arrays in the spectral transform routines `DIST_SPEC`, `GATH_SPEC`, and `SPECNORM`.

It can happen, especially in tests or benchmarks with low truncation and high number of MPI tasks, that not all ranks have any wave numbers to process, this PR allows then that the PSPEC array does not necessarily need to be passed, if it avoids compiler quirks or bugs dealing with zero-sized arrays.

This PR adds more robust checks for the presence of arrays, ensures correct behavior when local spectral sizes are zero, and updates the interface and internal logic to handle these cases gracefully. Additionally, a new `transi` test is added to verify correct handling of zero-sized or not-present local spectra in MPI runs.

Key changes include:

### Improved handling of optional and zero-sized arrays

* Updated `DIST_SPEC`, `GATH_SPEC`, and `SPECNORM` subroutines to check the presence and size of `PSPEC` before use, aborting with a clear message if necessary, and to call control routines with the correct argument combinations depending on which arrays are present.
* Modified internal control and normalization routines (`SPNORM_CTL`, `SPNORMC`) to handle cases where `PSPEC` or related arrays are not present or are zero-sized, including sending empty buffers in MPI communication as needed.

### More robust Fortran–C interface logic

* Updated the `trans_distspec`, `trans_gathspec`, and `trans_specnorm` C interfaces to nullify and check pointers before use, and to only call Fortran routines with `PSPEC` when it is associated, ensuring correct operation when local spectral sizes are zero (`nspec2==0`).

### Build system and testing

* Added a new test (`ectrans_test_transi_zero_local_spectral`) to verify correct handling of zero-sized local spectra in MPI runs, improving test coverage for edge cases.
* I expect that this is also a valid workaround for following issues with the CONTIGUOUS attribute:
    -  https://github.com/ecmwf-ifs/ectrans/pull/98
    -  https://github.com/ecmwf-ifs/ectrans/pull/217

### Contributor Declaration


wdeconinck:
The HPC ci is failing because the added tests requires 5 MPI ranks, and we don't have the allocation. I will try to make the test work with 4.

wdeconinck:
@samhatfield all ready for review.

src/trans/cpu/external/dist_spec.F90
```diff
@@ -200,8 +202,23 @@ SUBROUTINE DIST_SPEC(PSPECG,KFDISTG,KFROM,KVSET,KRESOL,PSPEC,&
...
+   & KUMPP=IUMPP,KALLMS=IALLMS,KPTRMS=IPTRMS,KN=IKN,KSORT=KSORT)
+ELSE IF (.NOT. PRESENT(PSPECG) .AND. PRESENT(PSPEC)) THEN
+  CALL DIST_SPEC_CONTROL(KFDISTG=KFDISTG,KFROM=KFROM,KVSET=IVSET,PSPEC=PSPEC,LDIM1_IS_FLD=LLDIM1_IS_FLD,&
+   & KSMAX=ISMAX,KSPEC2=ISPEC2,KSPEC2MX=ISPEC2MX,KSPEC2G=ISPEC2_G,KPOSSP=IPOSSP,KDIM0G=IDIM0G,&
+   & KUMPP=IUMPP,KALLMS=IALLMS,KPTRMS=IPTRMS,KN=IKN,KSORT=KSORT)
+ELSE
+  CALL DIST_SPEC_CONTROL(KFDISTG=KFDISTG,KFROM=KFROM,KVSET=IVSET,LDIM1_IS_FLD=LLDIM1_IS_FLD,&
+   & KSMAX=ISMAX,KSPEC2=ISPEC2,KSPEC2MX=ISPEC2MX,KSPEC2G=ISPEC2_G,KPOSSP=IPOSSP,KDIM0G=IDIM0G,&
+   & KUMPP=IUMPP,KALLMS=IALLMS,KPTRMS=IPTRMS,KN=IKN,KSORT=KSORT)
+ENDIF
```
samhatfield, line 221: Is this splitting of calls because some compilers can't handle optionals properly?

src/trans/cpu/external/specnorm.F90
```diff
@@ -130,7 +130,12 @@ SUBROUTINE SPECNORM(PNORM,PSPEC,KVSET,KMASTER,KRESOL,PMET)
   ENDIF
 ENDIF
 
-CALL SPNORM_CTL(PNORM,PSPEC,IFLD,IFLD_G,KVSET,IMASTER,PMET)
+
+IF(PRESENT(PSPEC)) THEN
+  CALL SPNORM_CTL(PNORM=PNORM,KFLD=IFLD,KFLD_G=IFLD_G,KVSET=KVSET,KMASTER=IMASTER,PMET=PMET,PSPEC=PSPEC)
+ELSE
+  CALL SPNORM_CTL(PNORM=PNORM,KFLD=IFLD,KFLD_G=IFLD_G,KVSET=KVSET,KMASTER=IMASTER,PMET=PMET)
+ENDIF
```
samhatfield, line 138: Same question as above.

src/transi/transi_module.F90
```diff
@@ -2621,28 +2621,39 @@ function trans_distspec(args) bind(C,name="trans_distspec") result(iret)
...
     if( .not. c_associated(args%rspecg) ) then
       call transi_error( "trans_distspec: ERROR: Array RSPECG was not allocated" )
     endif
     call c_f_pointer( args%rspecg, RSPECG, (/isend,trans%nspec2g/) )
     if( .not. is_lam(trans) ) then
-      call DIST_SPEC(KRESOL=trans%handle,KFDISTG=args%nfld,KFROM=NFROM,PSPEC=RSPEC,PSPECG=RSPECG)
+      if (associated(RSPEC)) then
+        call DIST_SPEC(KRESOL=trans%handle,KFDISTG=args%nfld,KFROM=NFROM,PSPECG=RSPECG,PSPEC=RSPEC)
+      else
+        call DIST_SPEC(KRESOL=trans%handle,KFDISTG=args%nfld,KFROM=NFROM,PSPECG=RSPECG)
```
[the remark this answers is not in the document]
samhatfield: Yes, very good point. Can you do this @wdeconinck?

tests/transi/transi_test_zero_local_spectral.c
```diff
@@ -0,0 +1,209 @@
...
+    fprintf( stderr, "rank %d/%d: rspecg_norm = [", trans.myproc, trans.nproc );
+    for( int field = 0; field < nfld; ++field )
+      fprintf( stderr, " %g", rspecg_norm[field] );
+    fprintf( stderr, " ]\n" );
+  }
+  compute_global_spectral_norm_from_spectral_values( nfld, trans.nsmax, expected_norm );
+
+  if( trans.myproc == 1 ) {
+    for( int field = 0; field < nfld; ++field ) {
+      ASSERT( is_approx_eq( rspecg_norm[field], expected_norm[field] ) );
```
samhatfield, line 152: Can this assertion ever actually fail? In other words, what are we actually testing here?

samhatfield (changes requested):
Looks good @wdeconinck, nice work. Just a few comments, and one outstanding issue: implementing something similar for the GPU versions (this is a nitpicking point because the GPU versions are actually identical (only `GPNORM_TRANS` is "GPU aware")).

src/trans/cpu/external/dist_spec.F90
wdeconinck: Yes, that should be an alternative workaround now for the problematic CONTIGUOUS attribute for some compilers, especially when the arrays are zero-sized, also in gath_spec and specnorm.

src/transi/transi_module.F90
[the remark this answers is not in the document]
wdeconinck: OK I will do this.

tests/transi/transi_test_zero_local_spectral.c
wdeconinck: This is more defensive.
- `expected_norm` has been computed from the prescribed `spectral_value` function
- `rspecg_norm` is the spectrum computed that was initialised in `rspecg` on myproc==1.
They should match.

Further then `rnorm` gets computed from a distributed `rspec` with `trans_specnorm`, which should also match.

Finally after gathering again the `rspecg` from the roundtrip, we should recover again the `expected_norm`.

src/transi/transi_module.F90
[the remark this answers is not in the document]
wdeconinck: The PR now contains transi tests for GPU/CPU SH/LAM

wdeconinck:
The additional tests added for LAM+GPU exposed a bug only present in the GPU version, fixed with commit https://github.com/ecmwf-ifs/ectrans/pull/476/commits/c60827f50e3fdf4fa179d8b933104469cab3f91b.
This also showed that the GPU and CPU versions of gath_spec and dist_spec have diverged, notably since 2019.
@samhatfield it should be a follow-up action after this PR to bring those closer together again.

cmake/ectrans_contiguous_workaround.cmake
```diff
@@ -20,4 +20,7 @@ if( NOT DEFINED ECTRANS_HAVE_CONTIGUOUS_ISSUE )
       set( ECTRANS_HAVE_CONTIGUOUS_ISSUE True )
     endif()
   endif()
+
+  # I believe the above is no longer necessary with PR https://github.com/ecmwf-ifs/ectrans/pull/476
+  set( ECTRANS_HAVE_CONTIGUOUS_ISSUE False )
```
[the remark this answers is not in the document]
wdeconinck: I expect that the splitting of the calls together with the "nspec2 == 0" checking solves this issue.
This is an opportunity to see if it solves it without completely removing the workaround yet. We can later easily remove this line 24-25 if it appears to be still needed. We can ask the authors of the issue to verify the develop branch after merge, and then completely remove it.
What do you think @samhatfield ?

wdeconinck:
I can confirm my findings with intel-llvm (IFX) compiler 2025.3 is that the `CONTIGUOUS` attribute in CPU `DIST_SPEC_CONTROL` is problematic when any of the fields `PSPEC` or `PSPECG` are not present.
When the `CONTIGUOUS` attribute is removed, the problem goes away. This is in line with previous `CONTIG_BUGGY_COMPILER` workaround in `DIST_SPEC_CONTROL`, which was applied for `PSPECG`.
With the changes of this PR I noticed that the workaround is also required for `PSPEC` now.
So I verified that if we add the `CONTIG_STATUS` macro in `DIST_SPEC_CONTROL` for `PSPEC` it works OK.
This workaround was previously only applied for older intel compiler and for some gfortran compilers, but it seems also required for the newer intel compilers. 

Another workaround I found with this branch development which allows us to still have the `CONTIGUOUS` attribute, is to split the calls to `DIST_SPEC_CONTROL`.
Since the GPU code path for `DIST_SPEC` doesn't use `CONTIGUOUS` in its `DIST_SPEC_CONTROL`, the previous changes made for split calls are not needed here, and I reverted this since last review (https://github.com/ecmwf-ifs/ectrans/pull/476/commits/f84258b036b7dd296ac8b9d2fa7eb50167dc841a)
Also other code paths (including CPU) for GATH_SPEC and SPECNORM do not have this problem, and I also reverted the changes there.

Probably it would be good in the future  to create a standalone reproducer that triggers this runtime problem, and confirm it as a compiler bug that can be reported.

samhatfield:
So my summary on this `CONTIGUOUS` issue:

Some compilers are buggy, they don't treat `CONTIGUOUS, OPTIONAL` arguments properly. We have to support these compilers. Previously we just stripped out `CONTIGUOUS` for these compilers. However it's better to leave `CONTIGUOUS` in place, and instead split calls so optionals are just not passed. That's what this PR does. However, it leaves in place the previous workaround (`CONTIG_STATUS`).

After this is merged, we should remove the `CONTIG_STATUS` workaround, and see if the call splitting allows us to leave the `CONTIGUOUS` keyword in place for the buggy compilers.

The `SPLIT_CALLS` function is purely for debugging this issue. When or if we eventually deprecate support for buggy compilers, we will remove this function, effectively setting `SPLIT_CALLS=.FALSE.`.

samhatfield (approved):
Good work @wdeconinck navigating this surprisingly complicated refactoring.

PR #477, opened 2026-09-10
Remove m=0 special treatment code path from GPU inverse Legendre transform

samhatfield:
In the CPU version of the global transforms, we treat the zonal wavenumber 0 mode differently to the others for the _direct_ transform: it is always performed at double precision. This is for scientific reasons: this mode is especially prone to "swamping" due to rounding errors, in which one value in a summation dominates the others and leads to inaccuracies. This was observed during testing of the single-precision variant of the IFS. The m=0 mode contains the global average of fields (m=0, l=0) which is generally much higher in value than the other modes (m=0, l/=0). Our solution is to perform this transform at double precision, always.

As far as we're aware, no such problem exists for the _inverse_ transform, so in `LEINV`, everything is done with single precision when `JPRB=JPRM`.

The GPU variant of ecTrans applies the m=0 special treatment in both directions. I think this was probably mistaken because, as mentioned above, there's no scientific justification. Perhaps @lukasm91 can confirm.

Wherever possible, this special treatment should be avoided. It adds additional code complexity and, more importantly, leads to a load imbalance in which the task handling m=0 contains just one special double-precision mode whereas all the others are exclusively single precision.

This PR removes the m=0 special treatment from the inverse Legendre transform on GPUs.

Benchmark results to follow.


samhatfield:
Single-node benchmarks on JUPITER indicate no impact.

`develop` (https://github.com/ecmwf-ifs/ectrans/commit/6c180d8500a0864aeb34df48e672d4eabb4d7d29)
| Truncation | Inverse transform (s) | Direct transform (s) | Round trip (s) |
|------------|---------|---------|----------|
|         79 |  0.0015 |  0.0011 |   0.0026 |
|        159 |  0.0034 |  0.0019 |   0.0053 |
|        319 |  0.0107 |  0.0051 |   0.0158 |
|        639 |  0.0407 |  0.0185 |   0.0596 |
|       1279 |  0.1619 |  0.0736 |   0.2368 |

`feat/remove_leinv_m0_dp` (https://github.com/ecmwf-ifs/ectrans/pull/477/commits/027bd3569aa2aaf86fc3dd4895c4e48de96f1d93)

| Truncation | Inverse transform (s) | Direct transform (s) | Round trip (s) |
|------------|---------|---------|----------|
|         79 |  0.0014 |  0.0011 |   0.0025 |
|        159 |  0.0033 |  0.0019 |   0.0052 |
|        319 |  0.0108 |  0.0054 |   0.0161 |
|        639 |  0.0404 |  0.0185 |   0.0589 |
|       1279 |  0.1613 |  0.0735 |   0.2359 |

All double precision, with options `--norms --niter 100 --nlev 137 --nfld 1 --uvders --scders --meminfo`. 4 GPUs / tasks per node.

Not sure if this is surprising or not, but in any case what we really care about is multi-node results.

PR #480, opened 2026-09-16
Propagate OpenMP flags as link options for NVHPC

PaulMullowney:
Downstream Fortran, C and C++ targets that link ectrans_gpu need the OpenMP offload flags on the link line, which was already handled for Cray. nvfortran has the same requirement, so extend the condition and add the matching NVHPC generator expressions for the C and C++ link languages.


wdeconinck (approved):
This looks OK to me. Thanks @PaulMullowney !

